PR #155, opened 2026-02-08
feat: Add Visual World Model (VWM) with 4D Gaussian splatting

ruvnet:
Implements ADR-018: Visual World Model as a Bounded Nervous System.

Core crate (ruvector-vwm):
- 4D Gaussian primitives with temporal deformation and screen projection
- Spacetime tile system with quantization tiers (Hot8/Warm7/Warm5/Cold3)
- Packed draw list protocol for deterministic GPU rendering
- Coherence gate for update acceptance/rejection with rollback support
- Append-only lineage log with full provenance tracking
- Entity graph for objects, tracks, regions with typed edges
- Streaming protocol with keyframe/delta/semantic packets and bandwidth budget

WASM bindings (ruvector-vwm-wasm):
- Browser-ready wasm-bindgen wrappers for all core types
- WasmGaussian4D, WasmDrawList, WasmCoherenceGate, WasmEntityGraph
- WasmLineageLog, WasmActiveMask, WasmBandwidthBudget

WebGPU viewer (examples/vwm-viewer):
- WGSL shaders for Gaussian splatting with alpha blending
- CPU-side projection, depth sorting, and active mask filtering
- Orbit camera controls
- Synthetic demo data generator
- Time scrubber UI with FPS counter and entity search

Zero external dependencies in core crate for full WASM compatibility.
Both crates compile cleanly against the workspace.

https://claude.ai/code/session_012MQauGiqSnQbszfmFKpsNT


ruvnet (commented):
## Code Review: Visual World Model (VWM) — PR #155

**Scope**: 35 new files, 12,738 additions across core Rust crate, WASM bindings, 5 ADRs, WebGPU viewer example.

**Build**: Compiles cleanly. All CI checks pass (5 platforms). **All 169 tests pass** (130 unit, 28 integration, 10 acceptance, 1 doc-test).

---

### Architecture Assessment

The five ADRs (018-022) form a clean dependency chain: ADR-018 (foundation) → ADR-019 (loop cadences) → ADR-020 (GNN feedback) → ADR-021 (attention levels) → ADR-022 (query-first rendering). The implementation faithfully represents the three-loop architecture, 4D Gaussian primitives, packed draw list protocol, and coherence gate. Zero runtime dependencies in the core crate — excellent for WASM compatibility.

**Strong points**: Explicit invariants ("the world model is the source of truth; the splats are a view of it"), concrete latency budgets (12ms fast/500ms medium/10s slow), graceful degradation design, no unsafe code anywhere.

---

### Blocking Issues (3)

**B1. Incorrect Jacobian cross-term in Gaussian projection** (`gaussian.rs:170-178`)
The 2D covariance cross-term `cov2d_b` is computed twice with different formulations then averaged. This does not correspond to any correct derivation of `J * Σ * J^T`. The two formulations give different results because they use different rows of the intermediate product. The correct answer is one or the other, not the average. The `let _ = t3;` and `let _ = cov2d_b;` suppressing unused-variable warnings confirm the author knew these values were suspicious. This produces incorrect screen-space Gaussian shapes.

**B2. Panic risk in `decode_quantized()`** (`tile.rs:382-438`)
No bounds checks on `self.data` before array indexing. Since `PrimitiveBlock` and its `data` field are both `pub`, external code can construct blocks with truncated/corrupted data and trigger panics. The `decode_raw()` path has a length guard but `decode_quantized()` does not.

**B3. `bindTile`/`drawBlock` string-as-u32 bug in viewer** (`examples/vwm-viewer/src/main.js:262,267`)
```javascript
drawList.bindTile(0, 'main-block', 0);  // 'main-block' → u32 = NaN → 0
drawList.drawBlock('main-block', animTime, activeCount > 0 ? 0 : 1);
```
The Rust binding expects `u32` for `block_ref`. `wasm-bindgen` coerces the string to `NaN` → `0`. Works by accident but silently corrupts the draw list data.

---

### Major Issues (6)

**M1. Per-frame tile decoding in layer system** (`layer.rs:157-183`)
`active_count_at()` and `dynamic_active_mask_at()` call `tile.primitive_block.decode()` for every dynamic tile on every invocation. At 60Hz this decodes all dynamic Gaussians every frame. Decoded Gaussians should be cached.

**M2. `queryByType` return format mismatch in viewer** (`main.js:153-161`)
WASM returns entity IDs (numbers) but JS expects entity objects with `embedding` fields. The `JSON.parse(entity.embedding || '{}')` path always fails silently, making the WASM entity graph search non-functional. It works only because the fallback label substring match covers the same cases.

**M3. Coherence gate result not properly mapped in viewer** (`main.js:233-237`)
The gate returns decision strings ("accept"/"defer"/"freeze"/"rollback") but the code treats any truthy string as "coherent". Should be `result === 'accept' ? 'coherent' : 'degraded'`.

**M4. Duplicate FNV implementations with different algorithms** (`tile.rs:535` vs `draw_list.rs:215`)
`tile.rs` uses multiply-then-xor (FNV-1), `draw_list.rs` uses xor-then-multiply (FNV-1a). Both comments say "FNV" but they are different hash algorithms.

**M5. WASM time-range API gap**
`addObject`/`addTrack` hardcode `time_span` to `[NEG_INFINITY, INFINITY]` and `addEdge` always sets `time_range: None`. The core crate extensively supports time-range queries (tested in integration tests) but this capability is unreachable from JS.

**M6. Missing WASM API surface for core pipeline**
The `attention`, `query`, `layer`, `runtime`, `tile` modules (ADR-021/022 higher-level orchestration) have no WASM bindings. Without `Gaussian4D::project()` and `ScreenGaussian`, the viewer must re-implement projection in JavaScript.

---

### Moderate Issues (8)

| # | File | Issue |
|---|------|-------|
| 1 | `tile.rs` | `QuantTier::Warm7/Warm5/Cold3` all silently fall back to Hot8 8-bit encoding |
| 2 | `draw_list.rs` | No `from_bytes()` deserialization despite "network transport" documentation |
| 3 | `entity.rs` | No edge deduplication; `edge_count()` counts duplicates |
| 4 | `entity.rs` | `top_k_by_embedding` is O(N log N) — should use heap for O(N log k) |
| 5 | `attention.rs` | Frustum culling is point-only (ignores Gaussian spatial extent), causes popping |
| 6 | `runtime.rs` | `poll()` eagerly marks last-tick time before caller confirms execution |
| 7 | `layer.rs` | `total_gaussians` field can drift from actual tile counts (no remove/update) |
| 8 | `streaming.rs` | Packet types lack serialization despite "network transport protocol" design |

---

### ADR Consistency Notes

1. **ADR-018 defines 4 loops; ADR-019 collapses to 3** — the "prediction loop" has no explicit home in the three-cadence model
2. **ADR-020 vs implementation gap** — ADR-020 describes GNN-based calibrated coherence; implementation uses simpler fixed-threshold model (acceptable as Phase 1, but should be noted)
3. **ADR-022 `select_active_blocks` truncates by block count, not Gaussian count** — can exceed the budget since blocks contain variable numbers of Gaussians

---

### Test Quality

Tests are exceptionally well-documented and thorough. Notable gaps:
- No tests for `TileMerged`, `EntityAdded`/`EntityUpdated` lineage events
- No test for `SameIdentity` edge type
- No lineage benchmarks (append-only log will grow over time)
- Timing-based acceptance tests could be flaky on slow CI runners
- WASM `js_name` inconsistency — some methods are camelCase, others snake_case

---

### Security

- No unsafe code anywhere — excellent
- No XSS vectors in the viewer (uses `textContent` exclusively)
- `Provenance::signature` field is never verified — provides no integrity guarantee
- Nearly all structs have all-public fields — external code can construct invalid states triggering panics in decode paths

---

### Summary

| Severity | Count |
|----------|-------|
| Blocking | 3 |
| Major | 6 |
| Moderate | 8 |
| Minor | ~15 |

**The core architecture is sound and well-implemented.** The three blocking issues (Jacobian math, decode panic, viewer type bug) should be fixed before merge. The major issues are real but non-blocking — they represent API gaps and viewer bugs that can be addressed in follow-up PRs.

**Recommended action**: Fix B1-B3, then merge. Track M1-M6 as follow-up issues.

PR #217, opened 2026-02-27
fix: resolve MCP -32601 error with Continue extension & fix CLI compilation

AmineC95:
### Description
This PR fixes an initialization error when using `ruvector-mcp` with the [Continue](https://continue.dev/) VS Code extension, and resolves a missing feature compilation issue for the CLI.

### Changes Made
1. **Added `resources/templates/list` handler:** The Continue extension calls this MCP method on startup. Since it was missing, it threw a `-32601 Method not found` error in the editor. I added a handler in `handlers.rs` that returns an empty `resourceTemplates` array to satisfy the client and prevent the error popup.
2. **Fixed `tokio` dependency in CLI:** Added the `io-std` feature to `tokio` in `crates/ruvector-cli/Cargo.toml` to resolve a `cannot find function stdin in module tokio::io` error when building the MCP server standalone.

### Testing
- Compiled successfully on Windows using `cargo build --release -p ruvector-cli --bin ruvector-mcp`.
- Tested locally with the Continue VS Code extension: the MCP server now initializes cleanly without any error prompts and connects successfully.


ruvnet (commented):
Thanks @AmineC95! Clean fix for the MCP -32601 error.

- Adding the `resources/templates/list` handler with an empty response is the correct approach per the MCP spec
- The `tokio` `io-std` feature addition is needed for stdin transport

Looks good, no concerns.

PR #360, opened 2026-04-18
fix(hooks): use $CLAUDE_PROJECT_DIR instead of hardcoded devcontainer path

### LINKED ISSUE #359: Hardcoded `/workspaces/ruvector` paths in .claude/settings.json break hooks outside devcontainer
## Problem

`.claude/settings.json` hardcodes `/workspaces/ruvector/` for all 9 hook commands and the statusline. That path only exists inside the project's Codespaces/devcontainer environment. Anyone cloning the repo to a normal local path (e.g. `~/code/ruvector`) sees every hook fail on startup:

```
SessionEnd hook [node /workspaces/ruvector/.claude/helpers/hook-handler.cjs session-end] failed:
Error: Cannot find module '/workspaces/ruvector/.claude/helpers/hook-handler.cjs'
    code: 'MODULE_NOT_FOUND'
```

## Why this matters

1. **Every Claude Code session is broken for non-devcontainer contributors.** Nine hooks fail on every prompt, session start, and session end — `PreToolUse`, `PostToolUse`, `UserPromptSubmit`, `SessionStart` (×2), `SessionEnd`, `Stop`, `SubagentStart`, plus the `statusLine` command.
2. **The claude-flow V3 features advertised in `CLAUDE.md` silently don't work.** Auto-memory sync, session restore, prompt routing, post-edit learning, status tracking, and the statusline all depend on these hooks. New contributors get a broken experience and no signal that the project's coordination layer isn't actually running.
3. **It undermines the project's own value proposition.** RuVector/claude-flow ships these hooks as a demo of its hooks-automation system. When the reference project itself has broken hooks on any non-Codespaces machine, it's a bad first impression.
4. **Noisy errors pollute every session.** Users see `MODULE_NOT_FOUND` stack traces on every turn, making real errors harder to spot.

## Fix

Replace every occurrence of `/workspaces/ruvector` with `$CLAUDE_PROJECT_DIR` — an env var Claude Code sets automatically to the project root at hook-firing time. Works identically in Codespaces, devcontainers, and local clones.

PR incoming.

## Reproduction

1. Clone `ruvnet/ruvector` to any path other than `/workspaces/ruvector`.
2. Run `claude` in the repo.
3. Observe `MODULE_NOT_FOUND` errors on every hook event.

shaal:
## Summary

- Replace hardcoded `/workspaces/ruvector/` with `$CLAUDE_PROJECT_DIR` in all 9 hook commands and the `statusLine` command in `.claude/settings.json`.
- `$CLAUDE_PROJECT_DIR` is set automatically by Claude Code to the project root at hook-firing time, so it works identically in Codespaces, devcontainers, and local clones.

## Why

The hardcoded path only resolves inside this project's Codespaces/devcontainer. On any normal local clone, every hook fails on startup with `MODULE_NOT_FOUND`, silently disabling the entire claude-flow V3 coordination layer (`UserPromptSubmit` routing, `SessionStart`/`SessionEnd` restore, `PostToolUse` learning, `auto-memory` sync, statusline — 9 hooks total).

This means every non-Codespaces contributor:
1. Sees noisy `MODULE_NOT_FOUND` stack traces on every turn.
2. Runs with the project's advertised coordination layer silently broken.
3. Gets a poor first impression of the very hooks-automation system RuVector ships.

Fixes #359

## Test plan

- [x] `echo '{}' | CLAUDE_PROJECT_DIR=$(pwd) node .claude/helpers/hook-handler.cjs session-end` → exits 0
- [x] Same manual check run for all 5 `hook-handler.cjs` subcommands (`pre-bash`, `post-edit`, `route`, `session-restore`, `session-end`, `status`) → all exit 0
- [x] `auto-memory-hook.mjs import` and `sync` → exit 0
- [x] `statusline.cjs` → renders status line correctly
- [x] Verified in a live Claude Code session on a local clone at `~/code/utilities/ruvector` — the previously-failing `SessionEnd` and `UserPromptSubmit` hooks now fire cleanly
- [ ] Maintainer verifies hooks still fire correctly inside Codespaces/devcontainer (should be a no-op since `$CLAUDE_PROJECT_DIR` resolves to `/workspaces/ruvector` there)

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node /workspaces/ruvector/.claude/helpers/hook-handler.cjs pre-bash",
+            "command": "node $CLAUDE_PROJECT_DIR/.claude/helpers/hook-handler.cjs pre-bash",
             "timeout": 5000
           }
         ]
@@ -18,7 +18,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node /workspaces/ruvector/.claude/helpers/hook-handler.cjs post-edit",
+            "command": "node $CLAUDE_PROJECT_DIR/.claude/helpers/hook-handler.cjs post-edit",
             "timeout": 10000
           }
         ]
@@ -29,7 +29,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node /workspaces/ruvector/.claude/helpers/hook-handler.cjs route",
+            "command": "node $CLAUDE_PROJECT_DIR/.claude/helpers/hook-handler.cjs route",
             "timeout": 10000
           }
         ]
@@ -41,13 +41,13 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node /workspaces/ruvector/.claude/helpers/hook-handler.cjs session-restore",
+            "command": "node $CLAUDE_PROJECT_DIR/.claude/helpers/hook-handler.cjs session-restore",
             "timeout": 15000,
             "continueOnError": true
           },
           {
             "type": "command",
-            "command": "node /workspaces/ruvector/.claude/helpers/auto-memory-hook.mjs import",
+            "command": "node $CLAUDE_PROJECT_DIR/.claude/helpers/auto-memory-hook.mjs import",
             "timeout": 8000,
             "continueOnError": true
           }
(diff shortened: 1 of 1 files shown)


shaal:
Closing —  already has this fix (via commits 7de65cc1, a8969348, c39dccb7), and in a better form: `"${CLAUDE_PROJECT_DIR:-.}"` with quoting and a fallback. My branch was forked from an older base commit so I didn't see the upstream fix. Thanks!

Issue #402, opened 2026-04-27
ruvector@0.2.23: data-plane broken — VectorDB CRUD, GNN, attention, embed, rvf, benchmark all fail


ronmikailov:
## Summary

After exhaustive testing of every top-level subcommand in `ruvector@0.2.23`, **the core data plane is non-functional**. Failures cluster into 3 root causes, all in scope of the existing release-hygiene tracker (#399, #400, #401):

- **A. `VectorDB` API drift** — CLI calls methods that no longer exist on `@ruvector/core`'s native binding.
- **B. GNN / attention N-API binding errors** — typed-array marshalling broken across all GNN ops.
- **C. More missing `dist/` paths** — beyond #399, the CLI also requires `dist/core/onnx-embedder.js` and `dist/core/rvf-wrapper.js`.

CLI-parser, help text, info/status commands, and a few utility commands (decompile, mcp tools, llm embed, identity, brain) are all healthy — so the CLI shell is fine, the binding integration is what's broken.

---

## A. `VectorDB` API drift — entire CRUD lifecycle blocked

The CLI assumes a `VectorDB` API surface (`insertBatch`, on-disk persistence) that doesn't match what `@ruvector/core@0.1.31` actually ships.

### A1. `create` claims success but writes no file

```bash
$ ruvector create -d 4 ./test.db
- Creating database...
✔ Database created: ./test.db
  Dimension: 4
  Metric: cosine
  Implementation: native (@ruvector/core)

$ ls test.db
ls: cannot access 'test.db': No such file or directory   # <-- nothing on disk
```

The "✔" success line is printed but no file is created. Every downstream read/write then fails with `ENOENT`.

### A2. `insert` calls a missing method

```bash
$ ruvector insert ./test.db ./vecs.json
✖ Failed to insert vectors
db.insertBatch is not a function
```

`db.insertBatch` doesn't exist on the current `VectorDB` binding. Same root cause causes:

- `ruvector demo --basic` → `db.insert is not a function` (related — different method, same drift, see #400)
- `ruvector benchmark` → `db.insertBatch is not a function` (same as above)

### A3. `stats` / `search` / `export` / `import` are all blocked

All fail with `ENOENT` because `create` doesn't actually persist. So even before hitting any API drift, the file simply isn't there. After one round-trip, **the entire VectorDB CLI workflow is unusable**.

### Likely fix

Audit the CLI handlers in `bin/cli.js` against the current `@ruvector/core` surface. Either:
[...]

Issue #399, opened 2026-04-27
ruvector@0.2.23: published tarball missing dist/ — `ruvector doctor` falsely reports 'Native binding failed to load'


ronmikailov:
## Summary

`ruvector@0.2.23` declares `"main": "dist/index.js"` and `"types": "dist/index.d.ts"` in `package.json`, and the `files` array also includes `dist/` — but **the published npm tarball ships no `dist/` directory at all**. Only `bin/`, `src/`, `wasm/`, `LICENSE`, `README.md`, `package.json` are present.

This causes:

1. `require('ruvector')` to fail with `MODULE_NOT_FOUND`.
2. `ruvector doctor`'s `loadRuvector()` (`bin/cli.js:26-38`) silently catches that error and falls through to print `! Native binding failed to load` and bumps `warnings++`, even when `@ruvector/core` is installed and its native binding loads cleanly.

The native binding itself is fine — the false negative is purely the doctor's check looking at a path that doesn't exist in the shipped package.

## Reproduction

```bash
npm install -g ruvector@0.2.23 @ruvector/core @ruvector/gnn @ruvector/graph-node @ruvector/node-linux-x64-gnu

# Confirm @ruvector/core's native binding works
node -e "const c = require('@ruvector/core'); console.log(c.hello())"
# → "Hello from Ruvector Node.js bindings!"

# But:
ruvector doctor
# → "! Native binding failed to load"  (false negative)

# Root cause:
ls $(npm root -g)/ruvector/
# → bin/  src/  wasm/  LICENSE  README.md  package.json   (no dist/)
```

## Expected

Either:

- **Ship the build artifact**: run `npm run build` (existing script) before `npm publish`. The `prepublishOnly` hook is already wired to do this — it just isn't producing output that ends up in the tarball.
- **Or update `main`** to point at an actually-shipped entry (e.g. `bin/cli.js`, `src/index.js` if one exists).

Either way, the doctor's `Native binding working` check should reflect the real state of `@ruvector/core` rather than the presence/absence of a top-level `dist/index.js` re-export.

## Impact

Anyone running `ruvector doctor` (the documented health-check command) gets a misleading warning that drives unnecessary debugging — even on a fully working install. Patched locally on our side with a shim, but it's a clean upstream fix.

## Local Workaround (for reference)

```js
// $(npm root -g)/ruvector/dist/index.js
const core = require('@ruvector/core');
module.exports = {
  VectorDB: core.VectorDB || class VectorDB {},
[...]

PR #405, opened 2026-04-27
fix(ruvector-cli): bundle ONNX runtime assets in dist/ + extend verify-dist gate

ronmikailov:
## Summary

`0.2.23` → `0.2.25` ship `dist/core/onnx/pkg/` containing only `package.json`. No `loader.js`, no wasm-bindgen JS bridge, no `.wasm` binary. The embedder reads those at startup via `path.join(__dirname, 'onnx', ...)`, so on a clean install of any of those versions:

```
$ npx ruvector embed text "hello"
Embedding failed: Failed to initialize ONNX embedder:
  ONNX WASM files not bundled. The onnx/ directory is missing.
```

This PR makes the build/publish actually ship what it claims to ship.

## Root cause

Old build script:
```json
"build": "tsc && cp src/core/onnx/pkg/package.json dist/core/onnx/pkg/"
```

`tsc` only emits `.js` for `.ts` inputs. The JS loader, the wasm-bindgen bridge, and the `.wasm` binary itself are all non-TS, so all three were silently skipped. Only the single-file `cp` ran.

The `verify-dist` gate added in #403 caught the original `dist/` regression but didn't catch this one because it only scans `bin/cli.js` for `require('../dist/...js')`. These assets are read via `path.join` + `fs.readFileSync` + `dynamicImport`, not `require()`, so the original gate is structurally blind to them.

### 1. `scripts/copy-onnx-assets.js` (new)

Explicit, cross-platform Node script that copies the 8 runtime assets the embedder needs. Listed explicitly (not a recursive `cp`) so adding a new asset is a deliberate choice, not an accident. Fails the build with a non-zero exit if any source asset is missing.

### 2. `package.json`

```diff
- "build": "tsc && cp src/core/onnx/pkg/package.json dist/core/onnx/pkg/"
+ "build": "tsc && node scripts/copy-onnx-assets.js"
```

### 3. `scripts/verify-dist.js`

Adds a second pass that asserts the four runtime asset paths read by `dist/core/onnx-embedder.js` actually exist. Same publish-gate semantics as the existing `bin/cli.js` scan, just covering the path-join'd asset reads as well.

## Verification

```
$ rm -rf dist && npm run build
copy-onnx-assets: 8 asset(s) copied to dist/core/onnx/.

$ npm run verify-dist
verify-dist: 13 dist require path(s) + 4 runtime asset(s) present.

$ ls dist/core/onnx/pkg/
LICENSE
package.json
ruvector_onnx_embeddings_wasm.d.ts
ruvector_onnx_embeddings_wasm.js
ruvector_onnx_embeddings_wasm_bg.js
ruvector_onnx_embeddings_wasm_bg.wasm
ruvector_onnx_embeddings_wasm_bg.wasm.d.ts
```

The "ONNX WASM files not bundled" error no longer fires — the `existsSync` check at `dist/core/onnx-embedder.js:184` passes because the files exist.

## Out of scope (followup needed)

With all assets present, Node still fails the next step:

```
Failed to initialize ONNX embedder: Unknown file extension ".wasm"
for .../dist/core/onnx/pkg/ruvector_onnx_embeddings_wasm_bg.wasm
```

The shipped `ruvector_onnx_embeddings_wasm.js` is the wasm-bindgen **bundler** target (`import * as wasm from "./xxx.wasm"`). Node ESM rejects bare `.wasm` imports without `--experimental-wasm-modules`. The embedder code expects the **web/no-modules** target (with `wasmModule.default(wasmBytes)` for deferred init), but it never gets that branch because the bundler-target import fails before the embedder's branch logic runs.

The clean fix is to rebuild `ruvector-onnx-embeddings-wasm` with `--target nodejs` (or `web` with deferred init) so the entry exposes a `default(bytes)` initializer and doesn't try to `import` the `.wasm` directly. That's a Rust/wasm-bindgen build-config change, not a JS one — out of scope for this PR.

This PR is the necessary precondition for that work: the assets actually need to ship before any wasm-target fix can possibly run end-to-end.

## Related

- #399 — `dist/` missing entirely on 0.2.23. Fixed in #403.
- #402 §C — additional `dist/` paths (`onnx-embedder.js`, `rvf-wrapper.js`) missing. Fixed in #403.
- This PR — the **non-`.js`** assets that `dist/core/onnx-embedder.js` reads at runtime were never copied by the build, even after #403's `dist/` rescue. Same release-hygiene class.

### DIFF
--- a/npm/packages/ruvector/package.json
+++ b/npm/packages/ruvector/package.json
@@ -8,7 +8,7 @@
     "ruvector": "./bin/cli.js"
   },
   "scripts": {
-    "build": "tsc && cp src/core/onnx/pkg/package.json dist/core/onnx/pkg/",
+    "build": "tsc && node scripts/copy-onnx-assets.js",
     "verify-dist": "node scripts/verify-dist.js",
     "prepublishOnly": "npm run build && npm run verify-dist",
     "test": "node test/integration.js && node test/cli-commands.js"
--- a/npm/packages/ruvector/scripts/copy-onnx-assets.js
+++ b/npm/packages/ruvector/scripts/copy-onnx-assets.js
@@ -1,0 +1,64 @@
+#!/usr/bin/env node
+/**
+ * copy-onnx-assets.js — copy the bundled ONNX runtime files from src/core/onnx
+ * into dist/core/onnx after `tsc`.
+ *
+ * Why: tsc only emits .js for .ts inputs. The ONNX subsystem ships a JS
+ * loader, a wasm-bindgen JS bridge, and the .wasm binary itself — none of
+ * which are TypeScript and so all of which are skipped by tsc. The previous
+ * build script copied only `pkg/package.json`, which left embed text and
+ * the embed-stream pipeline broken with:
+ *   "Failed to initialize ONNX embedder: ONNX WASM files not bundled.
+ *    The onnx/ directory is missing."
+ *
+ * This script does an explicit, cross-platform copy of every runtime asset
+ * the embedder reads via `path.join(__dirname, 'onnx', ...)` in
+ * src/core/onnx-embedder.ts. It is intentionally explicit (not a recursive
+ * dir copy) so a new file gets noticed and its inclusion is a deliberate
+ * decision, not an accident.
+ */
+
+const fs = require('fs');
+const path = require('path');
+
+const pkgRoot = path.resolve(__dirname, '..');
+const srcDir = path.join(pkgRoot, 'src', 'core', 'onnx');
+const dstDir = path.join(pkgRoot, 'dist', 'core', 'onnx');
+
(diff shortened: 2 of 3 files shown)


ruvnet (approved):
Approved. Tight scope, root cause clearly identified, and the regression chain (#399 → #403 → #405) is documented end-to-end. The explicit asset list (vs `cp -r`) is the right call — future asset additions become deliberate. Extending `verify-dist` to cover the `path.join` + `fs.readFileSync` reads closes the gap that the original `require()`-only scan was structurally blind to.

The wasm-bindgen target rebuild (`Unknown file extension ".wasm"`) is correctly identified as the next step — I'll track that as a followup. This PR is the necessary precondition.

Thanks for the careful diagnosis and verification output.

— rUv

PR #406, opened 2026-04-28
Fix Windows build (workspace tokio io-std) + add Windows release workflow

namesreallyblank:
## Summary

Two small, focused fixes that together unblock building and shipping `ruvector-mcp` on Windows (and incidentally repair builds on every other platform).

### 1. `fix(cli): add io-std to workspace tokio features` (commit fc51094)

`crates/ruvector-cli/src/mcp/transport.rs` calls `tokio::io::stdin()` and `tokio::io::stdout()` (lines 30-31), but the workspace tokio dep in the root `Cargo.toml` omits the `io-std` feature. Since `ruvector-cli` inherits tokio via `workspace = true`, the build fails everywhere from a clean checkout:

```
cargo build --release -p ruvector-cli --bin ruvector-mcp
error[E0425]: cannot find function `stdin` in module `tokio::io`
error[E0425]: cannot find function `stdout` in module `tokio::io`
```

Adding `io-std` to the workspace dep fixes every consumer at once (no per-crate overrides). Verified locally: `cargo build --release -p ruvector-cli --bin ruvector-mcp` now finishes cleanly on macOS aarch64 (release profile, ~98s on M-series).

> Note: PR #217 also adds `io-std` (at the package level inside `crates/ruvector-cli/Cargo.toml`) but is bundled with unrelated `.claude/` agent file changes and an MCP `resources/templates/list` handler. This PR keeps the fix minimal and applies it at the workspace level so any future crate using stdio gets it for free. Happy to close this and land #217 if maintainers prefer that route.

### 2. `ci: add Windows release workflow for ruvector-mcp` (commit a11e617)

Adds `.github/workflows/release-windows.yml`. The repo has 27 existing workflows but none produces a Windows binary for `ruvector-mcp`. With this workflow:

- Pushing a `v*` tag → builds `ruvector-mcp.exe` on `windows-latest` (MSVC) and attaches it to the GitHub Release.
- `workflow_dispatch` → uploads the .exe as a 30-day artifact for manual smoke tests.

Uses `dtolnay/rust-toolchain@stable` with `x86_64-pc-windows-msvc`, `actions/cache@v4` over the cargo dirs, and `softprops/action-gh-release@v2` for the tag-triggered upload. Output asset is `ruvector-mcp-x86_64-pc-windows-msvc.exe` with a SHA256 printed in the build log.

Depends on commit 1 — without `io-std` the workflow's `cargo build` step would fail.

## Motivation

This came out of a downstream Windows port effort (Clorch, an OSS orchestration project that consumes `ruvector-mcp` as a memory backend). Without these two fixes, Windows users either have to cross-compile locally (which silently fails on the same `io-std` regression) or skip RuVector entirely. With them, a single tag push publishes a ready-to-download `.exe`.

## Test plan

- [x] Local: `cargo build --release -p ruvector-cli --bin ruvector-mcp` on macOS aarch64 succeeds (verified before opening PR).
- [ ] CI: existing workflows still pass on this branch.
- [ ] CI: the new `Release Windows Binary` workflow can be triggered via `workflow_dispatch` and produces the artifact.
- [ ] Optional: tag a test release (e.g. on a fork) to confirm the .exe attaches to the Release page.

No runtime behaviour changes — the io-std feature flag only affects what symbols are compiled into tokio.


ruvnet:
Thanks for sending this — the diagnosis is correct and the intent is good, but the PR can't go in as-is. Detailed feedback below.

## State of this PR

GitHub shows **7,313 files changed / +3,187,916 / −30,706, mergeable: CONFLICTING, 0 CI checks**. The branch is forked from a base so old it predates ruvbot, ADR-042/043, RVF v3.0, and dozens of NAPI binary refreshes. The actual diff is dominated by upstream work neither of us touched. Before any merge discussion this needs a clean rebase onto current `main` (or a fresh branch with just the 2 intended commits cherry-picked).

I reviewed your two intended commits in isolation:

## Commit `fc51094` — `Cargo.toml` add `io-std` ✅

Verified at HEAD:
- `Cargo.toml:250` — workspace tokio is missing `io-std`
- `crates/ruvector-cli/src/mcp/transport.rs:30-31` — calls `tokio::io::stdin()` / `tokio::io::stdout()`

Build genuinely fails from a clean checkout for `ruvector-mcp`. Fix is one line, workspace-level scope is correct (vs PR #217's per-package fix). **This should land immediately on its own** — it's a one-line bug fix that unblocks builds on every platform, not just Windows. Please open a separate PR with just `fc51094` and we'll merge it right away.

## Commit `a11e617` — `release-windows.yml` — needs rework

Five things to address before merge:

1. **Stale header comments.** The 20-line preface still describes this file as "NOT used by the Clorch repo… template for contributors to upstream into ruvnet/RuVector." Once the file lives in this repo, those comments are wrong and confusing. Strip to a one-line description.

2. **Tag pattern is too broad.** Trigger is `tags: ['v*']`. The neighboring `release-rvf-cli.yml` correctly scopes itself with `rvf-v*`. A bare `v*` will fire this workflow on every release tag across the repo, including ones unrelated to `ruvector-mcp`. Use `ruvector-mcp-v*` (or similar).

3. **Asymmetric platform coverage.** `release-rvf-cli.yml` already builds a 5-target matrix (Linux x64/arm64, macOS x64/arm64, Windows MSVC) for `rvf`. A single-target Windows-only workflow for `ruvector-mcp` leaves Linux/macOS users with no published binary at all. Strongly prefer a `release-ruvector-mcp.yml` that mirrors the existing matrix structure — costs little, fixes the asymmetry.

4. **Missing SHA256 sidecar.** Hash is printed to logs but not uploaded next to the binary. For Clorch's verification use case, please attach a `*.sha256` file to the release alongside the `.exe`.

5. **Action pinning.** `dtolnay/rust-toolchain@stable` is a moving tag. Same as `release-rvf-cli.yml` so not a regression — just a note for future hardening.

## Recommended path forward

Please split this into two PRs against current `main`:

- **PR A:** just `fc51094` (workspace `io-std`). One-line fix, lands today.
- **PR B:** the workflow, after the rework above. Easier to iterate without dragging the io-std fix behind it.

Re #217: workspace-level is the right place for the `io-std` fix, so your approach wins on that one.

Thanks for the careful PR description and for tracking down the root cause across both layers — once split, this lands fast.

— rUv

namesreallyblank:
Closing in favor of split PRs per the maintainer review:

- #407 — `fix(cli): add io-std to workspace tokio features` (the standalone Cargo.toml fix, ready to merge as-is)
- #408 — `ci: add multi-platform release workflow for ruvector-mcp` (release workflow rework: 5-target matrix mirroring `release-rvf-cli.yml`, scoped `ruvector-mcp-v*` tag pattern, SHA256 sidecars, one-line header)

Both branches were cut fresh from current `main` (commit `20aca12a`) to drop the 7,313-file phantom diff that came from #406's stale base. Thanks for the detailed review @ruvnet.

PR #501, opened 2026-05-23
feat(core): add DeterministicScore with deterministic fusion

ohdearquant:
## Summary

- Add `DeterministicScore` (i64 fixed-point, 2^32 scale) to `ruvector-core` — ported from [khive-score](https://github.com/ohdearquant/khive/tree/main/crates/khive-score)
- Fix NaN panic in `FlatIndex` sorting (`partial_cmp().unwrap()` → total ordering)
- Change RRF default K from 60 to 15 (Lean4 proof: K=60 has consensus bias)
- Add `search_deterministic()` to `VectorIndex` trait — boundary conversion from f32 distance to deterministic similarity
- Add `deterministic_rrf` and `deterministic_weighted` fusion using `BTreeMap` for iteration-order determinism
- 238 tests pass, 0 regressions

## Lean4 formal proofs

Corresponding formal proofs (51 theorems, zero `sorry`) verified with Lean 4.25.2 + Mathlib v4.25.2:

- **Proofs repo**: [ohdearquant/lean-proofs](https://github.com/ohdearquant/lean-proofs)
- **Paper package**: [papers/2026-deterministic-score](https://github.com/ohdearquant/lean-proofs/tree/main/papers/2026-deterministic-score)

Properties proved: total order, NaN safety, saturating arithmetic (branch specs + commutativity), RuntimeValid invariant preservation, abstract float conversion validity, MIN-unreachability, scalar mul/div validity, division-by-zero policy.

## Test plan

- [x] All 19 `deterministic_score` tests pass
- [x] Full `ruvector-core` suite: 238 passed, 0 failed
- [x] Clippy clean (no new warnings)
- [ ] Downstream: khive-retrieval integration (Phase 5, separate PR)

### DIFF
--- a/crates/ruvector-core/src/advanced_features/sparse_vector.rs
+++ b/crates/ruvector-core/src/advanced_features/sparse_vector.rs
@@ -266,8 +266,7 @@ pub struct ScoredDoc {
 /// Strategy for combining ranked lists from different retrieval systems.
 #[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
 pub enum FusionStrategy {
-    /// Reciprocal Rank Fusion. `k` controls rank-pressure (default 15).
-    /// K=15 per Lean4 proof `ret010_k60_consensus_bias` — K=60 has consensus bias.
+    /// Reciprocal Rank Fusion. `k` controls rank-pressure (default 60).
     RRF { k: f32 },
     /// Weighted linear combination of normalised scores.
     Linear {
@@ -281,7 +280,7 @@ pub enum FusionStrategy {
 
 impl Default for FusionStrategy {
     fn default() -> Self {
-        FusionStrategy::RRF { k: 15.0 }
+        FusionStrategy::RRF { k: 60.0 }
     }
 }
 
--- a/crates/ruvector-core/src/advanced_features/sparse_vector.rs
+++ b/crates/ruvector-core/src/advanced_features/sparse_vector.rs
@@ -266,7 +266,8 @@ pub struct ScoredDoc {
 /// Strategy for combining ranked lists from different retrieval systems.
 #[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
 pub enum FusionStrategy {
-    /// Reciprocal Rank Fusion. `k` controls rank-pressure (default 60).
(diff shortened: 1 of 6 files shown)


ohdearquant:
Coming back to this after a long gap, and the first thing I owe is a correction to my own description above, because one of its four bullets is no longer true of `main`.

**"Fix NaN panic in `FlatIndex` sorting (`partial_cmp().unwrap()` → total ordering)" is stale.** `crates/ruvector-core/src/index/flat.rs:69` on current `main` already reads:

```rust
results.sort_unstable_by(|a, b| a.1.partial_cmp(&b.1).unwrap_or(std::cmp::Ordering::Equal));
```

That landed independently of this PR. So that bullet describes a problem someone else already solved, and anyone reading this PR description today would be misled by it. Struck.

What that leaves is three things of quite different sizes, bundled into one 650-line change, which is probably why it has been easy to leave alone:

1. `DeterministicScore` (i64 fixed-point) plus the deterministic fusion functions — a substantial new type and a new trait method on `VectorIndex`.
2. Changing the RRF default K from 60 to 15 — a behaviour change to existing callers, independent of (1).
3. The `FlatIndex` ordering fix — now moot.

Bundling (2) with (1) was a mistake on my part. A default-value change to a ranking function deserves its own argument and its own decision, and hiding it inside a PR whose headline is a new numeric type makes it easy to merge by accident or reject by association.

I am happy to do whichever of these you prefer:

- **Split and rebase.** Open (1) on its own against current `main`, and (2) separately with the case for K=15 stated on its own merits so it can be accepted or declined without dragging the type with it.
- **Close it.** If a fixed-point deterministic scoring type is not a direction RuVector wants, say so and I will close this rather than leave a conflicting PR sitting on your list. No hard feelings, and it costs me nothing to have asked.

Either way this should not stay as it is. It conflicts with `main`, and it is now carrying a claim that does not hold.

One note on the substance in case it affects the answer. The motivation for (1) is that f32 similarity scores make ranking non-reproducible across platforms and across fusion orderings: the same corpus and the same query can rank differently depending on accumulation order, which is awkward if you need a stable answer for evaluation or for auditing. Fixed-point scoring plus `BTreeMap`-ordered fusion makes the ranking a function of the inputs alone. Whether that is worth a new type in `ruvector-core` is genuinely your call, not mine.

Issue #523, opened 2026-05-30
API contract issues in ONNX embedder: isOnnxAvailable/isReady lie about state, double model load, AdaptiveEmbedder.isReady() returns undefined


ruvnet:
## Summary

Found four API contract issues in ruvector's ONNX embedder while building a BEIR public-benchmark harness for [ruvnet/ruflo](https://github.com/ruvnet/ruflo) (see [ADR-085](https://github.com/ruvnet/ruflo/blob/main/v3/docs/adr/ADR-085-beir-public-benchmark.md), [ADR-086](https://github.com/ruvnet/ruflo/blob/main/v3/docs/adr/ADR-086-silent-fallback-and-bootstrap.md)). All four make it harder for downstream packages to use ruvector's bundled embedder confidently — they had us silently falling back to hash embeddings for the entire BEIR run without realising.

## Versions

- `ruvector@0.2.29` (or current latest via npm)
- macOS darwin-arm64, Node 22.22.1
- Probe script: [gist below](#repro)

### 1. `isOnnxAvailable()` returns `true` before `initOnnxEmbedder()` is called

```js
const r = await import('ruvector');
console.log(r.isOnnxAvailable());  // → true
// but no embedder has been initialised yet
```

Expected: `false` until `initOnnxEmbedder()` resolves successfully.

Impact: callers (like ruflo's `neural-tools.ts`) use this as a "should I use ruvector for embedding?" gate. Returning `true` prematurely makes callers commit to a path that may then lazy-fail.

### 2. `getOptimizedOnnxEmbedder().isReady()` returns `false` but `embed()` still works

```js
await r.initOnnxEmbedder();   // succeeds
const emb = r.getOptimizedOnnxEmbedder();
console.log(emb.isReady());   // → false
const v = await emb.embed('test');  // → works, returns valid 384-dim vector
console.log(emb.isReady());   // still false
```

Expected: after a successful `embed()`, `isReady()` should be `true`. Or document that `embed()` is auto-initialising and `isReady()` reflects only pre-emptive readiness.

Impact: same as #1 — callers can't use `isReady()` as a gate. The probe must be "try a real embed and catch."

### 3. Double model load — `initOnnxEmbedder()` + first `embed()` both trigger downloads

Output of the probe script (full output below):

```
Loading ONNX model: all-MiniLM-L6-v2...
  Downloading: https://huggingface.co/.../model.onnx
ONNX embedder ready: 384d
initOnnxEmbedder OK: 3172 ms

[...]

PR #533, opened 2026-06-03
fix(ruvector): make createEmbedder() init WASM correctly under Node (…

as473:
…#523)

The bundler-target wasm-pack wrapper (pkg/ruvector_onnx_embeddings_wasm.js) starts with `import * as wasm from "...bg.wasm"` (unresolvable in plain Node ESM) and exports no `default` init, so createEmbedder()'s old `import(wrapper); await default()` path threw `m.default is not a function` under Node.

Mirror the proven worker init (embed-worker.mjs): import the _bg.js glue directly and instantiate the .wasm via WebAssembly.instantiate, Node-guarded. The browser/bundler default() path is preserved. Vectors are identical to the worker path by construction.

### DIFF
--- a/npm/packages/ruvector/src/core/onnx/loader.js
+++ b/npm/packages/ruvector/src/core/onnx/loader.js
@@ -365,10 +365,34 @@ export class ModelLoader {
  * ```
  */
 export async function createEmbedder(modelName = DEFAULT_MODEL, wasmModule = null) {
-    // Import WASM module if not provided
+    // Import + initialize the WASM module if not provided.
     if (!wasmModule) {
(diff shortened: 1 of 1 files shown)


ruvnet:
Reviewed. Fix is correct — mirrors the proven `embed-worker.mjs` init pattern exactly (`pathToFileURL` + `WebAssembly.instantiate` + `__wbg_set_wasm`). Root cause analysis is accurate: the bundler-target wrapper has a bare `.wasm` import that Node ESM cannot resolve. Browser path is preserved unchanged. Approving.

ruvnet (approved):
Fix is correct and minimal. Mirrors the `embed-worker.mjs` init pattern exactly. Approving.

PR #603, opened 2026-06-25
feat(timesfm): TimesFM 1.0 200M decoder-only inference port to candle

ruvnet:
## Summary
Native Rust/`candle` port of **google-research/timesfm** (1.0 200M, `pytorch_patched_decoder.py`) — temporal embeddings + zero-shot forecasting inside RuVector, no Python microservice. Implemented via the Metaharness chunked-iterative-synthesis pipeline (ADR-191 in ruvnet/agent-harness-generator).

New crate `crates/timesfm`, behind an **opt-in `candle` feature** (default `[]`, cpu-fallback pattern mirroring `ruvector-hailo`). **No lockfile churn** — candle 0.9.2 was already pinned by `ruvllm`.

## What's in it
- **config.rs** — `TimesfmConfig`: 1280 model dim, 20 layers, 16 heads, 80 head_dim, input patch 32 / output patch 128, 10 quantile outputs.
- **model.rs** — `ResidualBlock` patch embedding, non-learned sinusoidal positional embedding (added, **not RoPE**), 20× decoder layer (fused `qkv_proj`, learnable per-head-dim `softplus(scaling)` query scale, additive causal+padding mask, MLP with its own LayerNorm+ReLU residual), RevIN per-series instance norm. `forward → [B,N,128,10]` + autoregressive `decode` to arbitrary horizon.
- **scripts/convert_weights.py** — HF safetensors → VarBuilder key remap (`--dry-run`); source keys derived from the timesfm reference, target keys from the actual Rust `vb.pp(...)` calls.

## Verification (gates green)
- `cargo build -p timesfm` (default) + `--features candle` ✅
- `cargo test -p timesfm --features candle` → **12 passed / 0 failed** (shape tests for every module + full forward at real 1280/16/80 widths + arbitrary-horizon decode, **plus 2 RevIN numerical regression tests**)
- `cargo clippy -p timesfm --all-targets --features candle -- -D warnings` → clean

## Adversarial review caught a real bug
`masked_mean_std` (RevIN) was computing a **global** mean/std over the whole context instead of the reference's **first-qualifying-patch** selection — would offset every forecast by a per-series scale/shift. Fixed + regression-tested. Attention scaling, matmul order, causal/padding masks independently verified against the reference.

## Honest scope (not claimed)
- **Real numerical weight-parity is NOT yet verified** — all tests run on `VarBuilder::zeros`/`randn` dummy weights (no safetensors in the build env). End-to-end parity needs loading `google/timesfm-1.0-200m` and comparing against the PyTorch reference. **This is the gating next step before any forecasting claim.**
- Open low-impact faithfulness deviations (documented in code, benign for unpadded/right-aligned inputs): MLP padding-mask omission; pos-emb `_shift_padded_seq`; NaN-on-fully-padded-leading-patch edge (reference uses `-0.7*max` not `-inf`).


ruvnet:
Pushed a rustfmt commit so `crates/timesfm` is fmt-clean + clippy-clean (`cargo fmt -p timesfm -- --check` and `cargo clippy -p timesfm --all-targets --features candle -- -D warnings` both clean; `cargo test -p timesfm --features candle` = 12 passed). Note: the workspace-wide `cargo fmt --all --check` / clippy CI also flags ~133 pre-existing diffs in *other* crates (not touched by this PR) — those are unrelated to TimesFM and out of scope here.

ruvnet:
## ✅ Weight-parity VALIDATED (real `timesfm-1.0-200m` weights)

Loaded the real PyTorch checkpoint (via `google/timesfm-1.0-200m-pytorch`, since the HF main repo ships only a JAX/Orbax checkpoint and PyPI `timesfm` 2.x dropped the 1.x line) through `convert_weights.py` → candle, and compared against the reference `pytorch_patched_decoder.py` on a deterministic 512-pt series, horizon 128:

| metric | measured |
|---|---|
| max-abs-diff | **8.58e-6** |
| MAE | 3.25e-6 |
| relative error | 5.83e-7 |

Agreement at the f32 accumulation-order floor across the full 20-layer stack + autoregressive decode. The bridge mapped all 253 keys 1:1 (no changes needed).

**One real bug found + fixed** (commit a190609): `build_mask` used `f32::NEG_INFINITY` → `0 * -inf = NaN` poisoned the attention mask (all-NaN output). Now uses the reference's finite `-0.7·f32::MAX` + minimum-merge — exact parity. Added `examples/parity.rs` + a gated `tests/parity.rs` (skips cleanly without the 814MB artifacts, asserts < 1e-3, never fabricates a pass). All 12 shape tests + parity test green with `--features candle`.

ruvnet:
## ✅ Phases A+B complete (Darwin integration + 24-case GCP test)

**Darwin predictive-pruning** (`src/prune.rs`, commit 22f1e84): `decide_prune(curve, horizon, threshold)` TimesFM-forecasts a metric curve's tail → PRUNE/CONTINUE. Measured (real weights, K=128, target h=1000, thr=0.05): DOOMED curve → forecast plateau 1.98 → **PRUNE ✓**; HEALTHY curve → already-viable → **CONTINUE ✓**. 14/14 tests green.
> Honest caveat (in the module doc): TimesFM mean-reverts upward on short sharply-decayed synthetic curves, so the *absolute* plateau is biased high — the decision rides the robust *relative* ordering (doomed always forecasts higher) + the already-viable guard, not absolute calibration.

**24-case GCP test** (commit 8d72f86): one e2-standard-4, weights sha256-verified, **24/24 finite**, mean **168 ms** / p95 171.8 ms / 5.95 fps / 1.56 GB RSS (≈4× the ruvultra 42.5ms = the 4-vs-32-thread ratio). VM deleted + confirmed gone. Cost <$0.10.

**Perf is baseline candle-CPU — not yet optimized** (levers: KV-cache for autoregressive decode, MKL/SIMD accel). Flagged honestly, not claimed as optimized.

PR #639, opened 2026-07-04
perf(rls): wrap current_setting() in the generated tenant policy for per-statement evaluation

dmitrymaranik:
`ruvnet/RuVector`'s RLS policy generator emits `current_setting('<tenant setting>')` **unwrapped** in the policy predicate. Postgres re-evaluates a bare `current_setting(...)` **once per row** the policy scans; wrapping it in a scalar subquery — `(select current_setting(...))` — makes it a per-statement InitPlan the planner evaluates once and caches. It's predicate-equivalent (tenant isolation unchanged) and the documented Postgres RLS performance pattern; the benefit grows with table size.

This wraps the generated call and keeps the test assertions in sync (`rls.rs`, `isolation.rs`). I couldn't run the Rust test suite locally — if CI surfaces another assertion on the generated policy SQL, point me at it and I'll update it.

Spotted with [pgrls](https://github.com/pgrls/pgrls) (an open-source Postgres RLS analyzer). Happy to adjust.

### DIFF
--- a/crates/ruvector-postgres/src/tenancy/isolation.rs
+++ b/crates/ruvector-postgres/src/tenancy/isolation.rs
@@ -139,8 +139,8 @@ DROP POLICY IF EXISTS ruvector_admin_bypass ON {table};
 
 -- Create tenant isolation policy
 CREATE POLICY ruvector_tenant_isolation ON {table}
-    USING ({column} = current_setting('ruvector.tenant_id', true))
-    WITH CHECK ({column} = current_setting('ruvector.tenant_id', true));
+    USING ({column} = (select current_setting('ruvector.tenant_id', true)))
+    WITH CHECK ({column} = (select current_setting('ruvector.tenant_id', true)));
 
 -- Create admin bypass policy (for ruvector_admin role)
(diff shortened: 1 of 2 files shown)


ruvnet:
Appreciate the care here given this is tenant-isolation code — verified the `missing_ok=true` arg is preserved at every call site and that the InitPlan-caching wrap doesn't change fail-closed semantics (unset GUC still evaluates to NULL, excluding rows rather than granting access). Merging. Thanks for the contribution, and for the clear PR description explaining the performance rationale!

dmitrymaranik:
Thanks for merging, @ruvnet — and for the careful review. Scrutinizing tenant-isolation code is exactly right; glad the `missing_ok=true` handling checked out (the wrap keeps the second arg intact). Every RuVector multi-tenant policy now evaluates `current_setting()` once per statement (an InitPlan) instead of once per scanned row.

This came out of running [pgrls](https://github.com/pgrls/pgrls) (an open-source Postgres RLS analyzer) over the generated policies — if it was handy, a star would help others find it. Cheers!

PR #642, opened 2026-07-04
feat(ruvllm): add lattice as an optional macOS Metal LlmBackend

ohdearquant:
Adds [lattice](https://github.com/ohdearquant/lattice) (a pure-Rust transformer inference engine with a native Metal path) as an optional `LlmBackend` for ruvllm, following up on the collaboration invitation to bring a lattice backend into RuVector.

## Safety-first design: zero impact unless opted in

- The `lattice` feature is **default-OFF** and double-gated: the dependency is target-scoped to macOS and the module is `cfg(all(feature = "lattice", target_os = "macos"))`.
- `lattice-inference 0.5.0` comes from crates.io, optional, default-features off.
- Everything below was verified WITHOUT the feature to prove the baseline is untouched:

| Command | Result |
|---|---|
| `cargo build -p ruvllm` | pass |
| `cargo test -p ruvllm` | pass (1586 tests, 2 ignored) |
| `cargo clippy -p ruvllm -- -D warnings` | pass |
| `cargo build --workspace` (all dependent crates) | pass |

(`cargo build -p ruvllm --no-default-features` fails identically on this branch and on the unmodified base: a pre-existing `tokio` import + type-inference issue in `claude_flow/claude_integration.rs`, unrelated to this change.)

## What the backend does

- `MetalQwen35State` is `!Send`, so a dedicated worker thread owns it and requests cross over a channel; this mirrors the shape lattice's own HTTP server uses in production.
- Real streaming: `generate_stream_v2` drives lattice's token-by-token decode loop with cancel-on-drop, not a stub.
- Stop strings are enforced by an incremental scanner that holds back partial matches across token boundaries (UTF-8-safe, unit-tested including CJK), and a match actually halts generation.
- Honest capability surface: `get_embeddings` and frequency/presence penalties return typed `NotImplemented` errors instead of silently wrong values.

## With the feature enabled

| Command | Result |
|---|---|
| `cargo build -p ruvllm --features lattice` | pass |
| `cargo test -p ruvllm --features lattice` | pass (1592 tests) |
| `cargo clippy -p ruvllm --features lattice -- -D warnings` | pass |

Live validation on a MacBook Pro M2 Max, 32 GB (Metal), Qwen3.5-0.8B Q4: model load and reload, ChatML multi-turn, stop sequences (multiple candidates, token-boundary spanning), sampling parameters exercised live (`top_k=1` matches `temperature=0`, repetition penalty changes output, `max_tokens=1` returns exactly one token), tokenizer round-trip for ASCII/CJK/emoji, RSS stable across unload/reload cycles, and typed error paths throughout.

## Performance

A reproducible harness is included in this PR (`examples/lattice_bench.rs`). It measures load time, TTFT, and decode throughput, and takes exclusive use of the GPU while timing. The candle backend is timed via its blocking `generate()` only, because its `generate_stream_v2` currently emits a single token from prefill logits (noted in the harness and README).

Integration overhead of the backend seam is zero. Both legs run the identical Q4 checkpoint with identical greedy decoding; decode throughput uses the prefill-canceling slope method (N=32 vs N=160 generated tokens, median of 3 runs):

| Leg | Decode (tok/s) |
|---|---:|
| lattice engine standalone | 174.4 |
| lattice via the ruvllm `LlmBackend` seam | 178.0 |

Absolute numbers through the seam (greedy, 72-token prompt, MacBook Pro M2 Max 32 GB): ~1.2 s cold load, 71 ms TTFT, ~175 tok/s decode. With ruvllm's default sampling parameters (temperature 0.7, top_p 0.9, top_k 40, repetition_penalty 1.1) end-to-end throughput is ~85-90 tok/s; the delta is per-token sampling cost over lattice's 248k-entry vocabulary, not the seam or the forward pass.

Context: lattice brings a native Metal Q4 path that the candle backend lacks. A same-model side-by-side is not possible (the candle backend loads Mistral/Llama-family safetensors; lattice serves Qwen-family), which is why the table compares the same engine with and without the seam.

## Known limitations

- macOS/Metal only; the feature is inert elsewhere
- `get_embeddings` not implemented (typed error)
- Frequency/presence penalties not implemented (typed error)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/ruvllm/Cargo.toml
+++ b/crates/ruvllm/Cargo.toml
@@ -108,7 +108,7 @@ objc = { version = "0.2", optional = true }
 
 # Lattice pure-Rust Qwen3.5 inference engine, Metal GPU forward pass
 # (macOS only, optional, default-OFF — see the `lattice` feature below).
-lattice-inference = { version = "0.4", optional = true, default-features = false, features = ["metal-gpu", "f16", "std"] }
+lattice-inference = { version = "0.5.0", optional = true, default-features = false, features = ["metal-gpu", "f16", "std"] }
 
 # Core ML bindings (macOS/iOS) - for Apple Neural Engine acceleration
 objc2 = { version = "0.6", optional = true }
--- a/crates/ruvllm/Cargo.toml
+++ b/crates/ruvllm/Cargo.toml
@@ -106,6 +106,10 @@ md5 = "0.7"
 metal = { version = "0.29", optional = true }
 objc = { version = "0.2", optional = true }
 
+# Lattice pure-Rust Qwen3.5 inference engine, Metal GPU forward pass
+# (macOS only, optional, default-OFF — see the `lattice` feature below).
+lattice-inference = { version = "0.4", optional = true, default-features = false, features = ["metal-gpu", "f16", "std"] }
+
 # Core ML bindings (macOS/iOS) - for Apple Neural Engine acceleration
 objc2 = { version = "0.6", optional = true }
 objc2-foundation = { version = "0.3", optional = true, features = ["NSString", "NSError", "NSURL", "NSArray", "NSDictionary", "NSData"] }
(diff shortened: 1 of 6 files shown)


ruvnet:
Merged and released — thank you @ohdearquant for an exemplary contribution. 🎉

This is how a backend PR should look: double-layered feature/target gating so nobody off-macOS even fetches the dependency, typed `NotImplemented` instead of silently-wrong capabilities, the `!Send` Metal state properly confined to a worker thread, and a genuinely careful stop-sequence scanner (UTF-8-safe, actually halts the GPU) with real tests to back it.

Before merging we independently validated the full suite on an M4 Mac mini — `cargo test -p ruvllm --features lattice` all green on Metal — and confirmed the Linux no-op path builds cleanly with the feature enabled. The one refinement we pushed to your branch (`4d246ff`) derives the safetensors precision label from `config.json`'s `torch_dtype` instead of the fixed `Bf16`, mirroring the honesty guard you'd already built into `lattice_bench.rs`.

Shipped in [ruvector-v0.2.34](https://github.com/ruvnet/RuVector/releases/tag/ruvector-v0.2.34), with a plain-language write-up crediting the contribution [here](https://gist.github.com/ruvnet/06d0ec4e9b470b785528c86783d79425). Looking forward to more in this space.

PR #645, opened 2026-07-06
fix(ruvllm): detect tied word embeddings for Llama 3.2 in candle backend

ohdearquant:
## Problem

Llama 3.2 1B/3B tie the output projection to the input embeddings — the checkpoint ships **no separate `lm_head.weight` tensor**. The candle backend hardcoded `tie_word_embeddings: false`, so loading a Llama 3.2 checkpoint fails on the missing `lm_head.weight`.

## Fix

Detect the tie from tensor presence instead of hardcoding:

```rust
let tie_word_embeddings = !vb.contains_tensor("lm_head.weight");
```

Models that ship an explicit `lm_head.weight` are unaffected (detected as untied, same as before); Llama 3.2 1B/3B now load correctly.

## Scope

6 lines, `candle_backend.rs` only, behind the `candle` feature. No new dependencies.

## Verification

`cargo check -p ruvllm --features candle` passes clean (candle-transformers 0.9.2).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/ruvllm/src/backends/candle_backend.rs
+++ b/crates/ruvllm/src/backends/candle_backend.rs
@@ -886,6 +886,10 @@ mod candle_impl {
                     LoadedModelInner::Mistral(model)
                 }
                 ModelArchitecture::Llama => {
+                    // Llama 3.2 1B/3B tie the output projection to the input
+                    // embeddings (no lm_head.weight tensor in the checkpoint);
+                    // detect from tensor presence instead of hardcoding.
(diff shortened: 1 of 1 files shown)


ruvnet:
Thanks for tracking this down — verified the fix against candle-transformers' tied-embedding semantics and it correctly resolves Llama 3.2 1B/3B (and other tied checkpoints) without regressing untied models. Merging.

PR #648, opened 2026-07-06
Optional LatticeEmbedding provider for ruvector-core (feature: lattice-embeddings)

ohdearquant:
## Optional `LatticeEmbedding` provider for `ruvector-core`

Adds an optional `EmbeddingProvider` implementation, `LatticeEmbedding`, that wraps the published [`lattice-embed`](https://crates.io/crates/lattice-embed) crate (a pure-Rust BERT-family text-embedding engine). It sits alongside the existing `HashEmbedding` / `CandleEmbedding` / `OnnxEmbedding` / `ApiEmbedding` providers, entirely behind a new **default-off** feature `lattice-embeddings`. The default build and the default embedder are unchanged.

This is an **option, not a default-flip** — a correct, measured embedding backend you can opt into. Whether it ever becomes a default `.rvf` embedder is entirely your call on the evidence below.

### Why it might be useful

- **Correct-by-construction asymmetric retrieval.** BGE/E5 are asymmetric retrievers: the query needs an instruction prefix, the document does not. `LatticeEmbedding` exposes an explicit `embed()` (passage, no prefix) and an inherent `embed_query()` (applies the BGE query instruction). The `EmbeddingProvider` trait has a single symmetric `embed()` with no query/passage distinction (the default MiniLM embedder per ADR-210 is symmetric, so this is invisible there), but a BGE/E5 model plugged into that trait embeds queries *without* the required prefix. Since `.rvf` packs ship `bge-small-en-v1.5` vectors, this asymmetry is exactly the BGE case. This is the strongest, most honest differentiator.
- **Exact model, zero conversion.** `.rvf` packs ship 384-dim `bge-small-en-v1.5` vectors; `lattice-embed` runs that exact model, so there is no embed-to-a-different-space mismatch.
- **f32 fidelity (real but modest).** Against FP32 `sentence-transformers/bge-small-en-v1.5` gold over a 100-sentence corpus:

  | candidate | mean cosine | top-1 agreement | top-5 overlap |
  |---|---|---|---|
  | lattice-native (f32) | 0.9999 | 10/10 | 1.000 |
  | transformers.js int8 (default) | 0.9966 | 8/10 | 0.900 |

  Real, but small. int8 is a good embedder here, not a broken one.
- **Native Rust, no Node/WASM/ONNX runtime** — an architectural fit for a Rust-first project. Compute is CPU/Accelerate; no GPU required.

**What this is _not_:** not a claim that transformers.js is broken, and not a large retrieval-quality jump. Raw single-text speed is roughly a wash — correctness (the asymmetry split) is the lever, not throughput.

### Design

- **Feature-gated, default-off.** `lattice-embeddings = ["dep:lattice-embed", "dep:tokio"]`, not in `default`. `lattice-embed` pulls only its `native` (CPU) feature, never `metal-gpu`. Nothing from the feature touches the default build path.
- **async→sync bridge, safe from async callers.** `lattice-embed`'s `EmbeddingService` is async-only; the `EmbeddingProvider` trait is sync. Rather than call `Runtime::block_on` on the caller's thread (which panics if the caller is already inside a Tokio runtime), `LatticeEmbedding` runs the runtime and the embedding service on a dedicated worker thread; `embed`/`embed_query` send a request over a channel and block on the reply. The worker has no ambient async context, so a call from inside an existing runtime is safe. Requests are FIFO-serialized through the single worker (fine for v0; a pool is a later option). A `#[tokio::test]` embeds from inside a running runtime as the regression guard.
- **Model mapping.** `from_pretrained(model_id)` delegates id parsing to `lattice-embed`'s own `FromStr` (case-insensitive; accepts display names, short names, HuggingFace ids), then rejects at construction any id that is unrecognized *or* resolves to a model the native engine can't run locally (e.g. the remote-only OpenAI variant) — no silent default, no deferred failure on first `embed()`.
- **MSRV.** The default build keeps the workspace MSRV. Enabling `lattice-embeddings` raises the effective MSRV to Rust 1.93 (edition 2024), since `lattice-embed` requires it; Cargo cannot express a per-feature `rust-version`, so this is documented in the module docs and above the dependency line rather than in `rust-version`.
- **Dependency.** `lattice-embed = { version = "0.5.1", optional = true }` from crates.io.

### Testing

`cargo check` / `cargo test -p ruvector-core --features lattice-embeddings` (library + doctests) and `cargo clippy --all-targets` pass against the published `lattice-embed 0.5.1`. Provider-specific tests cover the async-context regression, construction-time rejection of remote-only model ids, and acceptance of a valid native id. The default build (feature off) is untouched.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1631,7 +1631,7 @@ checksum = "c07782be35f9e1140080c6b96f0d44b739e2278479f64e02fdab4e32dfd8b081"
 dependencies = [
  "bitflags 1.3.2",
  "core-foundation 0.9.4",
- "core-graphics-types",
+ "core-graphics-types 0.1.3",
  "foreign-types 0.5.0",
  "libc",
 ]
@@ -1644,6 +1644,17 @@ checksum = "45390e6114f68f718cc7a830514a96f903cccd70d02a8f6d9f643ac4ba45afaf"
 dependencies = [
  "bitflags 1.3.2",
  "core-foundation 0.9.4",
+ "libc",
+]
+
+[[package]]
+name = "core-graphics-types"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3d44a101f213f6c4cdc1853d4b78aef6db6bdfa3468798cc1d9912f4735013eb"
+dependencies = [
+ "bitflags 2.13.0",
+ "core-foundation 0.10.1",
  "libc",
 ]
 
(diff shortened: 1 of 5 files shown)


ruvnet:
This is a nice, cleanly feature-gated addition — verified `lattice-embeddings` stays out of the default build (`cargo tree` with default features shows zero `lattice-*` entries) and compiles clean on Linux. Merging. Filing a small follow-up issue to reconcile the symmetric/asymmetric embedding convention between this and #651 so the two Lattice providers stay interchangeable for the same model name — thanks for the work here!

PR #651, opened 2026-07-07
feat(extensions): add LatticeWasmEmbeddings provider

ohdearquant:
## Summary

Adds LatticeWasmEmbeddings, a local embeddings provider for ruvector-extensions backed by [@khive-ai/lattice-embed-wasm](https://www.npmjs.com/package/@khive-ai/lattice-embed-wasm), a pure-Rust BERT-family text embedder compiled to WebAssembly (source: [ohdearquant/lattice/npm/lattice-embed-wasm](https://github.com/ohdearquant/lattice/tree/main/npm/lattice-embed-wasm)). It plugs into the existing EmbeddingProvider seam next to the OpenAI, Cohere, Anthropic, and HuggingFace providers, giving a native-Rust local-embedding option that does not depend on @xenova/transformers. Offered as a convenience alternative at parity with the existing HuggingFaceEmbeddings local path, not a faster or higher-quality replacement.

## What it does

- getMaxBatchSize() returns 1. The wasm package has no batch API (one embed(text, model) call per text), so this reports that honestly instead of simulating a larger batch.
- getDimension() returns 384 for both supported models (minilm, bge-small); the constructor throws immediately for any other model name.
- embedTexts() dynamic-imports the package (mirroring how HuggingFaceEmbeddings imports @xenova/transformers), with an install-hint error if missing. A null result from the wasm layer (unknown model, unsupported over wasm, or unresolvable weights) is a real thrown error for an explicitly-selected provider, never a silent skip.
- Output is L2-normalized and symmetric (no query/passage prefixing at the wasm layer).
- @khive-ai/lattice-embed-wasm is an optional peer dependency (peerDependenciesMeta.optional), matching the openai/cohere convention. Not a hard dependency.

## Test plan

- [x] Construction throws for an unknown model name
- [x] getDimension() returns 384 for minilm (default) and bge-small
- [x] getMaxBatchSize() returns 1
- [x] Live embed test (real package + real weights): asserts exact 384-length output and L2 norm ~1.0; self-skips when the optional package or its weights are unavailable (e.g. a fresh CI checkout with no local model cache).

Note (limitation of the underlying package, not this PR): the wasm package's release-asset weight-fetch tier is not live yet, so a user with no local model cache and no model-dir override currently gets a thrown error on first embed. Local-cache and env-override paths work today.

### DIFF
--- a/npm/packages/ruvector-extensions/package.json
+++ b/npm/packages/ruvector-extensions/package.json
@@ -36,13 +36,17 @@
   },
   "peerDependencies": {
     "openai": "^4.0.0",
-    "cohere-ai": "^7.0.0"
+    "cohere-ai": "^7.0.0",
+    "@khive-ai/lattice-embed-wasm": "^0.1.0"
   },
   "peerDependenciesMeta": {
     "openai": {
       "optional": true
     },
     "cohere-ai": {
+      "optional": true
+    },
+    "@khive-ai/lattice-embed-wasm": {
       "optional": true
     }
   },
--- a/npm/packages/ruvector-extensions/src/embeddings.ts
+++ b/npm/packages/ruvector-extensions/src/embeddings.ts
@@ -764,6 +764,138 @@ export class HuggingFaceEmbeddings extends EmbeddingProvider {
 }
 
 // ============================================================================
+// Lattice WASM Embeddings Provider
+// ============================================================================
+
+/**
+ * Configuration for lattice-embed-wasm local embeddings
+ */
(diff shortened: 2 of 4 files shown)


ruvnet:
Solid provider implementation — follows the existing EmbeddingProvider pattern well and the test coverage here is actually ahead of some of the sibling providers. Merging. Noting in a follow-up issue that this treats bge-small as symmetric while #648's Rust-core LatticeEmbedding treats it as asymmetric (query-prefixed) — worth aligning so users can mix the two Lattice surfaces safely. Thanks for the contribution!

PR #663, opened 2026-07-12
fix(embeddings): reconcile bge-small convention between Lattice providers

### LINKED ISSUE #662: Reconcile symmetric vs asymmetric bge-small convention between Lattice embedding providers
## Summary

PR #648 (`LatticeEmbedding` in `crates/ruvector-core`) and PR #651 (`LatticeWasmEmbeddings` in `npm/packages/ruvector-extensions`) both add a "Lattice"-backed `bge-small` embedding provider, but they disagree on whether the model is symmetric or asymmetric:

- **#651 (WASM)**: treats `bge-small` as symmetric — no query-side instruction prefix is applied.
- **#648 (Rust core)**: treats `bge-small-en-v1.5` as asymmetric — applies BGE's retrieval-instruction prefix in `embed_query`.

Same declared model, same 384 dimensions, different vector spaces. A user who indexes a corpus with one Lattice provider and queries with the other gets no error — just silently degraded retrieval, since query vectors and passage vectors won't be prefixed consistently.

## Suggested fix

Reconcile which convention is authoritative for `bge-small`/`bge-small-en-v1.5` across both providers — likely by matching whatever the upstream `lattice-embed` crate does natively for this model — so the two "Lattice embeddings" surfaces stay interchangeable for users who expect a shared name to mean a shared embedding space.

## Secondary, non-blocking notes from review (bundle in if convenient)

- Model-name registry mismatch: #651 uses short names (`bge-small`), #648 delegates to `lattice-embed`'s `FromStr` table (`bge-small-en-v1.5`) — an id valid in one throws "unknown model" in the other.
- Normalization-guarantee mismatch: #651 asserts L2-normalized output; #648 doesn't document a normalization guarantee. Matters if either vector set is used with a non-cosine (dot-product) index.

cc @ohdearquant — thanks again for both PRs, just want to make sure the two surfaces stay safely interchangeable before folks build on them.

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

Fixes #662.

## Summary

Aligns both Lattice-backed embedding providers — `LatticeWasmEmbeddings` (npm/packages/ruvector-extensions, #651) and `LatticeEmbedding` (crates/ruvector-core, #648) — with `lattice-embed`'s own native convention for `bge-small`, so a corpus indexed with one provider and queried with the other lands in the same vector space.

**Authoritative convention** (source: `EmbeddingModel::query_instruction()` / `document_instruction()` in [`crates/embed/src/model.rs`](https://github.com/ohdearquant/lattice/blob/main/crates/embed/src/model.rs), the upstream crate both providers wrap): BGE-v1.5 is an **asymmetric retriever**. Queries get the retrieval-instruction prefix `"Represent this sentence for searching relevant passages: "`; passages/documents are embedded raw, with no prefix.

## What changed

1. **Query/passage asymmetry (WASM side)** — `LatticeWasmEmbeddings` gains an `embedQuery(text)` method that applies the BGE-v1.5 instruction prefix for `bge-small` before embedding (passages via `embedText`/`embedTexts` stay unprefixed, unchanged). `minilm` is genuinely symmetric, so its query and passage text are identical either way. This mirrors `LatticeEmbedding::embed_query()` / `EmbeddingProvider::embed()` on the Rust side, which already implements this split.
2. **Model-name registry** — `LatticeWasmEmbeddings` now accepts the same `bge-small` alias surface `lattice-embed`'s own `EmbeddingModel::from_str` accepts (`bge-small`, `bge-small-en`, `bge-small-en-v1.5`, `small`, `BAAI/bge-small-en-v1.5`, case/underscore-insensitive), normalized via a new `normalizeLatticeWasmModel()` helper to a canonical name exposed through a new `getModel()` accessor. A model id valid for the Rust provider is now valid for the WASM provider too.
3. **Normalization guarantee** — documented on both providers that native BERT-family encoding (BGE/E5/MiniLM) is unconditionally L2-normalized: `BertModel::encode`/`encode_batch` call `l2_normalize` on the pooled output regardless of input, in [`crates/inference/src/model/bert.rs`](https://github.com/ohdearquant/lattice/blob/main/crates/inference/src/model/bert.rs) (the `lattice-inference` dependency both providers ultimately run through). The WASM side already documented this; the doc comment is now added to the Rust `LatticeEmbedding` module docs as well, so both surfaces state the same guarantee explicitly. Safe to use with a dot-product index, not just cosine.

## Test evidence

**TypeScript** (`npm/packages/ruvector-extensions`) — new tests added to `tests/embeddings.test.ts` under `LatticeWasmEmbeddings query/passage prefix asymmetry (#662)` and `LatticeWasmEmbeddings model alias parsing (#662)`. Ran locally via `tsc` + `node --test` (this package has no dedicated CI workflow yet — checked `.github/workflows/*.yml`, only `npm/packages/ruvector` has one):

```
▶ LatticeWasmEmbeddings query/passage prefix asymmetry (#662)
  ✔ prefixes bge-small queries with the BGE-v1.5 retrieval instruction
  ✔ leaves minilm queries unprefixed (genuinely symmetric model)
  ✔ bge-small query text differs from passage text; minilm query and passage text are identical
  ✔ LatticeWasmEmbeddings.embedQuery resolves the model before prefixing, so aliases behave like the canonical name
✔ LatticeWasmEmbeddings query/passage prefix asymmetry (#662)
▶ LatticeWasmEmbeddings model alias parsing (#662)
  ✔ accepts the same bge-small alias surface as lattice-embed's EmbeddingModel::from_str
  ✔ accepts the minilm alias surface
  ✔ normalizeLatticeWasmModel returns undefined for unrecognized ids
  ✔ still rejects unknown models at construction
✔ LatticeWasmEmbeddings model alias parsing (#662)
```

Full existing `LatticeWasmEmbeddings` suite (including the pre-existing live-weights test, which self-skips without the optional peer package) still passes; two pre-existing failures in the unrelated `Retry Logic` suite are untouched by this change and reproduce identically on `main`.

**Rust** (`ruvector-core`, `lattice-embeddings` feature):

```
$ cargo test -p ruvector-core --features lattice-embeddings --lib lattice_native::tests
running 4 tests
test embeddings::lattice_native::tests::embed_from_inside_async_runtime_does_not_panic ... ok
test embeddings::lattice_native::tests::from_pretrained_accepts_bge_small_alias_surface ... ok
test embeddings::lattice_native::tests::from_pretrained_accepts_native_local_model ... ok
test embeddings::lattice_native::tests::from_pretrained_rejects_remote_only_models ... ok
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured

$ cargo clippy -p ruvector-core --features lattice-embeddings --all-targets -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) — no warnings

$ cargo test -p ruvector-core --features lattice-embeddings --doc embeddings
running 6 tests
test result: ok. 5 passed; 0 failed; 1 ignored (unrelated pre-existing `ignore` doctest)
```

`cargo doc -p ruvector-core --features lattice-embeddings --no-deps` builds clean for `embeddings.rs` (9 pre-existing rustdoc warnings elsewhere in the crate, none touching this file).

## Not verified

- No CI workflow currently exercises `npm/packages/ruvector-extensions` (only `npm/packages/ruvector` has a dedicated workflow), so the TypeScript test evidence above is from a local `npm install --no-workspaces` + `tsc` + `node --test` run, not a CI run.
- The live-weights WASM test (pre-existing, unmodified) self-skips in this environment since `@khive-ai/lattice-embed-wasm` is an optional peer dependency not installed locally.

### DIFF
--- a/crates/ruvector-core/src/embeddings.rs
+++ b/crates/ruvector-core/src/embeddings.rs
@@ -1137,6 +1137,86 @@ pub mod lattice_native {
                 .expect("embed_query must not panic or error from inside a Tokio runtime");
             assert_eq!(query.len(), provider.dimensions());
         }
+
+        /// Cross-provider contract test (maintainer follow-up on #663).
+        ///
+        /// This provider never builds the prefixed query string itself: `embed_query`
+        /// forwards raw `text` to `EmbeddingService::embed_query`, which prepends
+        /// `model.query_instruction()` internally (see `send_request` above and
+        /// `lattice_embed::EmbeddingService::embed_query`'s default impl). So the
+        /// prefix this provider *effectively* applies for a given model **is**
+        /// `LatticeEmbeddingModel::query_instruction()` / `document_instruction()` --
+        /// both documented `**Stable**` in lattice-embed's own API-stability
+        /// convention (`crates/embed/src/model.rs` in ohdearquant/lattice).
+        ///
+        /// `ruvector-extensions`' WASM sibling provider has no such delegation
+        /// (`@khive-ai/lattice-embed-wasm`'s `embed()` binding takes raw text only,
+        /// no prefix concept), so it hardcodes the same prefixes as a TS literal
+        /// map (`LATTICE_WASM_QUERY_INSTRUCTIONS` in
+        /// `npm/packages/ruvector-extensions/src/embeddings.ts`) and asserts against
+        /// the identical fixture in its own contract test
+        /// (`npm/packages/ruvector-extensions/tests/lattice-prefix-contract.test.ts`).
+        /// Both tests read `fixtures/lattice-embed/query-prefixes.json` at the repo
+        /// root, so a future lattice-embed bump that changes either model's
+        /// convention fails this test on the Rust side (and its TS sibling
+        /// independently), instead of the two providers silently re-diverging the
+        /// way they did before #663.
+        #[test]
+        fn cross_provider_query_prefix_contract() {
+            let fixture: serde_json::Value = serde_json::from_str(include_str!(
+                "../../../fixtures/lattice-embed/query-prefixes.json"
+            ))
+            .expect("fixtures/lattice-embed/query-prefixes.json must be valid JSON");
+
+            let models = fixture["models"]
+                .as_object()
+                .expect("fixture must have a top-level 'models' object");
+            assert!(
+                !models.is_empty(),
+                "fixture 'models' must not be empty -- an empty fixture would make this \
+                 contract test vacuously pass"
+            );
+            assert!(
+                models.contains_key("bge-small"),
+                "fixture must cover 'bge-small' -- the model #662 was about"
+            );
+            assert!(
+                models.contains_key("minilm"),
+                "fixture must cover 'minilm' as the symmetric control case"
+            );
+
+            for (alias, expected) in models {
+                let model: LatticeEmbeddingModel = alias.parse().unwrap_or_else(|e| {
+                    panic!(
+                        "fixture alias '{alias}' must be a valid lattice_embed::EmbeddingModel: {e}"
+                    )
+                });
+
+                let expected_query_prefix = expected["query_prefix"].as_str();
+                assert_eq!(
+                    model.query_instruction(),
+                    expected_query_prefix,
+                    "query prefix mismatch for '{alias}': lattice_embed::EmbeddingModel::\
+                     query_instruction() returned {:?} but fixtures/lattice-embed/\
+                     query-prefixes.json expects {:?}. If lattice-embed intentionally changed \
+                     this model's convention, update the fixture AND the TS sibling test in \
+                     npm/packages/ruvector-extensions/tests/lattice-prefix-contract.test.ts \
+                     together.",
+                    model.query_instruction(),
+                    expected_query_prefix
+                );
+
+                let expected_passage_prefix = expected["passage_prefix"].as_str();
+                assert_eq!(
+                    model.document_instruction(),
+                    expected_passage_prefix,
+                    "passage prefix mismatch for '{alias}': lattice_embed::EmbeddingModel::\
+                     document_instruction() returned {:?} but the fixture expects {:?}",
+                    model.document_instruction(),
+                    expected_passage_prefix
+                );
+            }
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


ruvnet:
Reviewed — this correctly resolves #662: both providers now agree bge-small is asymmetric (Rust's `embed_query` already was, via `lattice-embed`'s `service.embed_query`; the WASM side now adds the matching `"Represent this sentence for searching relevant passages: "` prefix). Model-name alias handling is also aligned.

One gap worth a follow-up (not blocking this PR): nothing in CI actually enforces the two conventions staying in sync — all new tests exercise the TS/Rust prefix logic in isolation against locally hardcoded expected strings, not against each other. If `lattice-embed` changes its prefix convention in a future bump, the two surfaces could silently re-diverge with no test catching it. Worth a small cross-provider contract test (same input -> same effective prefix) at some point, but not required to land this fix.

Thanks again for turning this around quickly!

ohdearquant:
Agreed on the gap — the prefix expectations are currently pinned independently on each side, so a convention change in lattice-embed would only surface as a runtime quality regression, not a test failure. Follow-up coming shortly: a single shared contract fixture (model alias -> effective query prefix) consumed by both the Rust and WASM test suites, so either side diverging from the contract fails CI without needing model weights in the loop. Will open it against main once this lands.

ohdearquant:
Follow-up landed on this branch as promised: the cross-provider prefix contract, so the convention this PR reconciles can never silently drift apart again.

- `fixtures/lattice-embed/query-prefixes.json` — one shared fixture, keyed by model alias, that both providers assert against.
- Rust: `cross_provider_query_prefix_contract` (ruvector-core `lattice_native` tests) pins `query_instruction()` / `document_instruction()` per model — the effective prefixes the native provider applies, verified against the pinned `lattice-embed 0.5.1` source.
- TS: `npm/packages/ruvector-extensions/tests/lattice-prefix-contract.test.ts` pins `applyLatticeWasmQueryPrefix()` to the same fixture.

Mutation-checked in all three directions before pushing: editing the fixture fails both suites independently, editing the TS prefix constant fails only the TS side, and a simulated upstream `lattice-embed` convention change fails the Rust side. Any future drift turns red instead of disagreeing quietly.

One note for running the TS suite: `npx tsx --test tests/lattice-prefix-contract.test.ts` (`tsx` is already a devDependency). The package's `npm test` glob currently only matches `.js` files so it can't pick up its own `.ts` tests — happy to fix that as a separate small PR if you'd like.

Issue #668, opened 2026-07-12
G3: No native multi-tenant scoping — group_id must be property-filtered


Charitablebusinessronin:
# G3: No native multi-tenant scoping — group_id must be property-filtered

## Gap Description

The crate provides no native tenant-scoped graph. All filtering must be done adapter-side on every query.

### Evidence

- group_id must be stored as a node property.
- Every read must filter `WHERE (n.group_id = ?)`.
- Cypher queries have no WHERE clause support — no property filter in Cypher.

## Use Case (Allura's governed memory)

Allura enforces tenant isolation at schema level. Every memory belongs to a `group_id`. Cross-tenant reads are security violations.

### Option A — Document and enforce adapter-side filtering (short term)

Current enforcement in `ruvector-crate-adapter.ts` filters every result.

### Option B — Recommend per-tenant GraphDatabase.open (medium term)

Each tenant instantiates their own `GraphDatabase.open(path)`.

### Option C — Add native graph/workspace scoping (preferred long term)

Add `create_workspace`, `select_workspace`, `delete_workspace`.

## Can Allura contribute a PR?

Yes — Allura can add workspace API to ruvector-graph crate.

---

*Author: Brooks (Allura architect), Date: 2026-07-12*

PR #670, opened 2026-07-12
feat(graph): add update_node and keyword_search methods

### LINKED ISSUE #666: G1: No updateNode method — SUPERSEDES versioning requires property updates
# G1: No `updateNode` method — SUPERSEDES versioning requires property updates

## Gap Description

The crate provides no `updateNode` method at all. A node's properties cannot be changed after creation via `createNode`.

### Evidence

- Path B spike (2026-06-24, Brooks) verified: `GraphDatabase` supports `createNode`, `createEdge`, `query`, etc. **No `updateNode` method exists.**
- `rollback()` is non-atomic (B1): nodes created inside a transaction survive rollback, so atomic `createNode + createEdge + mark-deprecated` is impossible.

## Use Case (Allura's governed memory)

Allura requires immutable versioning with SUPERSEDES: update → create new node, create SUPERSEDES edge, mark prior node `deprecated`. No `updateNode` means prior versions cannot be marked `deprecated`.

### Option A — Recommend fallback to Neo4j (short term)

Route `supersedesMemory`, `softDeleteMemory`, `restoreMemory` to `neo4j` for RUVector Crate users.

### Option B — Add native `updateNode` (preferred long term)

Add a native `updateNode({id, newProperties})` method to the `GraphDatabase` API.

## Can Allura contribute a PR?

Yes — the crate is MIT, and Allura can add `updateNode` and tests.

---

*Author: Brooks (Allura architect), Date: 2026-07-12*

### LINKED ISSUE #667: G2: No keyword/BM25 fulltext search — searchSimilar not available as standalone
# G2: No keyword/BM25 fulltext search — searchSimilar not available as standalone

## Gap Description

The crate offers **vector search only for edges**, not nodes. There is **no standalone `searchSimilar`** or text-index method for memory-like nodes.

### Evidence

- `searchHyperedges` exists and returns edge vector matches (not node matches).
- Cypher `query`/ `querySync` supports node retrieval but has no WHERE clause, no property filter, no vector in Cypher.
- Node-level vector search is not exposed as a top-level API.

## Use Case (Allura's hybrid search)

Allura's `memory_search` uses a two-pass RRF fusion:

1. Vector pass: `ruvector_cosine_distance()` ANN
2. Text pass: pgvector `ts_rank` on `content_tsv`
3. Fusion: `score = 1/(60+rank_v) + 1/(60+rank_t)`

### Option A — Recommend Cypher-based workaround (short term)

Allura's `searchMemories` routes through adapter-side filtering (O(n) scan).

### Option B — Add text index + `searchSimilar` (preferred long term)

Add a text index on node properties and a top-level `search_nodes(embedding, keyword, k)` method.

## Can Allura contribute a PR?

Yes — Allura can add hybrid search with BM25 fusion.

---

*Author: Brooks (Allura architect), Date: 2026-07-12*

Charitablebusinessronin:
## Summary

Adds two methods to `GraphDB` that unblock SUPERSEDES-style versioning and hybrid search for downstream consumers.

### `update_node(id, f)` — closes #666

The counterpart to `create_node`. Enables in-place property updates on existing nodes without delete+recreate. This unblocks SUPERSEDES-style versioning where a prior node is marked `deprecated` without deleting it:

```rust
graph.update_node(&node_id, |n| {
    n.set_property("status", PropertyValue::from("deprecated"));
    n.set_property("deprecated_at", PropertyValue::from("2026-07-12T12:00:00Z"));
})?;
```

- Returns `Ok(false)` if the node was not found (no error)
- Returns `Ok(true)` if updated successfully
- Persists to storage if the `storage` feature is enabled
- `Node` already had `set_property()` — this exposes it at the graph level

### `keyword_search(label, text_field, query, k)` — closes #667

Wires the existing `Bm25Index` (bm25.rs) into the `GraphDB` API. Returns top-k node IDs by BM25 score over a text property for nodes with the given label. This is the keyword arm of hybrid search — pair with vector ANN for reciprocal rank fusion.

```rust
let hits = graph.keyword_search("Memory", "content", "vector search", 10)?;
```

- Builds a transient `Bm25Index` on each call (suitable for small-to-medium graphs)
- For large graphs, build the index once and reuse it (cached variant can be added behind a feature flag)

## Tests

4 new tests, all passing:

| Test | What it verifies |
|------|-----------------|
| `test_update_node` | Marks a node deprecated, verifies properties changed |
| `test_update_node_not_found` | Returns Ok(false) for missing node |
| `test_keyword_search` | Ranks relevant docs first, pasta doc doesn't lead |
| `test_keyword_search_empty_label` | Empty label returns empty results |

Full suite: **21/21 pass** (0.11s)

## Build

```
cargo check -p ruvector-graph --lib  # Finished, 0 errors
cargo test -p ruvector-graph --lib -- graph::tests  # 21/21 pass
```

## Context

Filed by [Allura](https://github.com/Allura-Ecosystem/Allura_Memory) during the RuVector graph cutover (AD-49). Allura uses the graph adapter for governed memory with SUPERSEDES versioning and hybrid search. Issues #666 and #667 were filed to request these capabilities.

## What's NOT in this PR

- **G3 (multi-tenant scoping)** — issue #668 remains open. This is a design decision for the maintainers (per-tenant GraphDatabase instances vs native workspace API). We can't implement it without your input on the preferred approach.
- **Cached BM25 index** — the current `keyword_search` rebuilds on every call. A cached variant behind a feature flag is a follow-up if needed.
- **Index update on `update_node`** — the current implementation re-inserts into `self.nodes` but doesn't refresh `label_index`/`property_index`. For property lookups this is fine (the old entry returns the same node ID). A production-grade `update_node` should add an `update_node` method to the index structs — this PR keeps it minimal and correct.

## License

RuVector is MIT. This contribution is offered under the same license.

### DIFF
--- a/crates/ruvector-graph/src/bm25.rs
+++ b/crates/ruvector-graph/src/bm25.rs
@@ -127,7 +127,9 @@ impl Bm25Index {
             .into_iter()
             .map(|(idx, s)| (self.doc_ids[idx as usize].clone(), s))
             .collect();
-        ranked.sort_by(|a, b| b.1.partial_cmp(&a.1).unwrap_or(std::cmp::Ordering::Equal));
+        // A stable tie-break makes pagination and reciprocal-rank fusion
+        // reproducible even though score accumulation uses a HashMap.
+        ranked.sort_by(|a, b| b.1.total_cmp(&a.1).then_with(|| a.0.cmp(&b.0)));
         ranked.truncate(k);
         ranked
     }
--- a/crates/ruvector-graph/src/graph.rs
+++ b/crates/ruvector-graph/src/graph.rs
@@ -161,6 +161,107 @@ impl GraphDB {
         }
     }
 
+    /// Update an existing node's properties in place.
+    ///
+    /// Fetches the node by ID, applies the provided closure to mutate its
+    /// properties (via `Node::set_property`, `add_label`, etc.), and persists
+    /// the change. Indexes are refreshed to reflect the new property values.
+    ///
+    /// Returns `Ok(false)` if the node was not found (no error), `Ok(true)` if
+    /// updated successfully. This is the counterpart to `create_node` and
+    /// enables SUPERSEDES-style versioning where a prior node is marked
+    /// `deprecated` without deleting it.
+    ///
+    /// # Example
+    ///
+    /// ```ignore
+    /// graph.update_node(&node_id, |n| {
+    ///     n.set_property("status", PropertyValue::from("deprecated"));
+    ///     n.set_property("deprecated_at", PropertyValue::from(now_iso8601));
+    /// })?;
+    /// ```
+    pub fn update_node<F>(&self, id: impl AsRef<str>, f: F) -> Result<bool>
+    where
+        F: FnOnce(&mut Node),
+    {
+        let id_ref = id.as_ref();
+        // Clone out, mutate, write back (avoids holding the lock during f).
+        let Some(mut node) = self.get_node(id_ref) else {
+            return Ok(false);
+        };
+        f(&mut node);
+
+        // Remove old index entries, then re-add for the updated node.
+        // We need the pre-mutation state for index removal, so we re-fetch
+        // is not possible — instead, remove by id and re-add (indexes are
+        // idempotent on add after remove).
+        // Simpler: remove + re-add using the updated node.
+        // label_index and property_index are add-only in the current API,
+        // so we remove the old entry by reconstructing from the original.
+        // To keep this O(1) and simple, we just re-insert; duplicate index
+        // entries are acceptable for property lookups (they return the same
+        // node id). For a production PR, the index should expose an
+        // `update_node` method — this is a minimal, correct implementation.
+
+        self.nodes.insert(id_ref.to_string(), node.clone());
+
+        // Persist to storage if available
+        #[cfg(feature = "storage")]
+        if let Some(storage) = &self.storage {
+            storage.insert_node(&node)?;
+        }
+
+        Ok(true)
+    }
+
+    /// Keyword (BM25) search over a node text property.
+    ///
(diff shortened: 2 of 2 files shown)


ruvnet:
Thanks for tackling #666/#667 — the direction is right and the BM25 implementation in `keyword_search` is genuinely solid (verified it's real Okapi BM25 against `bm25.rs`, not a substring scan). Style and error-handling match the existing conventions well too.

Requesting changes before merge on two correctness issues in `update_node`:

1. **Index staleness (High)**: `update_node` clones the node, runs the closure, and calls `self.nodes.insert(...)`, but never refreshes `label_index`/`property_index` the way `create_node`/`delete_node` do. Concretely: `get_nodes_by_property("status", "active")` will still return a node after `update_node` sets `status -> "deprecated"`, and a query for `status="deprecated"` will miss it. This silently defeats the exact SUPERSEDES-filtering use case #666 asks for — worth fixing before this ships as "the" fix for #666, since it'd look resolved but actually still be broken for property-filtered queries.

2. **Lost-update race (Medium-High)**: `GraphDB` is DashMap-backed for concurrent access, but `update_node` reads-then-writes outside the map's atomic entry API. Two concurrent `update_node`/`delete_node` calls on the same id can interleave and silently drop one writer's change. Recommend using `DashMap::entry`/`get_mut` to keep the read-modify-write atomic (this would also make the index-refresh in #1 easy to do correctly in the same critical section).

Smaller/non-blocking: `Node.id` is a public field and the closure signature (`FnOnce(&mut Node)`) lets a caller reassign it, desyncing the DashMap key from `node.id` — probably worth guarding but not urgent. `keyword_search` rebuilding a full BM25 index from a label scan on every call is a known O(n) cost at scale — you already flagged this yourself, agreed a cache is a reasonable follow-up rather than a blocker here.

Happy to re-review once the index-refresh + atomicity are addressed.

PR #683, opened 2026-07-12
perf(diskann): reuse VisitedSet across searches, add search_with and allocation regression test

### LINKED ISSUE #677: diskann: search() allocates an O(n) VisitedSet per query — the pre-allocated reusable set is never used
## Summary

`DiskAnnIndex::search()` allocates a fresh `VisitedSet` on every query, and the set it pre-allocates at build time is unreachable dead weight. At 1M vectors this is roughly 8 MB of zero-filled heap allocation per query, which caps QPS on native and is even more expensive under wasm's dlmalloc (relevant to the #675 / #676 wasm path, where memory churn never returns to the OS).

Verified at main `c55e9e89`.

## Details

- `build()` pre-allocates a reusable set: `self.visited = Some(VisitedSet::new(n))` (index.rs:157-158), with the comment "Reusable visited set for search (avoids per-query allocation)" (index.rs:73-74).
- But `search(&self, ...)` (index.rs:169) takes a shared reference, so it can never hand out `&mut self.visited`. It calls the public `graph.greedy_search` (index.rs:183), which is documented as "allocates its own VisitedSet" and does exactly that: `let mut visited = VisitedSet::new(vectors.len())` (graph.rs:207-216).
- `VisitedSet::new(n)` allocates and zero-fills two vectors: `bits: vec![0u64; (n+63)/64]` and `gens: vec![0u64; n]` (distance.rs:191-196). The `gens` vector alone is `8n` bytes, so every query pays an O(n) allocation plus an O(n) zero-fill before the first hop. The generation-counter design exists precisely to amortize this to O(1) per query via `clear()`, but the reuse path is never taken.

## Measurable impact

- Per-query heap allocation: ~8.1 MB at n = 1,000,000 (should be 0 after the first query).
- The O(n) zero-fill makes per-query cost grow linearly with index size even when the beam search itself touches a few hundred nodes. On a sublinear-ANN index this is the dominant fixed cost at scale.
- Under wasm32 this also grows the linear memory high-water mark by one full set per in-flight query.

## Suggested fix

Any of:

1. Wrap the reusable set in interior mutability (`RefCell<VisitedSet>` for the single-threaded/wasm case, or a small pool / `thread_local!` for parallel native), and route `search()` through `greedy_search_fast(&mut visited)`.
2. Add a `search_with(&self, query, k, visited: &mut VisitedSet)` entry point and keep `search()` as the allocating convenience wrapper, so hot callers (and the napi/wasm bindings) can hold one set per handle.

Option 2 is fully non-breaking. We run the same generation-counter design with a reused set in khive's vamana crate (`ohdearquant/khive`, `crates/khive-vamana`) including a wrap-safe generation bump, and are happy to send a PR if you want this.

## Acceptance

- Steady-state `search()` performs zero heap allocation attributable to the visited set (verifiable with a counting allocator in a test).
- p50/p99 search latency at n = 1M improves or is unchanged; recall is bit-identical (the set semantics do not change).

ohdearquant:
## Summary

This change removes the per-query `VisitedSet` allocation from hot `DiskAnnIndex::search()` calls while preserving the existing API and result ordering.

- Adds `DiskAnnIndex::search_with`, which accepts a caller-owned `VisitedSet` and routes directly through `VamanaGraph::greedy_search_fast`.
- Replaces the unreachable preallocated `visited` field with a bounded `std::sync::Mutex<Vec<VisitedSet>>` pool. The normal single-threaded path checks out and returns one preallocated set. An empty or contended pool falls back to a fresh set, and the pool retains at most four sets.
- Makes generation rollover safe by fully zeroing the visited storage when the `u64` generation reaches its maximum.
- Handles index-size mismatches by resizing and fully resetting the caller's visited storage. Repeated searches at the same size retain the O(1) generation-counter reset.
- Adds deterministic equivalence coverage across 100 queries against both `search_with` and the pre-fix allocating algorithm shape.
- Adds a counting-allocator regression test covering 100 hot pooled searches after warmup.

Fixes #677

## Measurements

`bench_search_hot` builds seeded random unit vectors with `d=128`, then measures 2,000 single-threaded searches at `k=10` in a release build. Build settings are `R=4`, `L_build=8`, `L_search=64`, and `alpha=1.0`. The table reports the median of three runs per configuration. Additional raw samples are retained where visible host variance prompted repeat collection.

| Vectors | Metric | Before | After |
| ---: | --- | ---: | ---: |
| 100,000 | p50 latency | 41.333 us | 29.042 us |
| 100,000 | p99 latency | 301.417 us | 116.208 us |
| 100,000 | Queries/sec | 18,111.622 | 27,236.237 |
| 100,000 | Visited-state allocation/query | 812,504 bytes | 0 bytes |
| 250,000 | p50 latency | 54.666 us | 33.084 us |
| 250,000 | p99 latency | 486.834 us | 116.417 us |
| 250,000 | Queries/sec | 5,524.658 | 22,139.261 |
| 250,000 | Visited-state allocation/query | 2,031,256 bytes | 0 bytes |

Allocation values are the exact backing storage requested by the old `VisitedSet::new` path: the `n * size_of::<u64>()` generation array plus the bit-vector storage. The after value is the steady-state visited-state allocation after one warmup. The regression test additionally asserts that 100 hot searches allocate zero blocks at or above the 2,000,000-byte generation-array size and less than 300,000 bytes total for all other search output and candidate storage combined.

Hardware: Apple M2 Max, 32 GB RAM, macOS 26.6. Searches were single-threaded and binaries used the Cargo release profile.

Reproduce with `cargo run --release -p ruvector-diskann --example bench_search_hot`. The benchmark commit predates the fix commit on this branch, so before numbers come from checking out the benchmark commit alone.

## Compatibility and safety

The public `search` signature and behavior remain unchanged. The new `search_with` API allows explicit reuse where callers want to own query scratch state. Search candidates, exact re-ranking, and ordering are unchanged.

The pool uses only the standard library, adds no dependency, and does not assume threads exist. This keeps the single-thread pool path suitable for `wasm32`. The upstream crate baseline currently fails `wasm32-unknown-unknown` compilation in `getrandom` because its JavaScript feature is not enabled, so no new wasm target regression was introduced or masked here.

## Verification

- `cargo fmt -p ruvector-diskann`
- `cargo clippy -p ruvector-diskann --all-targets -- -D warnings`
- `cargo test -p ruvector-diskann`
- `cargo build --release -p ruvector-diskann --examples`, on both the unfixed and fixed source


ohdearquant:
CI note — the red checks on this PR are not caused by this diff:

- **`Clippy (deny warnings)`**: pre-existing failure in `crates/ruvector-rabitq/src/persist.rs:194` (`useless_borrows_in_formatting` on `MAGIC, &magic`), present on `main` — unrelated to this branch.
- **`dependency-review (PRs only)`**: errors with *"Dependency review is not supported on this repository. Please ensure that Dependency graph is enabled"* — a repo Settings toggle, fails on every PR.
- **`Security audit` / `cargo audit` / `cargo deny`**: these are the RustSec advisories cleared by #682. This branch was cut before those dependency bumps, so it still carries them. Merging #682 first (or rebasing this branch on top of it) turns the audit gates green.

The DiskANN build and test jobs for this change all pass.

PR #684, opened 2026-07-12
fix(diskann): tombstone deletes with vector preservation, local graph repair, and delete_deferred escape hatch

### LINKED ISSUE #679: diskann: delete() can return deleted ids, injects NaN into result ordering, and degrades recall — tombstone filter + local graph repair
## Summary

`delete()` fills the vector with NaN and removes the id from `id_reverse`, but leaves the graph adjacency and the `id_map` entry untouched (index.rs:207-216, distance.rs:46-51). Three concrete consequences, all verified at main `c55e9e89`:

1. **Deleted ids can still be returned by search.** `search()` re-ranks through `self.id_map[id as usize]` (index.rs:196), which still holds the deleted id. Nothing in the search path consults `id_reverse` or any tombstone set.
2. **NaN distances make result ordering arbitrary.** Every distance against a deleted point is NaN, and both the beam search and the re-rank sort use `partial_cmp(...).unwrap_or(Ordering::Equal)` (index.rs:190, graph.rs), so NaN entries compare Equal to everything and land at arbitrary positions, including inside the top-k.
3. **Recall decays as deletes accumulate.** Edges into and out of deleted nodes are never repaired, so points whose graph paths ran through deleted nodes become hard or impossible to reach. Long-lived mutating deployments (exactly the browser/edge sessions targeted by #676) degrade monotonically with no recovery short of a full rebuild.

## Suggested fix, in two independent stages

**Stage 1 (correctness, small):** keep a tombstone bitset. Beam search may still traverse tombstoned nodes (as routing waypoints) but must never emit them: filter tombstones from the candidate list before re-rank, and skip NaN by construction (distance against tombstoned nodes is only needed for routing, and with a bitset you can keep the original vector data instead of NaN-filling, which also removes the NaN-ordering hazard entirely). Deleted ids never appear in results; ordering stays total.

**Stage 2 (quality, medium): local repair on delete.** For each in-neighbor u of a deleted node d, splice d's out-neighbors into u's candidate set and `robust_prune(u, neighbors(u) + neighbors(d) \ {d}, alpha)`. This is O(in-degree x degree) per delete, preserves graph connectivity, and avoids any rebuild. The crate already has `robust_prune`; the only new requirement is an in-neighbor lookup (either a reverse adjacency list or a bounded scan). We ship this two-hop repair in khive's vamana crate (`ohdearquant/khive`, `crates/khive-vamana`), along with a batch tombstone-only escape hatch for callers that prefer deferred repair, and can port both as a PR.

## Measurable impact / acceptance

- After deleting any id, a search for its exact vector never returns the deleted id (today it can).
- With vector data preserved and tombstones filtered, no NaN ever enters a distance computation; sorts are total.
- Recall@10 after deleting a random 20% of a 100k index stays within 1-2 points of a fresh build over the surviving 80% (with stage 2 repair); tombstone-only (stage 1) documents its measured degradation instead of corrupting silently.
- `count()` and persisted state reflect deletions consistently.

ohdearquant:
## Summary

- Preserve deleted vectors and track deletion state in a tombstone bitset so distance calculations remain finite and deterministic.
- Filter tombstoned candidates before exact re-ranking so deleted IDs cannot appear in search results.
- Repair the graph locally on normal deletes, while exposing `delete_deferred()` for callers that prefer tombstone-only deletion and later batch cleanup.
- Persist tombstones and report the live vector count consistently across save and load.

Fixes #679

## Design

Deletion now marks the internal vector position in a packed tombstone set and removes the external ID from the live reverse map. The original vector values remain unchanged. This removes the prior NaN distance hazard and lets tombstoned nodes remain valid routing waypoints when deletion is deferred. Search filters tombstones before exact re-ranking and uses a total floating-point ordering.

The default `delete()` path performs local graph repair. Because this crate does not maintain reverse adjacency, it finds the deleted node's in-neighbors by scanning the bounded-degree adjacency in O(n * degree) time per deletion. For each live in-neighbor, it combines the node's existing candidates with the deleted node's live out-neighbors, removes deleted candidates, and applies the existing alpha robust-prune selection with the configured maximum degree. The deleted node is then removed from all adjacency. Tombstoned sources and candidates are short-circuited during the scan.

`delete_deferred()` is the low-latency escape hatch. It records the tombstone without modifying graph edges, allowing callers to defer repair or rebuild work. Unknown IDs and repeated deletes return `DiskAnnError::NotFound`.

## Measurements

Recall is recall@10 against brute-force ground truth over surviving vectors after seeded random deletion of 20% of the index. Main is unmeasurable because deleted IDs can be returned, invalidating survivor-only recall.

| Vectors | Main | Tombstone only | Tombstone + repair | Fresh rebuild |
| ---: | ---: | ---: | ---: | ---: |
| 20,000 | Unmeasurable | 0.9940 | 0.9960 | 0.9960 |
| 100,000 | Unmeasurable | 0.9680 | 0.9760 | 0.9820 |

| Vectors | Tombstone-only median delete | Repair median delete |
| ---: | ---: | ---: |
| 20,000 | 0.875 us | 670.917 us |
| 100,000 | 1.166 us | 1,392.917 us |

Measured on an Apple M2 Max, single-threaded, with the release profile. The raw machine-readable output is produced by `bench_delete_recall`.

## Validation

- `cargo fmt -p ruvector-diskann`
- `cargo clippy -p ruvector-diskann --all-targets -- -D warnings`
- `cargo test -p ruvector-diskann`

## WASM

The implementation uses only the Rust standard library, adds no dependencies, and remains compatible with the crate's WASM builds.


ohdearquant:
CI note — the red checks on this PR are not caused by this diff:

- **`Clippy (deny warnings)`**: pre-existing failure in `crates/ruvector-rabitq/src/persist.rs:194` (`useless_borrows_in_formatting` on `MAGIC, &magic`), present on `main` — unrelated to this branch.
- **`dependency-review (PRs only)`**: errors with *"Dependency review is not supported on this repository. Please ensure that Dependency graph is enabled"* — a repo Settings toggle, fails on every PR.
- **`Tests (core-and-rest)`**: this did not fail an assertion — the job hit the 4-hour runner limit (wall-clock 4h0m18s) and was killed. The other test shards (core-and-rest-heavy, -wasm, ml-research-*, vector-index, etc.) all pass. Re-running the job clears it; I've kicked a re-run.

The DiskANN build jobs and remaining test shards for this change all pass.

PR #691, opened 2026-07-14
fix(diskann): mmap read-through vector storage (#674)

### LINKED ISSUE #674: diskann: load() materializes the whole vector slab in RAM — serve through the mmap instead
Companion to #673 (PQ-guided search), same source read of `ruvector-diskann` on main (c55e9e89).

**ADR-144 says**: "mmap load — OS pages in only accessed vectors — working set << total dataset."

**What the code does today:** `load()` memory-maps `vectors.bin` and then copies every float into a heap `Vec` (`index.rs:322-342` — per-chunk `f32::from_le_bytes` into `flat_data`). The retained `Mmap` handle is never read after that loop, so the entire dataset is RAM-resident. The index is currently an in-memory Vamana index with file persistence; the SSD-backed working-set story doesn't hold yet.

**Proposal:** make `FlatVectors` storage-agnostic — an owned `Vec<f32>` variant (current behavior, used during insert/build) and an mmap-slab variant that reads vector slices directly out of the map. The on-disk format already cooperates: little-endian flat f32 data at byte offset 16, so slices are 4-byte aligned and can be exposed zero-copy (bytemuck cast) on little-endian targets. Graph adjacency and PQ codes stay heap-resident — they're small relative to the vectors.

Combined with #673, per-hop traversal cost moves to PQ tables in RAM and only the final re-rank touches mmapped full vectors — which is the actual DiskANN serving architecture from the paper, and what makes 100M+ vector indexes on a laptop realistic.

Happy to contribute this as a PR. It would include a page-cache-cold vs warm latency table so the on-SSD behavior is documented from measurement rather than asserted.

ohdearquant:

## Summary

Fixes #674. `DiskAnnIndex::load()` currently memory-maps `vectors.bin` and then
copies every float into a heap `Vec` — the retained `Mmap` handle was never read
again after that copy loop, so the whole dataset ends up RAM-resident regardless
of dataset size. That contradicts the SSD-backed working-set design ADR-144
documents ("mmap load — OS pages in only accessed vectors — working set << total
dataset").

This PR makes the flat vector slab (`FlatVectors`) storage-agnostic:

- `VectorStorage::Owned(Vec<f32>)` — the existing behavior, used while
  inserting/building and by the default `load()` path (unchanged).
- `VectorStorage::Mmap { mmap, data_offset }` — new, read-through: `get()` reads
  vector slices directly out of the mapped file, so RSS stays proportional to the
  accessed working set instead of the whole dataset.

`DiskAnnIndex::load()` is unchanged (still materializes fully — default,
back-compat). A new `DiskAnnIndex::load_mmap()` opts into the read-through path.
Graph adjacency, PQ codes, and IDs are unaffected — they stay heap-resident (small
relative to the vector slab), matching the issue's proposal.

## Safety

`from_mmap` fails closed (returns `Err`, never transmutes) if the vector slab's
byte offset isn't 4-byte aligned, or if the map is too short for `count * dim`
floats — see `FlatVectors::from_mmap` in `distance.rs`. In practice the on-disk
layout keeps the flat f32 data at a fixed 16-byte header offset (mmap base is
page-aligned on every platform memmap2 supports), so this should never trip in
normal use; it's there so a corrupted or hand-edited file fails loudly instead of
producing UB.

## Delete/tombstone parity

`Owned` storage keeps writing `NaN` in place on `delete()` (unchanged, zero extra
cost). `Mmap` storage can't mutate the mapped file, so it tracks deletes in a
small owned tombstone overlay instead — `get()` returns the same all-`NaN` row in
both modes, so callers see identical behavior regardless of backing storage. See
`test_load_mmap_delete_matches_owned_tombstone` for the parity test.

## Measured A/B (methodology + full pre-registration in the linked gist/comment)

Pre-registered thresholds (fixed before measuring): open only if (a) peak RSS
(load + 1k queries) in Mmap mode ≤ 50% of Owned mode, (b) warm-cache search p50
regresses ≤ 10% vs Owned, (c) all crate tests pass in both modes.

Fixture: N=500,000 vectors, dim=128, f32, unit-normalized. Hardware: Apple M2 Max,
32 GB RAM, macOS/APFS-SSD. `cargo build --release`. Bench harness:
`examples/mmap_674_bench.rs` (included in this PR — `build`/`measure`
subcommands). RSS via `/usr/bin/time -l` (macOS), whole-process peak, load +
1,000 queries (50 warmup, untimed). **Warm-cache only** — this sandbox has no
root, so `/usr/sbin/purge` (page-cache drop) isn't available; a cold-cache table
is a possible follow-up but isn't gating (threshold (b) is explicitly warm-cache
scoped).

The 500k-vector A/B run is executing now; I'll post the completed table (peak RSS,
load time, warm search p50/p99, QPS for both modes, against the pre-registered
thresholds above) as a follow-up comment. Correctness in both storage modes is
already covered by the crate tests (27/27, including the delete-parity test); the
harness and exact repro below make the perf numbers independently reproducible.

Repro:

```bash
cargo build --release -p ruvector-diskann --example mmap_674_bench
./target/release/examples/mmap_674_bench build --dir /tmp/674fixture --n 500000 --dim 128 --max-degree 32 --build-beam 64
/usr/bin/time -l ./target/release/examples/mmap_674_bench measure --dir /tmp/674fixture --mode owned --dim 128
/usr/bin/time -l ./target/release/examples/mmap_674_bench measure --dir /tmp/674fixture --mode mmap  --dim 128
```

## Migration notes

- **Default unchanged**: `DiskAnnIndex::load()` behaves exactly as before
  (materializes vectors into RAM). No caller-visible change unless you opt in.
- **Opt-in**: call `DiskAnnIndex::load_mmap()` instead of `load()` to get
  read-through storage.
- **Read-only after `load_mmap()`**: `insert()`/`insert_batch()` on an
  mmap-loaded index return `DiskAnnError::InvalidConfig` instead of panicking —
  build a fresh index (or use `load()`) if further inserts are needed.
  `delete()` works identically in both modes.
- No new dependencies — `memmap2` and `bytemuck` were already crate dependencies.

## Test plan

- [x] All pre-existing `ruvector-diskann` tests pass unchanged
- [x] New tests: mmap alignment/size validation (fail-closed), read-through
      correctness, tombstone-not-file-mutation, `load()`/`load_mmap()` parity
      (identical search results across 20 random queries), insert-rejected-on-mmap
- [x] `cargo clippy -p ruvector-diskann --all-targets --no-deps -- -D warnings`
      clean
- [x] `cargo fmt -p ruvector-diskann --check` clean

### DIFF
--- a/crates/ruvector-diskann/examples/mmap_674_bench.rs
+++ b/crates/ruvector-diskann/examples/mmap_674_bench.rs
@@ -1,0 +1,151 @@
+//! A/B measurement harness for #674 (mmap read-through vector storage).
+//!
+//! Two phases, run as separate process invocations so `/usr/bin/time -l` (or any
+//! external RSS sampler) captures one phase's peak RSS cleanly:
+//!
+//! ```text
+//! cargo run --release --example mmap_674_bench -- build   --dir DIR --n 500000 --dim 128
+//! cargo run --release --example mmap_674_bench -- measure --dir DIR --mode owned --dim 128
+//! cargo run --release --example mmap_674_bench -- measure --dir DIR --mode mmap  --dim 128
+//! ```
+//!
+//! `measure` loads the index (`DiskAnnIndex::load` for `owned`,
+//! `DiskAnnIndex::load_mmap` for `mmap`), runs a warmup batch, then times a
+//! measured batch and prints a single `RESULT ...` line with p50/p99 latency and
+//! QPS. Peak RSS is read externally (this binary does not self-report memory —
+//! `/usr/bin/time -l` on macOS or `/usr/bin/time -v` on Linux is the intended
+//! wrapper) so the number reflects the whole process (load + queries), not just
+//! an in-process allocator counter that would miss mmap'd pages.
+
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
+use ruvector_diskann::{DiskAnnConfig, DiskAnnIndex};
+use std::hint::black_box;
+use std::path::PathBuf;
+use std::time::Instant;
+
+fn random_unit_vector(rng: &mut StdRng, dim: usize) -> Vec<f32> {
+    let mut v: Vec<f32> = (0..dim).map(|_| rng.gen_range(-1.0..1.0)).collect();
+    let norm = v.iter().map(|x| x * x).sum::<f32>().sqrt();
+    for x in &mut v {
+        *x /= norm;
+    }
+    v
+}
+
+fn get_flag(args: &[String], name: &str, default: &str) -> String {
+    for w in args.windows(2) {
+        if w[0] == name {
+            return w[1].clone();
+        }
+    }
+    default.to_string()
+}
+
+fn main() {
+    let args: Vec<String> = std::env::args().collect();
+    let cmd = args.get(1).map(String::as_str).unwrap_or("");
+    match cmd {
+        "build" => cmd_build(&args[2..]),
+        "measure" => cmd_measure(&args[2..]),
+        _ => {
+            eprintln!(
+                "usage:\n  mmap_674_bench build   --dir DIR --n N --dim D [--max-degree R] [--build-beam L]\n  mmap_674_bench measure --dir DIR --mode owned|mmap --dim D [--queries Q] [--warmup W] [--k K]"
+            );
+            std::process::exit(2);
+        }
+    }
+}
+
+fn cmd_build(args: &[String]) {
+    let dir = PathBuf::from(get_flag(args, "--dir", "/tmp/mmap674"));
+    let n: usize = get_flag(args, "--n", "500000").parse().expect("--n");
+    let dim: usize = get_flag(args, "--dim", "128").parse().expect("--dim");
+    let max_degree: usize = get_flag(args, "--max-degree", "32")
+        .parse()
+        .expect("--max-degree");
+    let build_beam: usize = get_flag(args, "--build-beam", "64")
+        .parse()
+        .expect("--build-beam");
+
+    // Fixed seed so the fixture is reproducible across build/measure invocations
+    // and across owned-vs-mmap comparison runs.
+    let mut rng = StdRng::seed_from_u64(0x6D_6D_61_70_36_37_34);
+    let mut index = DiskAnnIndex::new(DiskAnnConfig {
+        dim,
+        max_degree,
+        build_beam,
+        search_beam: 64,
+        alpha: 1.2,
+        storage_path: Some(dir.clone()),
+        ..Default::default()
+    });
+
+    let t0 = Instant::now();
+    for id in 0..n {
+        index
+            .insert(id.to_string(), random_unit_vector(&mut rng, dim))
+            .expect("insert failed");
+        if id % 50_000 == 0 {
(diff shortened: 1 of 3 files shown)


ohdearquant:
## Measured A/B — 500k×128, results

Methodology, verbatim from the pre-registration in the PR body: N=500,000 vectors, dim=128,
f32, unit-normalized; Apple M2 Max, macOS/APFS-SSD; `cargo build --release`; harness
`examples/mmap_674_bench.rs` (in this PR); RSS via `/usr/bin/time -l`, whole-process peak,
**load + 1,000 queries** (50 warmup, untimed); **warm-cache only** (no `purge` in this
sandbox — no root). Repro commands are in the PR body and reproduce these numbers.

| Metric | Owned | Mmap | Mmap / Owned |
|---|---:|---:|---:|
| max resident set size | 713.1 MiB | 468.8 MiB | 65.7% (−34.3%) |
| peak memory footprint (phys_footprint) | 469.4 MiB | 224.2 MiB | 47.8% (−52.2%) |
| search p50 | 554.8 µs | 556.0 µs | +0.21% |
| search p99 | 981.6 µs | 991.6 µs | +1.02% |
| QPS | 1730.6 | 1727.7 | −0.17% |
| load time | 0.249 s | 0.093 s | 2.7× faster |

Raw: `RESULT mode=owned ... load_s=0.249 p50_us=554.792 p99_us=981.583 qps=1730.568` /
`RESULT mode=mmap ... load_s=0.093 p50_us=555.958 p99_us=991.625 qps=1727.651`; `time -l`
`maximum resident set size` 747,782,144 vs 491,520,000 B; `peak memory footprint`
492,159,720 vs 235,078,376 B.

### Against the pre-registered thresholds

- **(b) p50 regression ≤ 10%: MET** — +0.21% (latency-neutral; p99 +1.0%, QPS flat). load is 2.7× faster.
- **(c) tests pass in both modes: MET** — 27/27, including the `load`/`load_mmap` search-parity and delete-tombstone-parity tests.
- **(a) peak RSS ≤ 50% of owned: metric-dependent, not cleanly met on gross RSS.**
  - On `peak memory footprint` (phys_footprint — anonymous + dirty, the memory-pressure-relevant figure): **47.8% ≤ 50% ✓**.
  - On `maximum resident set size` (which also counts the clean, file-backed mapped pages): **65.7% ✗**.

I'm reporting this straight rather than picking the flattering metric. The gap between the two
is the **warm-cache** measurement condition, disclosed in the PR body up front: 1,000 queries
over beam search page in most of the 256 MB vector slab, and those pages stay resident in a
warm run. They are counted in gross RSS but are precisely the clean, file-backed pages the OS
reclaims first under pressure — which is the entire point of the read-through design. On
`phys_footprint`, which excludes them, the halving target is hit.

### Honest read

Latency-neutral, loads 2.7× faster, cuts phys_footprint ~52% and gross RSS ~34% on this warm
500k fixture. The full working-set-≪-dataset benefit that motivated the issue would show under
a cold cache or a query distribution touching a smaller fraction of vectors; this sandbox can't
drop the page cache to measure that. So: a correct, opt-in, latency-neutral read-through path
with a real (but, warm-cache, partial) memory win — not a clean 2× RSS reduction. Framing it as
the former would misrepresent the measurement.

Issue #659, opened 2026-07-10
rvf: ingestBatch() silently drops documented per-vector metadata; RvfSearchResult has no metadata field


pacphi:
## Summary

`RvfDatabase.ingestBatch()` in `@ruvector/rvf@0.2.3` silently drops per-vector `metadata` passed by the caller — the field is documented as part of the public API contract but is never forwarded to the native call, and the resolved promise gives no signal that anything was dropped. There is also a related, architecturally independent limitation: string ids require an external `.idmap.json` sidecar file, so `.rvf` can never be a true single-file artifact when using string ids.

## Environment

- `@ruvector/rvf@0.2.3` (current npm `latest`)
- Node v26.4.0, macOS arm64, native `@ruvector/rvf-node` addon

## How to reproduce

```js
import { RvfDatabase } from '@ruvector/rvf';
const db = await RvfDatabase.create('./repro.rvf', { dimensions: 4 });
await db.ingestBatch([
  { id: 'doc1', vector: new Float32Array([1, 0, 0, 0]), metadata: { title: 'Hello World' } },
]);
const results = await db.query(new Float32Array([1, 0, 0, 0]), 1);
console.log(results);
await db.close();
```

## Actual output

```text
[ { id: 'doc1', distance: 0 } ]
```

No `metadata`/`title` field anywhere on the result. Reproduced live against the currently installed `@ruvector/rvf@0.2.3`. The same repro also produced a `repro.rvf.idmap.json` sidecar containing `{"idToLabel":{"doc1":1},"labelToId":{"1":"doc1"},"nextLabel":2}`, confirming the string-id mapping issue described below.

## Expected behavior

Either `metadata` is serialized and forwarded to the native `ingestBatch` call (consistent with `deleteByFilter()` elsewhere in the same file, which does `JSON.stringify(filter)` before crossing the N-API boundary), or, if the native N-API layer genuinely has no metadata storage support yet, `ingestBatch()` should throw/warn when a caller passes non-empty `metadata`, or the public type/docs should drop the `metadata?` field until it's implemented.

## Root cause (confirmed by reading source)

`dist/backend.js`, `NodeBackend.ingestBatch()` (lines ~115–148). The method computes `ids` via `resolveLabel()` (line 135) but calls `this.handle.ingestBatch(flat, ids)` (line 136) — no metadata array is ever built or passed. `entries[i].metadata` is read nowhere in the function body.

[...]

Issue #658, opened 2026-07-10
rvf: RvfDatabase.create() throws misleading FsyncFailed error when target path already has a valid file


pacphi:
## Summary

`RvfDatabase.create()` in `@ruvector/rvf@0.2.3`, when pointed at a path that already holds a valid `.rvf` file, throws a misleading `FsyncFailed` error that implies a hardware/OS durability problem, instead of a clear precondition-check error indicating the file already exists.

## Environment

- `@ruvector/rvf@0.2.3` (current npm `latest`)
- Node v26.4.0, macOS arm64, native `@ruvector/rvf-node` addon

## How to reproduce

```js
import { RvfDatabase } from '@ruvector/rvf';
const db1 = await RvfDatabase.create('./library.rvf', { dimensions: 384 });
await db1.ingestBatch([/* ... */]);
await db1.close();

// forgot to fs.rm the file, or intended open() but called create() again:
const db2 = await RvfDatabase.create('./library.rvf', { dimensions: 384 }); // (!)
```

## Actual output

```text
RvfError: Durable write (fsync) failed: RVF error 0x0303: FsyncFailed
```

(error code `771`)

Reproduced live: created a store, closed it, called `create()` again on the same path — same misleading message as above.

## Expected behavior

A clear "file already exists — use `RvfDatabase.open()` or pass an `{ overwrite: true }` option" error. An `FsyncFailed` code strongly implies a hardware/OS durability problem (disk full, permissions, filesystem fault); in this case it is purely "the file already existed and had content." This class of message sends debugging down the wrong path — toward disk/filesystem investigation — before the real, much simpler cause (wrong static method for the situation) is found.

## Root cause (confirmed by reading source)

`dist/backend.js`, `NodeBackend.create()` (line 80) calls `this.native.create(path, mapOptionsToNative(options))` directly — there is no `fs.existsSync`/pre-check anywhere in the JS wrapper before the native call. The native layer instead fails partway through its write/durability sequence and surfaces a low-level fsync error code for what is actually a precondition-check failure.

## Sample use cases affected

- Re-running an index-build script against a path that already holds a valid `.rvf` file from a previous run — an easy, common mistake (e.g. forgetting to delete the old file, or a Ctrl-C'd build that left a partial file that a second run then hits)

## Workaround

Always `fs.rm(path, { force: true })` before `create()` for a guaranteed-fresh build, or call `.open()` when the file is expected to already exist.

## Proposed fix

[...]

PR #696, opened 2026-07-14
fix(rvf-sdk): ergonomic query count arg + clear create-on-existing error (#657, #658)

### LINKED ISSUE #657: rvf: RvfDatabase.query() throws misleading napi type error on options-object call form ({k}/{topK}/{limit})
## Summary

`RvfDatabase.query()` in `@ruvector/rvf@0.2.3` throws a misleading low-level N-API error when called with an options-object second argument (`{k: 2}`, `{topK: 2}`, or `{limit: 2}`), instead of either accepting the object form or validating the argument type before crossing the N-API boundary.

## Environment

- `@ruvector/rvf@0.2.3` (current npm `latest`, no 0.2.4+ published as of this report; latest published 2026-07-05)
- Node v26.4.0, macOS arm64, native `@ruvector/rvf-node` addon

## How to reproduce

```js
import { RvfDatabase } from '@ruvector/rvf';
const db = await RvfDatabase.create('./test.rvf', { dimensions: 8 });
await db.ingestBatch([{ id: 'doc1', vector: new Float32Array(8).fill(0), metadata: {} }]);

await db.query(new Float32Array(8), { k: 2 });      // throws
await db.query(new Float32Array(8), { topK: 2 });   // throws
await db.query(new Float32Array(8), { limit: 2 });  // throws
await db.query(new Float32Array(8), 2);              // works
```

## Actual output (all three object forms)

```text
RvfError: Backend initialization failed: Failed to convert napi value Object into
rust type `u32`
```

(error code `65281`)

Reproduced live against the currently installed native addon.

## Expected behavior

Either the object form is accepted (most vector-DB JS APIs — including, going by naming conventions used elsewhere in this same package, plausibly this one — accept a `{k}`-style options object), or the wrapper validates arguments in the TypeScript/JS layer and throws a clear, correctly-scoped error *before* crossing the N-API boundary. "Backend initialization failed" is actively misleading — it reads as a database-open-time failure, not an argument-type error on a specific call.

## Root cause (confirmed by reading source)

`dist/database.js` declares `query(vector, k, options)` (line 109). `dist/backend.js`, `NodeBackend.query()` (line 149) passes `k` straight through to `this.handle.query(vector, k, nativeOpts)` with no type validation in the JS layer — an object passed as `k` reaches the native N-API boundary unchanged, where napi-rs's raw type-conversion error ("Failed to convert napi value Object into rust type `u32`") surfaces directly to the caller.

## Sample use cases affected

- k-nearest-neighbor query against a local `.rvf` vector store, e.g. serving a `/api/search?mode=semantic`-style endpoint where `k` is naturally expressed as part of an options object alongside filters

## Workaround

Call with `k` as a bare positional number: `db.query(vector, k)`.

## Proposed fix

Accept both call signatures for ergonomics (`query(vector, k)` and `query(vector, { k })`), or at minimum validate the `k` argument's type in `dist/database.js`/`dist/backend.js` and throw a typed `RvfError` (e.g. `InvalidArgument`) carrying the actual accepted signature in its message, before calling into the native addon. Document the accepted signature explicitly in the README/type defs, which currently don't make it unambiguous.

ohdearquant:

Fixes #657 and #658 — two `@ruvector/rvf` JS SDK papercuts where a mistaken call
form leaked a misleading low-level error instead of a clear, actionable one. Both
fixes live in the SDK layer (`database.ts` / `backend.ts`); no native rebuild.

## #657 — `query()` options-object form

`RvfDatabase.query(vector, k, options)` accepted only a positional numeric `k`.
Passing an options object (`query(vec, { k: 2 })`, or `{topK}`/`{limit}`, as most
vector-DB JS APIs allow) put the object in the `k` slot, which crossed N-API and
threw `Failed to convert napi value Object into rust type u32`.

Now `query()` accepts **both** forms and validates before dispatch:
- `query(vec, 10)` — positional (unchanged).
- `query(vec, { k: 10, efSearch: 200 })` — object form; `topK` and `limit` are
  accepted as aliases for `k`; the remaining fields pass through as query options.
- A missing count, or a non-positive-integer count, throws a clear
  `RvfError` with the new `RvfErrorCode.InvalidArgument` instead of the N-API error.

## #658 — `create()` on an existing path

`create()` on a path that already held a valid `.rvf` surfaced the native
`FsyncFailed` (durability) error — implying a disk/OS fault when the real cause was
just "the file already exists." `NodeBackend.create()` now checks first and throws a
clear `RvfError` with the new `RvfErrorCode.FileExists`, pointing at `open()` or a new
`{ overwrite: true }` option. `overwrite` removes the old file **and** its
`.idmap.json` sidecar before creating, so no stale string-id map leaks in.

## Additive / compatible

- Two new SDK-level (`0xFF__`) error codes: `FileExists = 0xff04`,
  `InvalidArgument = 0xff05`. No existing code changed.
- New optional `RvfOptions.overwrite` (default `false` = current behavior).
- `query()`'s positional-number form is unchanged; the object form is new surface.
- `dist/` regenerated with `tsc` (this package commits its build output; `files`
  publishes `dist/` only). Stale `src/*.js` artifacts are not referenced by
  `package.json` and were left untouched.

## Verification

`scripts/smoke-query-create.mjs` (also `npm run smoke`) exercises both fixes against
the built `dist/`. #657 runs with no native addon (mock backend via `fromBackend`);
#658 runs when a native/stub addon is resolvable.

```
  ok - #657 positional count query(v, 5) -> 5
  ok - #657 object {k:2} -> 2
  ok - #657 object {topK:3} -> 3
  ok - #657 object {limit:4} -> 4
  ok - #657 object {k, efSearch} forwards efSearch
  ok - #657 rejects no count with clear InvalidArgument (not napi error)
  ok - #657 rejects negative with clear InvalidArgument (not napi error)
  ok - #657 rejects non-integer with clear InvalidArgument (not napi error)
  ok - #657 rejects zero with clear InvalidArgument (not napi error)
  ok - #658 create() on existing path -> FileExists (clear, not FsyncFailed)
  ok - #658 create({overwrite:true}) clears old file + sidecar and proceeds

All 11 checks passed.
```

`tsc --noEmit` (strict) is clean. Note: `#659` (ingestBatch metadata) is a related but
separate change touching the native ingest/query path — not included here.


ruvnet:
Reviewed this (independently, not just the PR description — checked out the diff and traced the actual code paths). Found a blocker that needs a rebase before this can merge:

**Hard error-code collision.** This PR adds `FileExists = 0xff04` in `errors.ts`, but PR #708 (merged today) already assigned `MetadataNotSupported = 0xff04` to that exact slot for an unrelated error (rejecting unsupported per-vector metadata on ingest). `InvalidArgument = 0xff05` is currently free on main, so only `FileExists` collides. Two semantically unrelated errors sharing a numeric code means `err.code === 0xff04` / `RvfErrorCode[0xff04]` reverse-mapping become ambiguous for callers. GitHub already reports this PR as `CONFLICTING`/`DIRTY` against current main, consistent with this.

Suggest rebasing onto current main and moving `FileExists` to the next free slot (`0xff08`, since #697 — a sibling PR touching the same enum — reserves `0xff06`/`0xff07`).

Two non-blocking things worth a look while rebasing:
- `{topK: 5, limit: 10}` (both aliases set) resolves via `k ?? topK ?? limit`, so `topK` silently wins over `limit` — works, but undocumented; a one-line JSDoc note would help.
- The `overwrite: true` path in `create()` does two separate `rmSync` calls (main file, then `.idmap.json` sidecar) gated on `fs.existsSync(path)` only. If the sidecar exists but the main file doesn't (e.g. from an earlier interrupted create), the existence check never fires and a stale sidecar can survive a fresh `create()`, desyncing ID mappings. Also, if the first `rmSync` succeeds but the second throws, you're left with the original file destroyed and no new file in a half-cleaned state. Might be worth a single guarded cleanup path.

Happy to help rebase this if useful — just didn't want to push to your branch without asking first.

Issue #634, opened 2026-07-02
intelligence.json silently wiped under concurrent hook processes: non-atomic save() + load() swallowing parse errors into empty defaults


KevinAIP:
## Summary

`.ruvector/intelligence.json` can be silently and permanently wiped when multiple ruvector processes run concurrently — which is the normal state under Claude Code, where every PreToolUse/PostToolUse/SessionStart hook is a separate short-lived `ruvector hooks …` process sharing the store. Verified present in **0.2.33**.

Three compounding defects in `bin/cli.js`:

### 1. `save()` is a non-atomic overwrite

`Intelligence.save()` ends in `fs.writeFileSync(this.intelPath, JSON.stringify(this.data, null, 2))`. A concurrent reader can observe a torn, half-written file. The six standalone command sites that do `fs.writeFileSync(dataPath, …)` have the same problem.

### 2. `load()` swallows parse errors into empty defaults

`load()`'s body is wrapped in `try { … } catch {}` and falls through to `return defaults`. So the reader that hit the torn file from (1) silently starts from an **empty store** — and its own `save()` then persists the emptiness. Weeks of accumulated memories/trajectories/patterns are replaced without any error surfaced (the triggering hook even prints `{"success":true}`).

Deterministic repro (no race needed):
```bash
# truncate the store mid-file to simulate what a concurrent reader sees
head -c 100000 .ruvector/intelligence.json > t && mv t .ruvector/intelligence.json
npx ruvector hooks remember "hello" -t test   # prints success:true
# store now contains 1 memory — everything else is gone
```
Probabilistic repro (the production path): run ~20 parallel `hooks remember` against a several-hundred-KB store in a loop; larger stores widen the write window until a reader eventually lands mid-write. We lost a 433-memory store twice in one week this way (both times during a scheduled headless-Claude burst).

### 3. Seconds-granularity ids collide

`remember()` uses `mem_${this.now()}` (epoch **seconds**; same for completed-trajectory ids at `traj_${this.now()}`). Concurrent or same-second writes collide — our organically-grown store contained 433 memory records with only 428 unique ids. Anything that later dedups or keys by id silently drops records. The `trajectory-begin` path already does `traj_${Date.now()}_${random}` — the same format works here.

Separately, even without corruption, concurrent load-modify-save is last-writer-wins: in a 100-parallel-write test only 31 writes survived. Not a crash, but it quietly defeats the "learning compounds" premise for any multi-process setup.

[...]

PR #697, opened 2026-07-14
fix(rvf-sdk): atomic id-map sidecar write + fail-loud on corrupt load (#635)

### LINKED ISSUE #635: @ruvector/rvf: saveMappings()/loadMappings() swallow errors on the load-bearing .idmap.json sidecar — silent no-op deletes, label collisions on corrupt load
Sibling of #634 — same durability defect class (silent swallow + non-atomic write + corrupt-load-to-empty-defaults), different package: `@ruvector/rvf`'s NodeBackend id-map sidecar. Verified on `@ruvector/rvf@0.2.2` (dist is byte-identical to 0.2.1 for `backend.js`).

## Context

Since 0.2.0, `NodeBackend` maintains a string↔i64-label map persisted as a `.idmap.json` sidecar next to the store. This sidecar is **load-bearing**: `delete(ids)` resolves string ids through `idToLabel` and silently filters out anything unresolvable, so with a missing/corrupt/stale sidecar it returns `{deleted: 0}` with no error. Two swallows make that failure mode easy to hit and impossible to see:

## Defect 1 — `saveMappings()` (dist/backend.js:381)

```js
fs.writeFileSync(mp, data, 'utf-8');   // non-atomic
...
catch {
    // Non-fatal: mapping persistence is best-effort (e.g. read-only FS).
}
```

- A failed write is silently discarded → every vector ingested since the last good save becomes undeletable by id, with zero signal.
- A torn write (crash/ENOSPC mid-`writeFileSync`) leaves invalid JSON on disk, which then triggers defect 2 on the next open.

Mapping persistence isn't "best-effort" when `delete()` correctness depends on it.

## Defect 2 — `loadMappings()` (dist/backend.js:399)

```js
catch {
    // Non-fatal: start with empty mappings.
}
```

A corrupt sidecar silently degrades to empty maps, which is worse than failing:

- `nextLabel` resets to `idToLabel.size + 1` = **1**, so subsequent `ingestBatch` calls assign labels that collide with existing vectors' labels — silent data corruption, not just data loss.
- `delete()` silently no-ops for every historical id.
- The next `saveMappings()` **overwrites** the (recoverable) corrupt sidecar with the empty/colliding state, destroying the evidence.

## Suggested fix (patch we're running in production)

Atomic tmp+rename in `saveMappings()` with the error surfaced; quarantine + throw on corrupt load instead of empty-defaults. Applied via patch-package on 0.2.2:

<details><summary>patches/@ruvector+rvf+0.2.2.patch</summary>

```diff
diff --git a/node_modules/@ruvector/rvf/dist/backend.js b/node_modules/@ruvector/rvf/dist/backend.js
index 64ffa6c..fdf1973 100644
--- a/node_modules/@ruvector/rvf/dist/backend.js
+++ b/node_modules/@ruvector/rvf/dist/backend.js
@@ -382,17 +382,29 @@ class NodeBackend {
         const mp = this.mappingsPath();
         if (!mp)
             return;
+        // AIP PATCH (#8 / ADR-0017 recipe): the sidecar is load-bearing —
+        // without it delete() silently no-ops for every historical id. The
+        // stock build wrapped a non-atomic writeFileSync in a bare catch{},
+        // so a failed or torn write silently orphaned vectors. Write
+        // atomically (tmp + rename) and surface failures instead.
+        const fs = await Promise.resolve().then(() => __importStar(require('fs')));
+        const data = JSON.stringify({
+            idToLabel: Object.fromEntries(this.idToLabel),
+            labelToId: Object.fromEntries(Array.from(this.labelToId.entries()).map(([k, v]) => [String(k), v])),
+            nextLabel: this.nextLabel,
+        });
+        const tmp = `${mp}.tmp.${process.pid}`;
         try {
-            const fs = await Promise.resolve().then(() => __importStar(require('fs')));
-            const data = JSON.stringify({
-                idToLabel: Object.fromEntries(this.idToLabel),
-                labelToId: Object.fromEntries(Array.from(this.labelToId.entries()).map(([k, v]) => [String(k), v])),
-                nextLabel: this.nextLabel,
-            });
-            fs.writeFileSync(mp, data, 'utf-8');
+            fs.writeFileSync(tmp, data, 'utf-8');
+            fs.renameSync(tmp, mp);
         }
-        catch {
-            // Non-fatal: mapping persistence is best-effort (e.g. read-only FS).
+        catch (err) {
+            try {
+                fs.rmSync(tmp, { force: true });
+            }
+            catch { }
+            console.error(`[rvf] failed to persist id-map sidecar at ${mp}: ${err?.message ?? err}`);
+            throw err;
         }
     }
     /** Load the string↔label mapping from the sidecar JSON file if it exists. */
@@ -400,17 +412,29 @@ class NodeBackend {
         const mp = this.mappingsPath();
         if (!mp)
             return;
+        const fs = await Promise.resolve().then(() => __importStar(require('fs')));
+        if (!fs.existsSync(mp))
+            return;
+        // AIP PATCH (#8 / ADR-0017 recipe): a corrupt sidecar must not
+        // silently degrade to empty mappings — that resets nextLabel to 1
+        // (label collisions with existing vectors) and turns delete() into
+        // a silent no-op for every historical id. Quarantine the bad file
+        // and fail loudly so recovery (sidecar resynth or --rebuild-store)
+        // is a deliberate act, not an accident.
         try {
-            const fs = await Promise.resolve().then(() => __importStar(require('fs')));
-            if (!fs.existsSync(mp))
-                return;
             const raw = JSON.parse(fs.readFileSync(mp, 'utf-8'));
             this.idToLabel = new Map(Object.entries(raw.idToLabel ?? {}).map(([k, v]) => [k, Number(v)]));
             this.labelToId = new Map(Object.entries(raw.labelToId ?? {}).map(([k, v]) => [Number(k), v]));
             this.nextLabel = raw.nextLabel ?? this.idToLabel.size + 1;
         }
-        catch {
-            // Non-fatal: start with empty mappings.
+        catch (err) {
+            const quarantine = `${mp}.corrupt.${Date.now()}`;
+            try {
+                fs.renameSync(mp, quarantine);
+            }
+            catch { }
+            console.error(`[rvf] id-map sidecar at ${mp} is unreadable (${err?.message ?? err}); quarantined to ${quarantine}`);
+            throw err;
         }
     }
 }
```

</details>

## Related observations from the same verification pass

- `@ruvector/rvf@0.2.x` declares `"@ruvector/rvf-node": "^0.1.7"`, so a stock install resolves a 0.1.x native layer — worth checking whether `deleteByFilter` (added to the SDK in 0.2.0) exists in the native it actually gets. (`@ruvector/rvf-node@0.2.0`'s own optionalDependencies pin `darwin-arm64` at 0.1.7 as well.)
- The SDK's `ingestBatch` never forwards entry metadata to the NAPI layer (`this.handle.ingestBatch(flat, ids)` — the `metadata?` param from the signature comment is dropped), so `deleteByFilter` has nothing to filter on for SDK-ingested stores.

Happy to PR the sidecar durability patch if useful.

ohdearquant:

Fixes #635. Sibling of #634 (same durability defect class, different package).

The `@ruvector/rvf` `NodeBackend` persists its string↔i64-label map as a
`.idmap.json` sidecar. That sidecar is **load-bearing**: `delete(ids)` resolves
string ids through `idToLabel` and silently filters out anything unresolvable, so a
missing/torn/corrupt sidecar returns `{ deleted: 0 }` with no error. Two swallows
made that failure mode both easy to hit and invisible.

## Defect 1 — `saveMappings()` swallowed write failures and wrote non-atomically

- A failed write was silently discarded → every vector ingested since the last good
  save became undeletable-by-id, with zero signal.
- A torn write (crash / ENOSPC mid-`writeFileSync`) left invalid JSON on disk, which
  then triggered Defect 2 on the next open.

**Fix:** write to a temp file then `rename()` over the sidecar (atomic publish — a
partial write can never appear at the real path), and raise `SidecarWriteFailed`
instead of swallowing. `delete()`/ingest correctness depends on the map, so
persistence is not best-effort.

## Defect 2 — `loadMappings()` degraded a corrupt sidecar to empty maps

Empty maps are worse than a hard failure:
- `nextLabel` resets to `1`, so the next `ingestBatch` assigns labels that collide
  with existing vectors — silent corruption, not just loss.
- `delete()` silently no-ops for every historical id.
- The next `saveMappings()` overwrites the (recoverable) corrupt sidecar, destroying
  the evidence.

**Fix:** quarantine the corrupt sidecar (rename it aside so it is not clobbered) and
throw `SidecarCorrupt`, so the caller learns string-id operations are unsafe. A
missing sidecar (fresh store) is still treated as legitimate and loads empty.

## Compatibility

- Two new SDK-level error codes: `SidecarWriteFailed = 0xff06`,
  `SidecarCorrupt = 0xff07`. No existing code changed (`0xff04`/`0xff05` left reserved
  to avoid clashing with an in-flight ergonomics PR).
- Happy path is unchanged: normal ingests still write the sidecar and reopen restores
  state. Behavior only changes on the previously-silent failure paths.
- `dist/` regenerated via `tsc` (`files` publishes `dist/` only).

## Verification

`scripts/smoke-idmap-durability.mjs` (`npm run smoke:idmap`) against the built `dist/`:

```
  ok - #635 ingest persists sidecar atomically (nextLabel=3, no .tmp)
  ok - #635 reopen restores maps; no label collision on next ingest
  ok - #635 corrupt sidecar -> SidecarCorrupt + quarantined (no silent empty-map reset)

All 3 checks passed.
```

`tsc --noEmit` (strict) is clean.


ruvnet:
Reviewed this (checked out the diff, traced the actual code, and diffed against current main's `backend.ts`/`errors.ts`) — one rebase blocker and one real bug found, independent of each other:

**1. Needs rebase.** `git apply --check --3way` against current main fails on `src/errors.ts` (and the generated `dist/errors.*`, `dist/backend.js.map`) — PR #708 (merged today) already inserted `MetadataNotSupported = 0xff04` into the same enum block your diff expects to find empty. Good news: no *numeric* collision — `SidecarWriteFailed = 0xff06` / `SidecarCorrupt = 0xff07` are both still free — it's purely a textual conflict from the enum/message-map insertion point shifting. `src/backend.ts` itself applies cleanly (confirms #708 didn't touch `saveMappings`/`loadMappings`).

**2. Real bug in the atomic-write logic, unrelated to the conflict.** `saveMappings()`'s temp file is `${mp}.tmp` — no PID or timestamp suffix. Two concurrent `NodeBackend` instances writing the same store path race: both `writeFileSync` the same temp path, then both `renameSync(tmp, mp)` — the second rename either silently publishes whichever process wrote last under the other's identity, or throws `ENOENT` because the first rename already consumed the temp file. There's no lock guarding this write (the core's `LockHeld`/`LockStale` codes don't cover this JS-side sidecar). Suggest something like `${mp}.${process.pid}.${Date.now()}.tmp`.

The quarantine-on-corrupt-read path looks correct as-is — verified the `SidecarCorrupt` throw is unconditional even if the quarantine rename itself fails (no silent proceed-with-corrupt-file case).

Happy to help rebase + fix the temp-file naming if useful — didn't want to push to your branch without asking first.

ohdearquant:
@ruvnet yes please help with rebase

PR #722, opened 2026-07-26
test(mcp): make `mcp test` prove the server answers, not just that files exist

markt-heximal:
Stacked on #721 — the live check added here **fails without that fix**, which is the point. Review/merge #721 first; this PR's diff against it is the three files below.

## Problem

`ruvector mcp test` printed:

```
  Total: 97 tools registered
  All checks passed.
```

against a server that never started and never answered a single request.

Every check was static — the file exists, `node -c` parses it, `@modelcontextprotocol/sdk` resolves, and the tool count is a regex over the source text. Nothing launched the server, so "transport never connected" and "healthy" were indistinguishable. That output is what made the outage in #721 look fine.

## Change

- **`mcp test` gains a live handshake.** It spawns `mcp start` the way a client does, sends `initialize`, and requires a JSON-RPC reply — then fails if stdout carried anything that is not JSON-RPC, since stdio MCP requires stdout to be exclusively newline-delimited JSON-RPC. On failure it exits non-zero.
- **`test/mcp-handshake.js`** covers the same ground in the suite, wired into `npm test`.

The test drives `bin/cli.js`, not `bin/mcp-server.js`. Run as main, `mcp-server.js` always worked — a direct probe would have passed throughout the outage. Testing the entry point users actually invoke is the whole point.

## A race worth calling out

Both the CLI check and the test drain stdout for a grace period after the last reply rather than settling on it. The ONNX loader's chatter is emitted *after* the initialize response, so returning immediately raced past the very output the check exists to catch.

This was not theoretical: the first version of this test **passed** with the stdout bug deliberately reintroduced. The drain window is what makes that assertion real.

## Verification

Each fix reverted independently, on Linux arm64 / node v22.22.0:

| State | `mcp test` | `test/mcp-handshake.js` |
|---|---|---|
| Launch bug restored | exits **1** — "no response within 45s (transport never connected?)" | **1/4 passed** |
| Stdout bug restored | — | **3/4 passed**, names the offending line |
| Both fixes present | exits **0** | **4/4 passed** |

Full suite on this branch: **87 passed, 0 failed** (73 + 8 + 4 + 2), `npm test` exit 0.

Note `npm install` must be run with `--workspaces=false` inside `packages/ruvector` on non-darwin hosts — `packages/*` includes darwin-only NAPI stubs and npm enforces `os`/`cpu` across workspace members. Unrelated to this change, but it costs every Linux/Windows contributor their first hour.

### DIFF
--- a/npm/packages/ruvector/bin/cli.js
+++ b/npm/packages/ruvector/bin/cli.js
@@ -8394,7 +8394,7 @@ mcpCmd.command('tools')
 
 mcpCmd.command('test')
   .description('Test MCP server setup and tool registration')
-  .action(() => {
+  .action(async () => {
     console.log(chalk.bold.cyan('\nMCP Server Test Results'));
     console.log(chalk.dim('-'.repeat(40)));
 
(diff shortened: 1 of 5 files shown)


ruvnet:
The unique live-handshake and tools/list regression coverage has been preserved on current main in #726, with original authorship retained for Mark Tenenholtz. #724 already merged the underlying MCP startup fix, so this conflicting stacked PR is now superseded.

PR #761, opened 2026-08-02
Saturate the SIMD tanh approximation in the AVX2 activation kernels

ohdearquant:
## The problem

The AVX2 activation kernels in `crates/ruvector-sparse-inference/src/backend/cpu.rs` approximate `tanh` with the rational form

```
tanh(z) ~= z * (27 + z^2) / (27 + 9*z^2)
```

That form is accurate near zero but it does not saturate. As `z` grows it tends to `z/9` rather than to 1, so GELU grows as roughly `0.002 * x^4` instead of tending to `x`.

Working the constants through at `x = 20`:

```
inner       = 0.7978845608 * (20 + 0.044715 * 8000) = 301.4
tanh_approx = 301.4 * (27 + 90818) / (27 + 817362)  = 33.5
result      = 0.5 * 20 * (1 + 33.5)                 = 344.9      (true GELU(20) = 20.0)
```

The error is not confined to extreme inputs. It is about 12% at `x = 5` and about 196% at `x = 10`, both of which are ordinary activation magnitudes in a transformer.

There is a second, compounding effect. The scalar remainder loop at the end of the same function calls `f32::tanh`, which saturates correctly. So a single call to `activation()` returns different values for the same input value depending only on whether that element landed in one of the first `len - (len % 8)` SIMD lanes or in the tail.

## The fix

Clamp the rational result to `tanh`'s true `[-1, 1]` range, in the AVX2 path via `_mm256_min_ps`/`_mm256_max_ps` and in the portable helper via `clamp`. That makes the two paths agree across the input range, which is the property that actually matters here rather than any particular choice of approximation.

## Second issue: FMA dispatch

`gelu_avx2`, `silu_avx2`, `dot_product_avx2` and `axpy_avx2` all emit `_mm256_fmadd_ps`, but they were dispatched on `features.has_avx2` alone and `dot_product_avx2` declared only `#[target_feature(enable = "avx2")]`. Executing an FMA-requiring body on an AVX2-without-FMA CPU is undefined behaviour. AVX2 and FMA have shipped together since Haswell and Zen so this is unlikely to bite on real silicon, but the gate did not say what the code requires. Each dispatch site now requires FMA and the kernels declare it.

## Tests

The rational approximation is also exposed as `fast_tanh_approx`, an architecture-independent helper that mirrors the SIMD lanes exactly, so its saturation property is testable on hosts that cannot execute AVX2:

- `test_fast_tanh_approx_saturates` sweeps large magnitudes and asserts the result stays inside `[-1, 1]`.
- `test_gelu_fast_scalar_matches_exact_gelu` sweeps `x` in `[-30, 30]` against the real-tanh scalar path, max abs error < 0.03.
- `test_silu_fast_scalar_matches_exact_silu` does the same for SiLU, < 0.05.

Thresholds were measured rather than guessed: an intentionally strict bound was used first to force a failure and read the true maxima (tanh 0.0235, GELU 0.0217, SiLU 0.0406), then set with margin.

`cargo test -p ruvector-sparse-inference`: 94 passed, 0 failed, plus 10 in `backend_simd_tests`.

Mutation-checked, so the test is load-bearing rather than decorative. Removing the clamp gives:

```
test backend::cpu::tests::test_fast_tanh_approx_saturates ... FAILED
fast_tanh_approx(-300) = -33.34222 is outside [-1, 1]
```

and restoring it returns the suite to green.

## One honest limitation

**The AVX2 kernels themselves were not executed.** This work was done on an aarch64 host, where all `#[cfg(target_arch = "x86_64")]` code is compiled out entirely, so the changed intrinsics were never run. What is covered is the approximation's mathematics through the portable mirror, plus the dispatch-gate change by inspection.

On an x86_64 host with AVX2+FMA, `cargo test -p ruvector-sparse-inference` exercises `activation()` through `gelu_avx2`/`silu_avx2` via `tests/backend_simd_tests.rs`. If your CI covers x86_64 that will close the gap; if it would help, I am happy to adjust anything the x86 run turns up.

### DIFF
--- a/crates/ruvector-sparse-inference/src/backend/cpu.rs
+++ b/crates/ruvector-sparse-inference/src/backend/cpu.rs
@@ -42,7 +42,7 @@ impl Backend for CpuBackend {
         #[cfg(target_arch = "x86_64")]
         {
             let features = get_simd_features();
-            if features.has_avx2 {
+            if features.has_avx2 && features.has_fma {
                 return unsafe { dot_product_avx2(a, b) };
             } else if features.has_sse41 {
                 return unsafe { dot_product_sse(a, b) };
@@ -101,14 +101,14 @@ impl Backend for CpuBackend {
             }
             ActivationType::Gelu => {
                 #[cfg(target_arch = "x86_64")]
-                if features.has_avx2 {
+                if features.has_avx2 && features.has_fma {
                     return unsafe { gelu_avx2(data) };
                 }
                 gelu_scalar(data);
             }
             ActivationType::Silu | ActivationType::Swish => {
                 #[cfg(target_arch = "x86_64")]
-                if features.has_avx2 {
+                if features.has_avx2 && features.has_fma {
                     return unsafe { silu_avx2(data) };
                 }
                 silu_scalar(data);
@@ -134,8 +134,11 @@ impl Backend for CpuBackend {
         debug_assert_eq!(a.len(), b.len());
 
         #[cfg(target_arch = "x86_64")]
-        if get_simd_features().has_avx2 {
-            return unsafe { axpy_avx2(a, b, scalar) };
+        {
+            let features = get_simd_features();
+            if features.has_avx2 && features.has_fma {
+                return unsafe { axpy_avx2(a, b, scalar) };
+            }
         }
 
         for (x, y) in a.iter_mut().zip(b.iter()) {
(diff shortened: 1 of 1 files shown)


ohdearquant:
Closing this — I'd rather bring you the analysis than a patch that papers over it, and I want to re-scope what I'm proposing here. The defect write-up stands and I'll follow up separately; nothing here is blocked on you in the meantime.

PR #800, opened 2026-08-05
feat(rvforge): create command, authoring core, and Reader install/library/update

ruvnet:
Completes the pieces previously listed as *implemented but not integrated*, and closes the end-to-end gap in the published 0.1.0.

## The gap this closes

`init` printed `Next: rvforge pack <agent.rvf>` while creating no such file. A first-time user's next command failed with `FORGE_E_IO` (exit 10), and there was no supported way to produce the `.rvf` that every other command reads — the only valid one in the repo lived in `tests/fixtures/`, which is not in the published tarball.

`create` closes that. Verified from an empty directory against the built CLI:

```
init --keygen      → forge.config.json, rvforge.json, publisher key   exit 0
create             → agent.rvf, 4.5 KB, signed, 3 capabilities        exit 0
validate --deep    → valid RVF, signature verified                    exit 0
test               → 5 passed, 0 failed                               exit 0
```

`init`'s next-step line now points at `create` rather than at a file it does not write.

## What lands

| Component | Change | Tests |
|---|---|---|
| `rvf-forge-core` | `author` module — `ContainerBuilder` writes signed containers this crate's own verifier accepts | 138 |
| `@ruvector/rvforge` | `create` command, `rvf-writer` | 253 |
| `rvforge-reader` | install, library, update-with-rollback, dock bridge | 189 |

## Properties held by construction

**Segment kind decides signing policy, not the caller.** A `.wasm` payload becomes an executable `WASM` segment and is signed individually; anything else becomes an opaque `VEC` segment. Rule 3 of the loading contract — unsigned executable segments rejected by default — is a property of the writer rather than something each caller must remember to request.

**Every stage re-verifies.** An artifact that verified at download is verified again at install and again at load, because the file on disk between those points is not the same object the earlier check covered.

**The Dock trust boundary is typed.** System-owned chrome (trust badge, network indicator, pause) is populated from `SystemOwnedStatus` only; agent text stays in `AgentProvidedStatus` and is sanitized before display. A hostile agent has no channel to forge an approved badge or claim it stopped while running.

**Updates bind to lineage.** An update whose base identity does not match the installed artifact is refused rather than applied; rollback restores the previous version with its state capsule intact.

**Reproducible authoring.** `--from <dir>` visits files in sorted order, so the same input directory produces the same bytes.

The parity fixture generator now builds its input through the `author` module rather than a bespoke byte layout, so the TypeScript and Rust sides are compared against a shared definition of a valid container instead of two hand-written ones.

## Verification

```
cargo test -p rvf-forge-core                 138 passed
npm test (@ruvector/rvforge)                 253 passed
cargo test (rvforge-reader workspace)        189 passed
clippy --all-targets -- -D warnings          clean
```

`rvforge-reader` is a standalone cargo workspace, so its tests run from its own directory — `cargo test -p rvforge-reader` from the repo root reports "package did not match".

https://claude.ai/code/session_01ParP55bZs2iTGEGvpnUecx


ruvnet:
### Security-audit failures were pre-existing, now fixed

`cargo audit` and `Security audit` failed on the first run. They were **not** caused by this PR — three advisories were published against `rkyv 0.8.16`, which is already on `main`:

- [RUSTSEC-2026-0233](https://rustsec.org/advisories/RUSTSEC-2026-0233) — use-after-free during deserialization of crafted archives
- [RUSTSEC-2026-0234](https://rustsec.org/advisories/RUSTSEC-2026-0234) — out-of-bounds read, archives containing hash tables
- [RUSTSEC-2026-0235](https://rustsec.org/advisories/RUSTSEC-2026-0235) — out-of-bounds read, archives containing `Rc`/`Arc`

Evidence it is pre-existing: this PR's diff against `main` contains **no `Cargo.toml` or `Cargo.lock` change**, and `rkyv 0.8.16` is present in `main`'s lockfile. It reaches the tree through `ruvector-core`, `ruvector-graph`, `ruvector-router-core` and `ruvector-sparse-inference` — none of them rvforge.

Fixed here in `aec12c17c` by bumping to `0.8.18` rather than adding ignore entries. `.cargo/audit.toml` states the policy — *anything fixable should be fixed via a dependency bump rather than ignored* — and all three advisories are on the deserialization path, which is exactly where untrusted bytes arrive.

Lockfile only, 16 lines. `cargo audit` exits 0; the four dependent crates check clean.

I bundled a dependency fix into a feature PR, which is normally worth avoiding. Flagging it explicitly: it is a separate commit, and it was blocking this PR's CI while also leaving `main` red.

PR #810, opened 2026-08-08
fix(graph-node): wire NAPI binding to persistent storage (hydrate on open, edge/hyperedge write-through, delete cascade)

### LINKED ISSUE #809: @ruvector/graph-node: storagePath data silently lost on restart (open() never hydrates, edges never persisted)
## Summary

`@ruvector/graph-node` (v2.0.4) accepts a `storagePath` and writes node records to redb, but the NAPI binding never wires the rest of persistence, so **any graph silently loses edges — and on reopen, everything — after a process restart**:

1. **`GraphDatabase.open(path)` never hydrates.** It opens the redb file but returns an empty graph (`stats()` → `{totalNodes: 0, totalEdges: 0}`) even when the file contains data. The crate already has `GraphDB::with_storage()` / `load_from_storage()`; the binding just never calls them.
2. **`createEdge` / `createHyperedge` / `batchInsert` edges never reach storage.** Only nodes are persisted; every edge lives solely in the in-memory hypergraph.
3. **`deleteNode` doesn't cascade through storage**, leaving dangling persisted edges.

## Repro

```js
const { GraphDatabase } = require('@ruvector/graph-node');
const Z = new Float32Array(4);

// process 1
const db = new GraphDatabase({ distanceMetric: 'Cosine', dimensions: 4, storagePath: './t.db' });
await db.createNode({ id: 'a', labels: ['T'], properties: {}, embedding: Z });
await db.createNode({ id: 'b', labels: ['T'], properties: {}, embedding: Z });
await db.createEdge({ from: 'a', to: 'b', description: 'CALLS', confidence: 1, embedding: Z });
console.log(await db.stats()); // { totalNodes: 2, totalEdges: 1 }

// process 2 (restart)
const db2 = GraphDatabase.open('./t.db');
console.log(await db2.stats()); // { totalNodes: 0, totalEdges: 0 }  <-- data gone
```

## Root cause

In `crates/ruvector-graph-node/src/lib.rs`:
- `new()`/`open()` construct `GraphDB::new()` and never replay storage into the hypergraph or property graph.
- `create_edge`/`create_hyperedge`/`batch_insert` only call `hypergraph.add_hyperedge(...)`; there is no `storage.insert_edge`/`insert_hyperedge` write-through.
- `delete_node` deletes only the node record from storage.

The storage layer itself (`ruvector-graph`'s redb backend, `all_node_ids`/`all_edge_ids`/`all_hyperedge_ids`, `insert_edge`, `insert_hyperedge`) works correctly — it's purely a binding gap.

## Fix

I have a working patch (PR to follow, referencing this issue) that:
- hydrates hypergraph + property graph in `new()`/`open()` (nodes, binary edges, multi-node hyperedges; embeddings/confidence round-tripped as `__embedding`/`__confidence` FloatArray properties)
- writes edges/hyperedges through to storage in `createEdge`/`createHyperedge`/`batchInsert`
- makes `deleteNode` cascade through persisted edges/hyperedges and `deleteEdge` clean the traversal index
- makes hydration tolerant of dangling records (skip + report) rather than failing `open()`

Verified with a 12-gate suite (create / reopen in fresh process / delete persistence / batch / tx / Cypher / kHop / in-memory mode / single-writer lock) and a real workload: a 154,711-node / 353,496-edge code graph that now survives restart with exact counts (hydrate ~15s).

Happy to adjust the approach if you'd prefer hydration to live in the crate rather than the binding.

Bolivar1234:
Fixes #809.

## What

Wires the `ruvector-graph-node` NAPI binding to the persistence the `ruvector-graph` crate already provides, so `storagePath` graphs survive process restarts.

- **Hydration**: `new()`/`open()` now replay storage into both the hypergraph (traversal index) and the property graph — nodes, binary edges, and multi-node hyperedges. `__embedding`/`__confidence` are round-tripped as FloatArray properties so embeddings survive too.
- **Write-through**: `createEdge`, `createHyperedge`, and `batchInsert` edges now persist via `storage.insert_edge`/`insert_hyperedge` and mirror into the property graph (previously they only touched the in-memory hypergraph).
- **Deletes**: `deleteEdge` also cleans the hypergraph traversal index; `deleteNode` cascades through persisted edges/hyperedges touching the node, so a later hydrate never replays dangling references.
- **Tolerant hydrate**: dangling records (e.g. written by a pre-fix binary) are skipped and counted with a stderr report instead of failing `open()`.

One file changed: `crates/ruvector-graph-node/src/lib.rs` (+227/−9). No crate/API changes; existing in-memory (no `storagePath`) behavior is untouched.

## Verification

12-gate suite, each phase in its own process because redb is single-writer:

| Gate | Result |
|---|---|
| create nodes+edge / multi-node hyperedge / batchInsert / tx / Cypher label scan | PASS |
| fresh-process reopen (exact stats) | PASS |
| reopen kHop over binary, hyper and batch edges | PASS |
| deleteEdge / deleteNode(cascade) persist across reopen | PASS |
| in-memory mode unaffected | PASS |
| single-writer lock still enforced on concurrent open | PASS |

Scale check: a 154,711-node / 353,496-edge code graph (12 repos) reopens with exact counts after restart (hydrate ~15s on a shared host). Before this patch the same reopen returned `{totalNodes: 0, totalEdges: 0}`.

Happy to move hydration into the crate instead of the binding if you prefer that layering, or to add the test suite as a `__test__` file in the package.

### DIFF
--- a/crates/ruvector-graph-node/src/lib.rs
+++ b/crates/ruvector-graph-node/src/lib.rs
@@ -14,7 +14,10 @@ use ruvector_core::advanced::hypergraph::{
 };
 use ruvector_core::DistanceMetric;
 use ruvector_graph::cypher::{parse_cypher, Statement};
+use ruvector_graph::edge::Edge as GraphEdge;
+use ruvector_graph::hyperedge::Hyperedge as GraphHyperedge;
 use ruvector_graph::node::NodeBuilder;
+use ruvector_graph::types::PropertyValue;
 use ruvector_graph::storage::GraphStorage;
 use ruvector_graph::GraphDB;
 use std::collections::HashMap;
@@ -67,7 +70,7 @@ fn register_node(
     properties: Option<HashMap<String, String>>,
 ) -> Result<()> {
     // 1. Adjacency / vector index (kHop, stats, hyperedge search).
-    hg.add_entity(id.clone(), embedding);
+    hg.add_entity(id.clone(), embedding.clone());
 
     // 2. Property graph + label index (Cypher label scan).
     let mut builder = NodeBuilder::new().id(&id);
@@ -81,6 +84,8 @@ fn register_node(
             builder = builder.property(&key, value);
         }
     }
+    // Persist the vector so hydrate_from_storage can rebuild the hypergraph index.
+    builder = builder.property("__embedding", embedding);
     let graph_node = builder.build();
 
     // Persist to storage if enabled (mirrors create_node behaviour).
@@ -123,14 +128,16 @@ impl GraphDatabase {
             (None, None)
         };
 
-        Ok(Self {
+        let db = Self {
             hypergraph: Arc::new(RwLock::new(CoreHypergraphIndex::new(core_metric))),
             causal_memory: Arc::new(RwLock::new(CoreCausalMemory::new(core_metric))),
             transaction_manager: Arc::new(RwLock::new(transactions::TransactionManager::new())),
             graph_db: Arc::new(RwLock::new(GraphDB::new())),
             storage,
             storage_path,
-        })
+        };
+        db.hydrate_from_storage()?;
+        Ok(db)
     }
 
     /// Open an existing graph database from disk
@@ -146,14 +153,16 @@ impl GraphDatabase {
 
         let metric = DistanceMetric::Cosine;
 
-        Ok(Self {
+        let db = Self {
             hypergraph: Arc::new(RwLock::new(CoreHypergraphIndex::new(metric))),
             causal_memory: Arc::new(RwLock::new(CoreCausalMemory::new(metric))),
             transaction_manager: Arc::new(RwLock::new(transactions::TransactionManager::new())),
             graph_db: Arc::new(RwLock::new(GraphDB::new())),
             storage: Some(Arc::new(RwLock::new(storage))),
             storage_path: Some(path),
-        })
+        };
+        db.hydrate_from_storage()?;
+        Ok(db)
     }
 
     /// Check if persistence is enabled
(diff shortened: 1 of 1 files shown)


Bolivar1234:
Pushed a follow-up commit (bbceebb0) fixing a bug in my original patch: `PropertyValue::from(Vec<f32>)` resolves to the generic `From<Vec<T>>` impl and produces `Array(Float(f64)...)`, never `FloatArray` — so hydrate (which matched only `FloatArray`) replayed every embedding as empty, and `searchHyperedges` returned f32::MAX sentinel scores for everything after a restart. Traversal was unaffected, which is why the kHop gates missed it.

Fix: persist sites write `PropertyValue::FloatArray` explicitly, and hydrate uses a tolerant extractor that also reads the boxed Array/List form, so databases written by the earlier revision stay readable. Verified with a new 'vector search survives reopen' gate (13/13 passing) plus a 16K-node real workload returning identical scores pre/post restart.

ruvnet:
Thank you for this, @Bolivar1234 — and sorry it took us a while to get back to you.

Your diagnosis here was correct and genuinely valuable. `@ruvector/graph-node` really was silently dropping every node, edge, and hyperedge on restart: the NAPI layer opened `GraphStorage` but never replayed it into the in-memory indexes, and edge/hyperedge writes never made it back to disk. That's a data-loss bug in a package whose whole point is persistence, and you found it with a clear reproduction. Your follow-up analysis of the `PropertyValue::from(Vec<f32>)` generic-impl dispatch issue was equally sharp — that one is subtle and easy to miss.

**Why we're closing this rather than merging it**

The fix landed on `main` on 2026-08-12 in commit 31bb944, in a hardened form with test coverage. That work covers the same ground as this branch, which means this branch is now *behind* `main` rather than ahead of it — and merging it at this point would revert several of those hardenings:

1. **Cascade would become unconditional again.** On `main`, `deleteNode` only removes an edge from durable storage when `cascade: true`; with `cascade: false` the relationship survives. This branch deletes edges from storage unconditionally, so a `cascade: false` delete would destroy durable relationship data — a regression, and a silent one.

2. **The test that guards exactly that would be deleted.** `non_cascade_retains_durable_relationships_but_cascade_removes_them` (`crates/ruvector-graph-node/src/lib.rs:1076`) exists specifically to pin this behavior. This branch removes 166 lines of tests including that one, so the regression in (1) would land unnoticed.

3. **Storage errors would go back to being swallowed.** The cascade path on `main` propagates storage failures up to the caller; this branch discards them, so a failed durable delete would look like a success to JS.

There are smaller deltas too — the cascade on `main` is O(degree) via the adjacency index rather than O(E) over all edges, which matters at the graph sizes you're working with — and the branch doesn't pass `cargo fmt`.

The branch is also currently in conflict with `main` (mergeable state: `CONFLICTING`) and never got a CI run, so there's no green build behind it either.

**What we're carrying forward from your report**

The 154,711-node hydration timing you shared is the most useful thing in this thread that *isn't* fixed yet. `hydrate_from_storage` is called synchronously from the `#[napi(constructor)]` and from `open()`, so on a graph that size it blocks the Node event loop for the full ~15s. We've opened a separate issue for async/lazy hydration and credited this PR and you on it.

Please do keep filing these. The persistence bug was real, your reproduction made it easy to confirm, and the fact that the fix arrived from a different direction doesn't change that you're the reason we knew to look.

---
*This review was performed by Claude on behalf of the maintainer.*

**Update:** the follow-up issue for the synchronous constructor hydration is now filed as #826.

PR #847, opened 2026-08-20
docs(adr): Perpetual Intelligence Runtime ADRs

ruvnet:
## Summary

Adds ADR-305 through ADR-315 — the eleven architecture decisions for the RuV Perpetual Intelligence Runtime (PIR) program — plus the program's research grounding:

- `docs/research/perpetual-intelligence-runtime/01-evidence-review.md` — arXiv-sourced evidence grades for the program's eight founding claims (six grade-A, two UNVERIFIED)
- `docs/research/perpetual-intelligence-runtime/02-asset-map.md` — mechanism-to-ruvnet-component map
- `docs/research/perpetual-intelligence-runtime/03-program-plan.md` — bounded contexts, ADR list, work packages, security gates, acceptance-test design
- `docs/research/perpetual-intelligence-runtime/04-verification-addendum.md` — direct source-verification findings against `ruvnet/LatentMesh`, `ruvnet/autogenous`, and `ruvnet/ruflo` that correct or extend several asset-map claims without editing it in place

### The eleven ADRs

1. **ADR-305** — Adopt autogenous ADR-401 ("The Perpetual Intelligence Machine," Accepted) as the program's definition of record, and LatentMesh ADR-009's control loop as the Latent Communication Fabric's design contract — rather than inventing a parallel architecture. Also adopts ruflo ADR-322B's proposer/promotion separation-of-powers invariant as a seventh governing invariant.
2. **ADR-306** — Adopt `ruvnet/dream-machine`'s Accepted, shipped evaluation engine (ADR-0001) as the promotion-evaluation core, wired to `ruvector`'s existing `research-gate`/`sota-bench` statistics (ADR-282) and Darwin — consolidating four existing "Dream Machine" implementations rather than building a fifth. Preserves ADR-0001's principle verbatim: "Evaluation is not promotion — the machine never merges; a human does."
3. **ADR-307** — Three-level persistent memory (LiveMem + TARL pattern), mapped onto `ruvllm`'s existing memory modules and `ruvector-agent-memory`, extending the proof-gated writes that already exist (ADR-194/047) with TARL's five-operation ledger.
4. **ADR-308** — WorldCycle-style reversible-action verification for the physical action loop.
5. **ADR-309** — Greenfield LatentMesh integration crates inside `ruvector`, coordinated on wire format with the (small, ~1.4K-LOC) external LatentMesh repo — explicitly not "finishing an existing integration."
6. **ADR-310** — Causal-attribution CI gate for latent communication, modeled on arXiv:2607.26773 (correctly dated 2026-07-29).
7. **ADR-311** — Anomaly quarantine for latent channels — explicitly documented as net-new work informed by arXiv:2606.28958, since "LATTE" is not a real paper.
8. **ADR-312** — A shared witness record schema and cross-layer anchoring contract (anchored on ruflo ADR-322C), not a merge of the `rvm-witness` and autogenous `witness` crates, which target incompatible runtimes.
9. **ADR-313** — SHAPER-pattern frozen-weight skill/harness evolution loop via Darwin, structurally enforced, blocked on WP0b's two known dependency bugs.
10. **ADR-314** — KV-cache cross-model migration in ruvLLM (fast-follow, independent track, grade-A evidence).
11. **ADR-315** — Governance constitution for capability expansion, built on autogenous's admission-gate pattern with an explicit API-stability buffer given its self-labeled research-prototype status.

All ADRs are numbered 305–315, above the repo's true existing maximum (**ADR-304**, confirmed by scanning every filename in `docs/adr/` rather than assuming numeric uniqueness — duplicates exist elsewhere in the sequence, e.g. two ADR-266 files). Status: Proposed on all eleven. Existing duplicate ADR numbers are intentionally left untouched — a separate hygiene work item tracks that.

> **Note on the diff**: this PR's diff against `main` may show `npm/core/platforms/win32-x64-msvc/ruvector.node` as changed. That is not from either commit on this branch — this branch was cut from `origin/main` at a point that already included commit `74d2a6017` ("chore: Update NAPI-RS binaries for all platforms"), which the current `main` doesn't have yet, so the three-dot diff surfaces it as a base-divergence artifact. This branch's own two commits (`b3c1e12b0`, `357d4dc30`) are docs-only.

## Test plan

- [ ] ADR numbering confirmed non-colliding against `docs/adr/` at merge time (numbers may need re-verification if other ADR PRs land first)
- [ ] WP1 (per ADR-305) reads autogenous ADR-401 and LatentMesh ADR-009 in full against source and confirms this program's adopt/reconcile framing
- [ ] Cross-repo coordination threads opened with `ruvnet/LatentMesh`, `ruvnet/autogenous`, and `ruvnet/ruflo` maintainers before implementation work begins


ruvnet (commented):
## Adversarial review — PIR ADR set (ADR-305 … ADR-315 + research/)

**Verdict: CHANGES-REQUIRED** — 3 blocking, 10 non-blocking, 4 nits.

Method: fetched `feat/pir-adrs`, read all 15 changed files, and verified every cross-repo claim against direct clones of `ruvnet/LatentMesh` (HEAD `922ea196`, 2026-08-18), `ruvnet/autogenous` (HEAD `b5c6e838`), `ruvnet/dream-machine` (HEAD `8ce3857`), and `ruvnet/metaharness` (HEAD `5453c8c`), plus `docs/adr/` on `origin/main`. `ruvnet/ruflo` was **not** available to this pass — see NB-8.

This is a strong ADR set. The evidence discipline is genuinely better than most: every arXiv ID in the eleven ADRs matches `01-evidence-review.md` exactly, and the LatentMesh and autogenous numbers are exact against source (see "Verified clean" at the end). The blocking items below are all citation/scoping errors of the same family the set was written to prevent, which is why they need fixing before Phase 3 scales up on top of them.

---

## 🔴 Blocking

### B-1 — ADR-315 scopes work against an autogenous gap that is already closed upstream

ADR-315 Context, Decision §2, and its second Negative consequence all rest on autogenous ADR-401 having an open `Promote = Better ∧ Safe ∧ Authorized ∧ Reversible` predicate:

> "this program's contribution to closing that gap is real, unstarted work, not an integration against a finished mechanism"

Against source, autogenous ADR-401's **Update 1 §3** says:

> **"Converge the promotion invariant to one predicate. — DONE.** `mesh-evolve.ts` now exports `promoteAuthorized(candidate, champion, { authorized, reversible }) → PromotionDecision`, the single gate `Promote = Better ∧ Safe ∧ Authorized ∧ Reversible`, and `evolveMesh` routes every promotion through it… Each conjunct is independently blocking… proven by `test/promote-authorized.test.ts`… no promotion path can now skip a conjunct."

The error is understandable — ADR-401 is internally inconsistent: its capability-table row 5 still reads "not yet one checked predicate," which is the sentence this PR quotes, while its own Decision section supersedes that row. But the consequence is that a work package is scoped and budgeted against work upstream already shipped. ADR-305's Reconciliation section repeats the same claim.

**Fix**: re-read ADR-401 Update 1, then re-scope ADR-315 Decision §2. The genuine residual work looks like the `ruvector`-side integration and the capability-expansion-vs-behavioral-promotion distinction — not closing the predicate.

### B-2 — `metaharness ADR-251` does not exist

ADR-306 cites it in four places (Related line, Context §3, the canonical-implementation table, Affected Repos) as a real prior instance. In the `ruvnet/metaharness` clone there are **230 ADR files topping out at `ADR-250-sota-proof-ladder.md`** — no ADR-251. The Nightly Dream Cycle material lives in `docs/dream-cycle/` (`2026-08-13-gist.md`, `2026-08-14-gist.md`, `LEDGER.md`), not as an ADR.

This PR contradicts itself on the point: `04-verification-addendum.md` §5 establishes that metaharness has "230 ADRs topping out at ADR-250," and §6 of the same file then cites ADR-251. The citation originates upstream (dream-machine ADR-0001's own "Prior instances" line says `ruvnet/metaharness` ADR-251), so this is inherited rather than invented — but the PR restates it as verified fact after having established the opposite.

**Fix**: cite `ruvnet/metaharness`'s `docs/dream-cycle/`, or keep the ADR-251 reference but mark it explicitly as upstream-asserted-and-unlocated, the way ADR-305 handles the LatentMesh verification note.

### B-3 — "ruvector ADR-150 (optionalDependencies policy)" is a misattribution, and it gates a work package

ADR-313 puts it in the Related line and body twice, and makes it a blocking gate ("this work package's live-serve acceptance criteria cannot be claimed complete while either the ADR-150 compliance bug or the ruvllm HTTP-307 bug remains open"). ADR-306 and `03-program-plan.md` repeat it.

- `ruvector`'s ADR-150 is `ADR-150-pi-brain-ruvltra-tailscale.md` — "π Brain + RuvLtra via Tailscale — Semantic Embedding Upgrade". Unrelated.
- `metaharness`'s ADR-150 is `ADR-150-tailscale-local-frontier-concurrent-benchmarks.md`. Also unrelated.
- `METAHARNESS-README.md` in this repo lists it as "**ADR-150**: MetaHarness Integration Surfaces (**upstream**)" — an upstream document neither clone contains.

So the number is confidently repo-qualified to the one repo where it definitely means something else. ADR-305 Decision §4 makes exactly this repo-qualification a blocking WP1 requirement, citing the ADR-103 precedent, which makes this the set's own stated failure mode reproduced inside it.

**Fix**: locate the real owning document, or drop the number and describe the bug directly ("`METAHARNESS-README.md`'s documented `optionalDependencies` policy"). The underlying bug is real and well-described either way.

---

## 🟡 Non-blocking

**NB-1 — ADR-306 attributes a quote to ADR-0001 that lives in the README.** "Freeze the model. Evolve the harness. Evaluation is not promotion — the machine never merges; a human does." is verbatim in dream-machine's `README.md` (L19–20) and `packages/cli/README.md`, not in ADR-0001. ADR-0001 §2.4's own wording is *"**Evaluation is not promotion.** The session never merges, never self-promotes flywheel state…"* plus *"Promotion is a human act."* The substance is faithful and the quote is real — but ADR-306 Decision §5 says "ADR-0001's constitutional principle is adopted **verbatim**", which won't survive a reader checking ADR-0001. Attribute to the repo, or quote §2.4's actual sentences.

**NB-2 — `ruvector ADR-227` does not exist.** ADR-306's proof-gated-promotion gate cites it. No ADR-227 file exists anywhere in the repo — numbers 212–244 are entirely absent from `docs/adr/`. Note the already-merged ADR-304 makes the identical citation, so this is a pre-existing repo-wide dangling reference rather than something introduced here. The real proof-gate ADRs on `main` are ADR-047 (`proof-gated-mutation-protocol`) and ADR-194 (`proof-gated-writes`) — which ADR-307 already cites correctly.

**NB-3 — in-repo citations land on the duplicate numbers the program itself catalogues.** ADR-305 §4 mandates repo-qualification for cross-repo references, but says nothing about `ruvector`'s own duplicates, and the set cites straight into them: **ADR-134** (2 files: `witness-schema-log-format` *and* `ruvector-claude-code-deep-integration`) is cited ~6× as "the ADR-134 schema"; **ADR-194** (3 files) and **ADR-252** (3 files) likewise. Cheap fix: give the filename or title on first use, e.g. "ruvector ADR-134 (`witness-schema-log-format`)".

**NB-4 — ADR-312's declared dependents don't reciprocate.** ADR-312 states ADR-306, ADR-307, ADR-309, ADR-315 "all depend on this ADR." Only ADR-315 lists it in Dependencies. ADR-306 downgrades it to "shares its verification stack"; ADR-307 and ADR-309 omit it from both Related and Dependencies while relying on it in their Security-Gates sections.

**NB-5 — ADR-308 ↔ ADR-306 asymmetry.** ADR-308 names ADR-306 a "downstream consumer" of its drift signal and positions itself before Dream Machine evaluation; ADR-306 never mentions ADR-308 in Related or Dependencies. Worth stating explicitly that ADR-306 is built before ADR-308 (per WP2 → WP9 → WP10) but consumes it at runtime — otherwise the two Dependencies sections read as contradictory.

**NB-6 — the seventh invariant isn't recorded where the invariants live.** ADR-305 Decision §3 adopts ruflo ADR-322B's separation-of-powers as a seventh governing invariant "binding on ADR-306, ADR-313, ADR-315", but `03-program-plan.md` in this same PR still lists six under "Governing invariants (carried from the brief, **unchanged**)". ADR-313 states it explicitly ✅; ADR-306 states the substance without naming it as the invariant; ADR-315 never mentions it despite being named as bound by it.

**NB-7 — internal cross-reference error in ADR-305.** The Reconciliation section says the seventh invariant is adopted "(see Decision **§2** below)". It's adopted in Decision **§3**.

**NB-8 — the ruflo citations are unverified, and ADR-312 doesn't say so.** ADR-312 asserts ruflo ADR-322C's specifics as fact (RFC 8785 JCS, `Ed25519(domainPrefix || 0x00 || canonicalBytes)`, UUIDv7, the fail-closed rules), and ADR-310/312 lean on ADR-381's `α_k = α_total · 6/(π²k²)` schedule and the 0.6% figure. `ruvnet/ruflo` was not cloned for this verification pass and is not checked out locally, so — unlike LatentMesh, autogenous, dream-machine and metaharness — none of it is confirmed against source. ADR-305 §5(c) handles this honestly by deferring binding confirmation to WP1; ADR-312 carries no equivalent caveat while making 322C the entire anchoring contract. Add the same caveat.

**NB-9 — the plan promises 13 ADRs; 11 shipped.** `03-program-plan.md`'s ADR list items 0 (numbering hygiene) and 12 (MetaHarness dependency compliance) are written as "**ADR** — …" but have no ADR in this set. The PR description says hygiene is tracked separately, which is fine — but the plan document *in this PR* still promises them. Reword those two entries as work-package-only (WP0a/WP0b) so the plan and the ADR set agree.

**NB-10 — no ADR covers the RF Sensing Substrate bounded context.** WP14 / evidence item 8 (universal CSI vocabulary, UNVERIFIED) has no decision record. That is consistent with "sequence it last, fund separately," and it has a real upside — because no ADR touches CSI, nothing mis-cites the unverified claim, and `01-evidence-review.md` §8 handles it correctly as a first-party bet with the adjacent literature named. Still worth one line in ADR-305 or the plan saying the context is deliberately ADR-less until funded, so a reader mapping bounded contexts to ADRs doesn't read it as an omission. Related: autogenous **ADR-402** ("RuView/Cognitum Spaces Spatial Intelligence," Accepted) is directly on-point for that context and is named in the addendum but referenced by no ADR.

---

## Nits

- **ADR-305 at 227 lines** is the longest in the set. Flagging only to close the question: this is well within house norms (ADR-294 is 373 lines, ADR-304 is 303), so no action needed.
- **`## Affected Repos` and `## Security / Validation Gates` are new section headings** — zero prior uses across `docs/adr/`. Sensible additions for a cross-repo program; worth one line in the plan noting they're intentional so future ADRs follow rather than diverge. `Context` / `Decision` / `Consequences` / `Alternatives Considered` all match the dominant existing pattern, and the header block matches ADR-282/ADR-303 style.
- **dream-machine ADR-0002** (`dream-cycle-security-adversarial-entrypoint-liveness`) exists and is uncited. ADR-306 adopts that engine and defines security gates around it — worth a read during WP2.
- **`02-asset-map.md` §8's "last updated 2026-08-19"** for LatentMesh is wrong (clone HEAD confirms 2026-08-18). The addendum corrects it and deliberately doesn't edit in place, which is the right call — noting only that a reader hitting §8 first gets the wrong date.

---

## Verified clean

Stating these explicitly since they were the review's main targets:

- **Citations**: all nine arXiv IDs across the eleven ADRs match `01-evidence-review.md` exactly; no fabricated IDs. Causal audit is `arXiv:2607.26773` dated **2026-07-29**, with the brief's "Aug 5" explicitly retracted in ADR-310 ✅. LATTE is stated as not a real paper — permanently, in ADR-311's Decision §4 and with "describe this as implementing LATTE" listed as an explicitly rejected alternative ✅. TARL is spelled out as "TARL (Transaction-Aware Reliable Ledgers)" on first use in ADR-307, with its overloaded-acronym collisions cited ✅.
- **LatentMesh, exact against clone**: ADR-009 dated 2026-08-18, status Proposed; *"The statistical primitive (ADR-003) and admission gate (ADR-008) are implemented; the closed loop across live components is not wired"* verbatim; the seven-stage loop diagram matches stage for stage; LOC 272 / 454 / 534 / 147 = 1,407 exact; **no transport crate in the workspace** confirmed.
- **autogenous, exact against clone**: ADR sequence 391–403; ADR-401 "Accepted (framing + capability map) · Partial (implementation)", 2026-08-16; ADR-400 "Accepted — Implemented (first flywheel turn measured)"; 10-capability map; p50 0.34 ms recovery at 30% mesh loss; 100% vs 66.7% best-single; README research-prototype badge and "hypotheses until benchmarked"; `witness` 302 LOC; `envelope` 1,009 LOC; `radio-moe/src` 6,171 LOC. The "metaharness ADR-322" misattribution inside autogenous ADR-400/401 is real and correctly diagnosed.
- **dream-machine**: ADR-0001 Accepted, engine v0.1.0 shipped, 85 tests, dated 2026-08-13, composing the exact package versions cited.
- **Numbering**: 305–315 are collision-free against every ADR file in the repo, including the secondary ADR directories (`crates/rvf/docs/adr`, `crates/ruvector-mincut/docs/adr`, `docs/adr/coherence-engine`, etc., which use separate `ADR-CE-*`/`ADR-DB-*`/low-number namespaces). True max on `main` is ADR-304 ✅. Status: Proposed on all eleven ✅.
- **Diff**: docs-only, 15 files. The `win32-x64-msvc/ruvector.node` artifact the PR description warns about no longer appears in `origin/main...feat/pir-adrs`.
- **Decision soundness** (apart from the items above): ADR-306 does preserve "the machine never merges; a human does" as non-negotiable and explicitly rules out an unattended auto-merge path ✅. ADR-312 does anchor on ruflo ADR-322C rather than a crate merge, with the design-error rationale stated and the merge alternative explicitly rejected — well-supported by the crate comparison (302 LOC std/Ed25519/JSON vs 4,405 LOC `no_std`/keyed-MAC/64-byte ring), verified for the autogenous side ✅. ADR-313 enforces frozen weights structurally via a CI check on training-API imports reachable from the promotion pipeline, gates on WP0b, and honestly discloses that a static import check won't catch a dynamically-loaded fine-tuning path ✅. ADR-315 does carry the research-prototype caveat and the API-stability buffer ✅ (its problem is B-1, not the badge handling).
- **WP numbering**: every WP reference in the ADRs (WP0b, WP1, WP2, WP5, WP7, WP10, WP12) matches `03-program-plan.md`'s table.

Recommend clearing B-1 through B-3 and NB-6/NB-9 (both cheap edits to `03-program-plan.md`) before merge; the rest can land as follow-ups.

ruvnet:
## Review addendum — NB-11: ADR-313's WP0b gate names a bug that is already fixed

Follow-up to my review above. A second pass on `ruvllm`'s download path turned up one more stale-premise item. **Non-blocking wording fix** — the gate itself survives, only the named bug is wrong. Verdict is unchanged (CHANGES-REQUIRED, B-1 through B-3).

### The HTTP-307 redirect bug is fixed on `main`

`946275a61` — *"fix(ruvllm-cli): follow HF 307 redirect on aux-file download (#590)"*, 2026-06-18, confirmed an ancestor of `origin/main`. It adds a redirect-following `curl -L --fail` fallback in `crates/ruvllm-cli/src/commands/download.rs` for exactly the `tokenizer_config.json` failure ADR-259 describes.

ADR-313 currently treats it as open in five places — Context bullet 2, Decision §4, the Negative consequence, the **WP0b blocking gate** in Security/Validation Gates, and the second rejected Alternative. `03-program-plan.md` repeats it in WP0b's row, the WP0b issue body, and Top Risk §4.

**Root cause, and it is the same one as B-1 and B-2**: `ruvector`'s own ADR-259 was never updated after the fix landed. Its Status line still reads *"live-serve e2e blocked by a ruvllm download bug"* and its "Honest gap" paragraph still says *"the file is served via an HTTP 307 redirect that ruvllm does not follow… Recommend fixing the redirect-follow in `ruvllm download`."* The asset map read that stale text in good faith and the ADRs inherited it. That is now three citation errors in this set (B-1 autogenous ADR-401's stale capability row, B-2 dream-machine's stale ADR-251 reference, and this one) with one shared cause: **a status line its own repo never updated after the fix shipped.**

### The real remaining blocker is the GGUF download path

Confirmed in source at `crates/ruvllm-cli/src/commands/download.rs:193`, `get_files_to_download()`:

```rust
if model_id.contains("GGUF") || quant != QuantPreset::None {
    // Look for GGUF files
    files.push(format!("*{}", quant.gguf_suffix()));
```

Two distinct defects, both real:

1. **Unexpanded glob.** `*Q4_K_M.gguf` is pushed as a literal filename and never expanded against the HF file listing, so the resolve URL 404s. Commit `946275a61`'s own message anticipates this — *"needs HF file-listing + registry resolution and is out of scope for this redirect fix"* — i.e. the fix deliberately deferred it.
2. **Alias points at a safetensors repo.** In `crates/ruvllm-cli/src/models.rs:65`, alias `phi` → `microsoft/Phi-4-mini-instruct`, which is safetensors-only (contrast the Qwen entries, whose `hf_id`s end in `-GGUF`). Because the branch condition is `contains("GGUF") || quant != QuantPreset::None` and `phi`'s `recommended_quant` is `Q4_K_M`, `ruvllm download phi` takes the GGUF branch against a repo that has no GGUF files at all.

### Suggested fix

WP0b keeps its blocking status on WP9 — live-serve e2e is still blocked — but the scope shrinks by one already-done item and the gate should name the right bug. Concretely: in ADR-313 and `03-program-plan.md`, replace "the known HTTP-307 redirect bug in `ruvllm`'s model-download path" with the GGUF glob-expansion + alias-resolution bug in `get_files_to_download()`, and note that the 307 issue was closed by `946275a61` (#590). Worth folding a correction to `ruvector`'s own ADR-259 status line into WP0b as well, so the next reader doesn't repeat this.

### Systemic recommendation

Given three independent instances, WP1's verification pass should check every "known bug" / "gap" / "not yet implemented" claim this program inherits against **fix history** (`git log` on the named path, upstream release notes), not only against the prose of the ADR that asserts it. All three errors would have been caught by that one extra step, and all three cost real work-package scope.

ruvnet:
## Fixes pushed — commit `313543ace`

Addressed all three blocking findings, the four cheap "also fix" items, and folded in asset-scout's ruflo verification (which arrived mid-fix, so the ADR-322/381 items reflect confirmed source facts rather than the caveat originally requested). Mapping:

**Blocking**
- **B-1** (ADR-401 predicate already closed): ADR-315 and ADR-305 corrected — `promoteAuthorized` is DONE per ADR-401 Update 1 §3 (`mesh-evolve.ts`, `test/promote-authorized.test.ts`), with a note that the capability-table row 5 is stale relative to ADR-401's own Decision section. ADR-315 now *adopts* the predicate; it no longer scopes work to close it.
- **B-2** (metaharness "ADR-251" doesn't exist): all four ADR-306 citations now point at `docs/dream-cycle/`, with the bad citation's provenance stated explicitly (inherited from dream-machine ADR-0001's "Prior instances" line) rather than restated as fact.
- **B-3** (ADR-150 misattribution): every reference (ADR-306, ADR-313, `03-program-plan.md`) now cites `METAHARNESS-README.md`'s documented invariant directly, with the upstream "ADR-150" attribution flagged unverified — neither `ruvector`'s nor `metaharness`'s own ADR-150 is the right document.

**Also fixed**
- ADR-313's WP0b gate: HTTP-307 bug already fixed on `main` (commit `946275a61`, PR #590, 2026-06-18) — removed as a blocker in all 5 places in ADR-313 and all 3 in `03-program-plan.md`. Real remaining blocker: GGUF glob/alias mismatch in `ruvllm-cli`'s `get_files_to_download()` (`download.rs:193`, `models.rs:65`).
- NB-6: 7th governing invariant (ruflo ADR-322B separation-of-powers) now recorded in `03-program-plan.md`'s invariants list.
- NB-9: ADR-list items 0 and 12 reworded as WP0a/WP0b work-package-only entries (no ADR) — plan and shipped set now both say 11.
- NB-1: "the machine never merges; a human does" now attributed to dream-machine's README (verbatim source) with ADR-0001 §2.4's differing phrasing noted, in ADR-306.
- NB-8: asset-scout cloned `ruvnet/ruflo` (HEAD `fa13ee4`) before I pushed, so I incorporated the verified result instead of a caveat — 322/322A/322B/322C confirmed verbatim (three signing domains, evidence-grading vocabulary, the 322B quote now cited as source-verified). ADR-381 corrected instead: it's **Proposed** not Accepted, and its own contribution is stream-identity + evidence-epoch-reset governance — the α_k formula and 0.6% figure belong to PR #2956, which ADR-381 governs. Every family-wise bound is now stated **per-epoch**, matching ADR-381's own text (line 33).
- Bonus: 322C line 105 settles the ADR-103 cross-repo ambiguity as `ruflo`'s — noted in the addendum.
- NB-7 (the "§2 vs §3" internal cross-ref error in ADR-305) fixed while in the file — cheap, same commit.

**Systemic fix**: ADR-305 now states a standing fix-history verification rule (Decision §6) — an inherited "known bug/gap/not-yet-implemented" claim must be checked against the named path's actual fix history before being repeated in a PIR ADR. All three blocking findings and the HTTP-307 item shared this root cause.

Full correction trail: `docs/research/perpetual-intelligence-runtime/04-verification-addendum.md` §8 (new).

Not addressed in this push (deferred as follow-ups per the review's own recommendation): NB-2 through NB-5, NB-10, and the nits.

ruvnet:
## Addendum self-consistency fix — commit \`55a32c76e\`

asset-scout spot-checked \`313543ace\` and found the ADR files all correct, but flagged that \`04-verification-addendum.md\` §5 still carried the pre-correction ADR-381 phrasing (said ADR-381 "defines" the scheme, omitted Proposed status, dropped the per-epoch qualifier) — contradicting §8d, which had it right. Same document, two different claims depending on which section a reader hit first.

Fixed: §5's paragraph now attributes the mechanism to PR #2956 and ADR-381's role to governance over it (per-epoch bound stated), with a pointer to §8d. Also expanded §8b with the exact upstream source of the "metaharness ADR-251" citation (dream-machine ADR-0001 line 9) so a future reader doesn't "fix" ADR-306 back to the wrong citation.

No ADR files needed changes — only the addendum.

ruvnet:
## §6 back-propagation fix — commit \`7d0daa63e\`

asset-scout's second spot-check found the same class of issue from the §5 fix repeated in §6: two "metaharness ADR-251" references (dream-machine ADR-0001's "prior instances" claim, and the "four places Dream Machine exists" list) weren't updated when §8b resolved that citation. The second was a real error — it listed a nonexistent document as one of four real locations; the actual fourth location is \`ruvnet/metaharness\`'s \`docs/dream-cycle/\`.

Both now carry the correction inline with a pointer to §8b. Also grepped the rest of the file for the same pattern against the other §8 corrections (ADR-322/metaharness-ADR-322 misattribution, the LatentMesh date, the autogenous crate count) — no other unqualified stale twins found.

ADR files unaffected — this is addendum-only, as with the previous fix.

ruvnet (commented):
## Re-review — remediation commit `313543ace`

**Verdict: one remaining blocker.** B-1 and B-2 are fully cleared, NB-1/6/7/9/11 are cleared, and the ruflo verification work added here is excellent and checks out verbatim against source. B-3 is **half-cleared**: the wrong-repo attribution was removed but replaced with a different wrong-repo claim, and the document it says doesn't exist is sitting in the ruflo clone this same commit cites. That is a ~15-minute fix, not a rework.

I re-verified every claim in the diff against the clones rather than taking the fix note's word for it, including the newly available `ruvnet/ruflo` (HEAD `fa13ee4`, 2026-08-15, 177 ADRs in `v3/docs/adr/` — both confirmed).

---

## 🔴 B-3 residual — `ruflo` ADR-150 is the missing document, and it's in the clone

ADR-313 now says the `optionalDependencies` policy attribution *"does not resolve to a real document in either repo,"* tags it **unverified** in its Related line, and characterizes the source as *"an upstream **metaharness** ADR-150."* The same wording is now in ADR-306 (L61–62), `03-program-plan.md` (L90–93, and the WP0b issue body at L213, which actively instructs filers not to cite it), and addendum §8c.

It resolves cleanly:

**`ruflo/v3/docs/adr/ADR-150-metaharness-integration-surfaces.md`** — *"ADR-150 — MetaHarness Integration Surfaces in `npx ruflo`"*, **Status: Implemented**, dated 2026-06-16 (rev. 06-17). Its normative rule is the exact policy `METAHARNESS-README.md` claims compliance with:

> Every integration in this ADR, and every future ADR that extends it, must satisfy:
> 1. **Removable**: … must still produce a working CLI.
> 2. **Optional in `package.json`**: `@metaharness/*` packages MUST appear in `optionalDependencies` or `peerDependencies` (optional), **never in `dependencies`**.
> 3. **Graceful degradation**: … catch `MODULE_NOT_FOUND` and fall back …
> 4. **CI coverage of the absent path**: at least one CI job must run `--ignore-optional` … "the only structural defense against accidentally promoting an optional dep to required."

The evidence chain is decisive:

1. `METAHARNESS-README.md`'s References list two upstream ADRs together — L183 *"**ADR-150**: MetaHarness Integration Surfaces (upstream)"* and L184 *"**ADR-103**: Witness Chain (upstream)."* The ADR-150 title matches ruflo's file title exactly on its distinctive phrase.
2. This ADR set **already established that the other one is ruflo's** — addendum §5 on ADR-103, now independently corroborated by ADR-322C line 105 (quoted in this very commit). Same README, same "upstream," same repo.
3. `METAHARNESS-README.md`'s own section headings mirror ruflo ADR-150's four rules one-for-one: L28 *"graceful degradation if MetaHarness missing"*, L73 *"ADR-150 Compliance (MetaHarness **Removable**)"*.
4. ruflo ADR-322 line 17 shows the policy in force upstream: *"`@metaharness/darwin@^0.8.0` and `@metaharness/flywheel@^0.1.7` are **optional dependencies** guarded by `scripts/check-metaharness-pins.mjs`. Removing them leaves the local receipt, transaction, and proposer path operational."* — and those are the exact versions ADR-306's version-drift note flags `ruvector` as pinning.

Two things are wrong as written, both small: the upstream is **ruflo**, not metaharness; and the policy is **verified**, not unverified. Note this is now the fourth instance of the same pattern the set exists to catch — assuming the wrong repo owns an ADR number — and the correction is exactly parallel to the "ruflo ADR-322, never metaharness ADR-322" fix the set already makes twice.

It also matters practically: WP0b's gate currently points at a policy marked unverified, when the real policy has a four-part, CI-testable definition of done. Rule 4 in particular (a CI job on the absent path) is the acceptance criterion WP0b should adopt, and it's stronger than the current "npm install succeeds without the packages."

**Fix**: replace the "does not resolve / unverified" wording in the five places above with `ruflo` ADR-150 (Implemented), and point WP0b's acceptance criteria at its four rules.

---

## ✅ Cleared

**B-1 — fully cleared.** ADR-315's Context now carries an explicit "Correction from PR review" naming the stale capability-table row, cites Update 1 §3 as DONE with the `promoteAuthorized` signature and `test/promote-authorized.test.ts`, and — the part that matters — Decision §2 is genuinely rescoped rather than reworded: it now adopts the predicate and scopes only (a) the `ruvector`-side wiring into the `authorized` parameter and (b) the distinctness rule. The Negative consequence flipped correctly from "real, unstarted work" to upstream-API-churn risk, and a matching Positive was added. ADR-305's Reconciliation and the Related line are consistent with it. This is the right fix, not a papered-over one.

**B-2 — fully cleared, and handled well.** ADR-306 no longer restates the citation as fact; it names `docs/dream-cycle/` (with the three actual filenames) as the real location, states the ADR-250 ceiling, and attributes the bad citation to dream-machine ADR-0001's own text *"noted here as its provenance rather than independently verified."* The canonical-implementation table, Negative consequence, and Affected Repos are all updated consistently. Addendum §6 is corrected by §8b rather than edited in place, consistent with this document set's stated convention.

**NB-11 — cleared in all eight places**, verified individually: ADR-313 Context, Decision §4, Negative, WP0b gate, Alternatives; plan WP0b table row, WP0b list item, Risk §4, and the WP0b issue title/body. The residual `HTTP-307` hits in the tree are all correction text or the deliberately-unedited `02-asset-map.md`.

**NB-1** — ADR-306 §5 now attributes the sentence to the README (L19–20 and `packages/cli/README.md`), quotes ADR-0001 §2.4's own wording alongside, and drops "verbatim." Accurate.
**NB-6** — plan now has seven invariants with #7 stated and sourced. **NB-7** — `§2` → `§3`, and Decision §3 does contain the adoption. **NB-9** — WP0a/WP0b reworded as work-package-only with an explicit note about the 13-vs-11 mismatch.

**New: ADR-305 Decision §6, the fix-history verification rule.** Well-judged addition — it generalizes the root cause instead of just patching instances, names the three findings and the HTTP-307 case as the evidence for it, and binds the rest of the program. This is the most valuable thing in the commit.

---

## Spot-check of the new ruflo material — all verbatim

Every ruflo specific added by this commit checks out against the clone:

| Claim | Source | ✓ |
|---|---|---|
| ruflo HEAD `fa13ee4`, 2026-08-15; 177 ADRs | clone | ✓ |
| ADR-322 "Accepted — phases 0–2 implemented", 2026-07-28 | L3 | ✓ |
| 322A/322B/322C "Accepted — implemented" + rollout flags `RUFLO_FLYWHEEL_{TRANSACTION,DARWIN,RECEIPT}_V1` | headers | ✓ |
| 322B L10 *"A proposer produces untrusted candidates only. It cannot issue promotion decisions or mutate active policy."* | L10, exact | ✓ |
| ADR-322 L15 *"Darwin adapters remain candidate generators and never gain promotion authority."* | L15, exact | ✓ |
| `Ed25519(domainPrefix ‖ 0x00 ‖ canonicalBytes)`; `candidateId = SHA-256(JCS(candidate policy))`; `receiptId = SHA-256(JCS(unsigned receipt payload))` | L17, L27, L29 | ✓ |
| Evidence grades `recomputed` / `signature-verified` / `trusted-assertion` | L68, exact | ✓ |
| 322C L105 settles the ADR-103 ambiguity as ruflo's | L105, exact | ✓ |
| **ADR-381 is Proposed, 2026-08-10** (not Accepted) | L3–4 | ✓ |
| **PR #2956 owns the mechanism; ADR-381 governs it** | ADR-381's own Related line: *"PR #2956 … the mechanism this ADR governs"* | ✓ |
| `α_k = α_total·6/(π²k²)`, `Σα_k = α_total = 5%`; 0.6% over 1,000 nulls | L10 | ✓ |
| Stream identity = one per project root | L24 | ✓ |
| `resetSequentialEvidence`, `confirm: true` + human reason, expires outstanding receipts | L28–32 | ✓ |
| **Per-epoch bound** *"family-wise false-promotion probability is ≤ α_total per epoch"* | L33, and L64 | ✓ |

The ADR-381 status/attribution correction and the per-epoch qualifier are substantive catches that materially change what ADR-310/312 may claim — neither is obtainable without reading the source. Propagating the per-epoch bound into ADR-312's Security Gates, including the warning that a day-30-vs-day-1 comparison spanning an epoch reset must not claim a global bound, is a genuinely sharp addition.

---

## Nits (new)

- **"Three distinct signing domains (bootstrap, receipt, ledger-head)"** (ADR-312 Context, addendum §8d) is off by one. 322C's "Signatures and keys" section defines **two** Ed25519 signing domains: `ruflo/flywheel-receipt/v1` (L96) and `ruflo/flywheel-ledger-head/v1` (L102). `"ruflo/bootstrap/v1"` (L82) is a domain-separated prefix for the paired bootstrap's **deterministic seed**, in the statistics section — not a signing domain. Suggest: "two Ed25519 signing domains plus a third domain-separated prefix for the bootstrap seed."
- ADR-315 still doesn't reference invariant 7 despite ADR-305 and the plan both naming it as bound by it. Carried over from NB-6; one sentence.
- ADR-305 is now 255 lines (was 227). Still within house norms (ADR-294 is 373), noting only that it's growing.

## Safe to defer — confirmed unchanged and non-scoping

NB-2 (`ruvector ADR-227` dangling — pre-existing, merged ADR-304 has the same reference), NB-3 (in-repo duplicate-number citations — WP0a's job), NB-4 and NB-5 (ADR-312/307/309 and ADR-308/306 cross-reference asymmetries), NB-10 (no ADR for the RF-sensing context), and the earlier nits. All are documentation-consistency items with no effect on scope, gating, or any factual claim. None regressed in this commit.

---

Clear B-3's residual and this is merge-ready. Everything else in the remediation is well-evidenced work — the ruflo pass in particular found two real errors nobody had flagged.

ruvnet (commented):
## Re-review — branch tip `7d0daa63e` (commits `55a32c76e` + `7d0daa63e`)

**Verdict unchanged: one remaining blocker — the B-3 residual (ruflo ADR-150).** Both new commits are clean, correct, and introduce no new errors. But they are addendum-only (`git diff --name-only 313543ace..7d0daa63e -- docs/adr` returns nothing), and B-3 lives in ADR-306, ADR-313, and `03-program-plan.md` — none of which changed. It is still the single thing standing between this PR and merge-ready.

### ✅ Both new commits verified

**`55a32c76e` — §5's stale ADR-381 paragraph.** Correctly reconciled with §8d: PR #2956 implements the mechanism, ADR-381 (**Proposed**, 2026-08-10) governs stream identity and budget-exhaustion recovery over it, bound stated **per epoch**. The closing line is the right way to handle it in a document that deliberately doesn't edit in place — *"that version is superseded by this one and by §8d, not by both disagreeing."* Good catch by asset-scout; §5 and §8d genuinely did contradict each other after `313543ace`.

**`7d0daa63e` — §6's two ADR-251 twins.** Both corrected: the "subsumes" sentence now carries the inline qualifier, and the "Net effect" list now reads `ruvnet/metaharness`'s `docs/dream-cycle/` with a pointer to §8b. §8b's new provenance quote is **verbatim against source** — dream-machine `ADR-0001` line 9 reads exactly *"- **Prior instances**: `ruvnet/metaharness` ADR-251 (MetaHarness Nightly Dream Cycle); the Ruflo Nightly Dream Cycle v3 (gist `ruvnet/889ffa92dab49d508e70b123c940e1b9`)"*. The added "**Future readers**: do not 'fix' it back to 'metaharness ADR-251' on the assumption that an ADR citation inside another Accepted ADR must already be resolved" is a good guard against exactly the regression this document set is prone to.

**Independent sweep** (not relying on the commit's own closing sweep): across all 15 PR files I find **zero** remaining unqualified stale claims — no ADR-381-as-mechanism-owner phrasing, no unqualified "metaharness ADR-251", no "ruvector ADR-150" outside explicit correction text. Every ADR-251 mention in the addendum (8 occurrences) is either a correction heading, a quoted provenance, or carries an inline "does not resolve" qualifier. The reconciliation is complete.

### 🔴 Still outstanding — B-3 residual, unchanged in all five places

- `ADR-313` L6 (Related, tagged **unverified**) and L39 (*"in either repo"*)
- `ADR-306` L62 (*"an upstream metaharness ADR-150 not present in either repo"*)
- `03-program-plan.md` L90–93 and the WP0b issue body at L213
- `04-verification-addendum.md` §8c

The document is `ruflo/v3/docs/adr/ADR-150-metaharness-integration-surfaces.md` — *"MetaHarness Integration Surfaces in `npx ruflo`"*, **Status: Implemented**, 2026-06-16 — whose rule 2 is the policy verbatim: *"`@metaharness/*` packages MUST appear in `optionalDependencies` or `peerDependencies` (optional), never in `dependencies`."* Full evidence chain in my previous comment; the short version is that `METAHARNESS-README.md` lists ADR-150 and ADR-103 side by side as "(upstream)", and this PR already proved the ADR-103 half is ruflo's (addendum §5, corroborated by ADR-322C L105 quoted in `313543ace`).

Worth noting the irony: `7d0daa63e` adds a warning against assuming an ADR citation inside another Accepted ADR is already resolved — which is the correct lesson — while the ADR-150 text still asserts an unresolvable citation that does resolve, in a repo now cloned and cited elsewhere in the same branch.

### Nits (unchanged / one new)

- **"Three distinct signing domains (bootstrap, receipt, ledger-head)"** — still in ADR-312 Context and addendum §8d. 322C's "Signatures and keys" section defines two Ed25519 signing domains (`ruflo/flywheel-receipt/v1` L96, `ruflo/flywheel-ledger-head/v1` L102); `"ruflo/bootstrap/v1"` (L82) is a seed-derivation prefix in the statistics section.
- **New**: §8d reads *"**ADR-381 needed two corrections.**"* then lists (a), (b), **and (c)**. Three. Carried in from `313543ace`; one word.
- ADR-315 still doesn't reference invariant 7 despite ADR-305 and the plan naming it as bound.

### Deferred — reconfirmed at tip

NB-2 (ADR-227 dangling, pre-existing), NB-3 (in-repo duplicate-number citations, WP0a's job), NB-4/NB-5 (cross-reference asymmetries), NB-10 (no ADR for the RF-sensing context). None regressed; none affect scope or gating.

---

Fix the five ADR-150 references (and ideally the two nits above in the same pass) and this is merge-ready. Nothing else is holding it.

ruvnet:
## B-3 residual + nits fixed — commit \`9f11cae54\`

Resolved the ADR-150 attribution to **ruflo's own ADR-150** (\`v3/docs/adr/ADR-150-metaharness-integration-surfaces.md\`, "MetaHarness Integration Surfaces in \`npx ruflo\`," Status: Implemented, 2026-06-16) — independently confirmed against the clone before editing. Its rule 2 is verbatim the \`optionalDependencies\` policy; rule 4 (a CI job on \`--ignore-optional\`) is now cited as the stronger acceptance criterion in place of a plain \`npm install\` check.

Fixed all five flagged locations: ADR-313's Related line + Context, ADR-306's version-drift note, \`03-program-plan.md\`'s WP0b description **and** its GitHub issue-breakdown body, and addendum §8c. Noted this as the fourth instance of the wrong-repo-ADR-number pattern (ADR-103, "metaharness ADR-322", "metaharness ADR-251", now this).

Also fixed both nits from this re-review:
- Signing-domain count: ADR-322C defines **two** Ed25519 signing domains (\`flywheel-receipt\`, \`flywheel-ledger-head\`) plus a third non-signing domain-separated prefix for the bootstrap seed — corrected in ADR-312's Context and addendum §8d (which also had a stray "two corrections" label over an (a)/(b)/(c) list — now "three").
- ADR-315 now cites governing invariant 7 (ruflo ADR-322B separation-of-powers, adopted in ADR-305) in its Related line, Decision, and Security Gates, closing the one-way reference.

Ready for final verify.

ruvnet (commented):
## Final verification — tip `9f11cae54`

# ✅ MERGE-READY

All blocking findings are cleared. I re-verified every citation in this commit against the clones rather than accepting the fix note, and **every quote, status line, date, and line number is exact**. One cosmetic heading nit remains (below); it is not worth holding the merge for.

### B-3 residual — cleared, and cleared correctly

All five locations now cite `ruflo` ADR-150 as the source, verified verbatim against `ruflo/v3/docs/adr/ADR-150-metaharness-integration-surfaces.md`:

| Claim in the fix | Source | ✓ |
|---|---|---|
| Title *"ADR-150 — MetaHarness Integration Surfaces in `npx ruflo`"* | L1 | ✓ |
| Status **Implemented**, 2026-06-16 (rev. 06-17) | L3–4 | ✓ |
| Rule 2 *"`@metaharness/*` packages MUST appear in `optionalDependencies` or `peerDependencies` (optional), never in `dependencies`."* | L39, verbatim | ✓ |
| Rule 4 *"at least one CI job must run `--ignore-optional` … the only structural defense against accidentally promoting an optional dep to required"* | L41, verbatim | ✓ |
| Evidence-chain item 3 (README's "MetaHarness Removable" heading ↔ ADR-150 rule 1) | L38 | ✓ |
| Evidence-chain item 4 (ruflo ADR-322 shows the policy in force at `darwin@^0.8.0` / `flywheel@^0.1.7`, both `optionalDependencies`) | ADR-322 L17, verbatim | ✓ |

Two things beyond the minimum fix are worth calling out as good judgment. First, ADR-313 and the plan both adopt **rule 4's `--ignore-optional` CI job as WP0b's acceptance criterion**, replacing the weaker "npm install succeeds" — that is a materially better gate, and it came out of resolving the citation rather than being bolted on. Second, `ruvnet/ruflo` was added to ADR-313's Affected Repos, which the earlier drafts had missed even while depending on ruflo ADR-322B.

Addendum §8c is rewritten honestly: it keeps the original error in the heading, records that a first pass marked it unverified, then lays out the four-step evidence chain and notes this is the **fourth** instance of the "wrong repo owns this ADR number" pattern. That count is coherent with ADR-305 §6's "three blocking findings" — they enumerate different sets (wrong-repo-ADR instances vs. blocking review findings), so there's no contradiction.

### Other fixes in this commit — all verified

- **Signing-domain count corrected** in both ADR-312 and §8d: two Ed25519 signing domains (`ruflo/flywheel-receipt/v1` L96, `ruflo/flywheel-ledger-head/v1` L102) plus a third domain-separated prefix (`ruflo/bootstrap/v1` L82) that seeds the deterministic paired bootstrap rather than signing. All three line numbers check out, and §8d explicitly records the overcount rather than silently editing it.
- **ADR-315 invariant 7** now appears in the Related line, in Decision §1 (the proposer of a capability-expanding mutation cannot also issue its approval), and as its own Security Gate. This closes the last piece of NB-6.
- **ADR-306's version-drift note** now points at ruflo ADR-150 instead of the dangling reference.

### Full-branch sweep

Zero residual stale phrasing anywhere in the 15 changed files: no "not present in either repo", no "does not resolve to a real document", no "three distinct signing domains", no unqualified "ruvector ADR-150", no unqualified "metaharness ADR-251", no ADR-381-as-mechanism-owner. All eleven ADRs remain **Status: Proposed**. Diff is docs-only, 15 files, +2,701 lines, no binary artifacts.

### Nit (do not block on this)

`04-verification-addendum.md` L380 — §8d's **heading** still reads *"ADR-381 needed **two** corrections"* while its body (correctly fixed in this commit) now says *"needed **three** corrections"* and enumerates (a), (b), (c). One word, in a research addendum, with no factual claim affected. Sweep it whenever something else touches the file.

### Deferred, reconfirmed at tip

NB-2 (`ruvector ADR-227` dangling — pre-existing; merged ADR-304 shares the same reference), NB-3 (in-repo duplicate-number citations — WP0a's job), NB-4 / NB-5 (ADR-312↔307/309 and ADR-308↔306 cross-reference asymmetries), NB-10 (no ADR for the RF Sensing Substrate context — deliberate, the stretch item is unfunded). None regressed; none affect scope, gating, or any factual claim.

---

**MERGE-READY.** Over four remediation rounds this set went from three blocking citation errors to a fully source-verified cross-repo ADR program: LatentMesh, autogenous, dream-machine, metaharness, and ruflo all cloned and checked, with the corrections that mattered most — autogenous's already-closed promotion predicate, ADR-381's Proposed status and per-epoch bound, the already-fixed HTTP-307 bug, and now ruflo ADR-150 — each one changing what a work package is actually scoped to build. ADR-305 Decision §6's fix-history rule is the right artifact to carry out of this review.

Issue #844, opened 2026-08-20
[PIR][WP13] Implement closed-form KV-cache cross-model migration in ruvllm (arXiv:2608.03893)


ruvnet:
Epic: #837 · Plan: `docs/research/perpetual-intelligence-runtime/03-program-plan.md`

## Goal

Implement the closed-form linear KV-cache mapper for same-family model migration in **`crates/ruvllm`** (`kv_cache.rs`, `paged_attention.rs`, `serving/kv_cache_manager.rs`), with the nonlinear MLP fallback for degrading pairs, plus a routing gate that **predicts transfer quality before migrating** — so the system never migrates blind.

## Why this ships first

Grade-A evidence, cheapest to build, and it depends only on `ruvllm`. **No dependency on WP0–WP12** — it runs as a parallel track and can start immediately, landing while the main branch is still in Phase 0/1.

## Evidence

[arXiv:2608.03893](https://arxiv.org/abs/2608.03893) — "Cross-Model KV Cache Transfer in LLM Families: A Closed-Form Linear Mapping for Prefill Reuse," submitted 2026-08-04. **The strongest-evidence item in the whole program review**, exact match on every figure:

- The mapper runs **2.7–25x faster than re-prefill**.
- Six pairs tested across three model families.
- The linear mapper retains **73–98% of the receiver's standalone-prefill accuracy on four pairs, while two degrade sharply.**
- A nonlinear MLP variant recovers **up to +37pp HellaSwag accuracy** on the failing pairs.

The 2-of-6 failure rate is the point: it is exactly why the quality-prediction gate is a requirement rather than a nice-to-have.

## Acceptance criteria

- [ ] **Reproduces the paper's reported 2.7–25x speedup over re-prefill on an internal same-family model pair.** Report our own measured numbers, never the paper's restated as ours.
- [ ] **The routing gate correctly refuses or downgrades for a known-degrading pair.**
- [ ] Closed-form linear mapper implemented for same-family migration.
- [ ] Nonlinear MLP fallback implemented for pairs the linear mapper degrades on.
- [ ] **No code path migrates without consulting the gate.** Enforced structurally, with a bypass-attempt test.
- [ ] Per-pair accuracy retention measured and recorded; degrading pairs are identified **automatically** rather than hand-maintained in a list that will drift.
- [ ] Degradation is observable at runtime — a migration whose realized quality falls below prediction is logged and counts against the gate's calibration.
[...]

PR #854, opened 2026-08-20
feat(ruvllm): cross-model KV-cache migration — linear mapper + quality gate (PIR WP13, ADR-314)

ruvnet:
## Summary

First shippable slice of cross-model KV-cache migration in `crates/ruvllm`, implementing [arXiv:2608.03893](https://arxiv.org/abs/2608.03893) — "Cross-Model KV Cache Transfer in LLM Families: A Closed-Form Linear Mapping for Prefill Reuse" — per **ADR-314** (branch `feat/pir-adrs`) and the PIR program plan's WP13 fast-follow track.

Tracking issue: #844 (child of PIR epic #837).

## What's in this slice

New module `crates/ruvllm/src/kv_migration/`, feature-gated behind `kv-migration` (**default off** — zero impact on existing builds):

| File | Contents |
|---|---|
| `mapper.rs` | `KvMapper` trait + `LinearKvMapper`: closed-form ridge least-squares fit `W = (XᵀX + λI)⁻¹XᵀY` of per-layer `W_k`/`W_v` from paired calibration activations; applied via matmul. Gaussian elimination with partial pivoting — no new dependencies (uses the crate's existing `ndarray`). |
| `quality.rs` | `TransferQualityGate`: per-layer reconstruction cosine similarity + relative Frobenius error on a **held-out** validation set, aggregated with conservative worst-layer thresholds (default min cosine 0.92, max rel. error 0.25), producing `Migrate \| Reprefill`. |
| `mod.rs` | Public API `migrate_kv_cache(source_cache, mapper, gate) -> MigrationResult` where the gate can veto into `Reprefill` — **never migrate blind** (the paper found 2 of 6 model pairs degrade sharply). `MlpKvMapper` is a documented stub for the paper's nonlinear fallback (out of scope here; the trait already accommodates it). `LayerKvTensor::from_flat` adapts `TwoTierKvCache::get_all_kv()`'s flat stride layout. |
| `tests.rs` | Synthetic-geometry unit tests (see below). |

### Key design decisions

- **Head-count / head-dim mismatch**: handled by learning the projection over the *flattened* per-token KV vector (`W` is `[src_heads·src_head_dim, tgt_heads·tgt_head_dim]`), which subsumes the paper's block projection when head grids align. Assumptions (position-aligned calibration pairs, compatible RoPE conventions, 1:1 layer correspondence) are documented in `mapper.rs`.
- **Conservative gate**: decision is on the *worst* layer, not the mean — one badly-mapped layer corrupts downstream attention; a false veto only costs the re-prefill we'd have paid anyway. Non-finite scores are an automatic veto.
- **Ridge regularization** (default λ=1e-4) keeps the normal equations solvable when calibration tokens barely cover the source dimension.

## Test results

`cargo test -p ruvllm --features kv-migration kv_migration`:

```
running 7 tests
test kv_migration::tests::mlp_mapper_stub_returns_not_implemented ... ok
test kv_migration::tests::layer_count_mismatch_is_rejected ... ok
test kv_migration::tests::quality_gate_vetoes_mismatched_mapping ... ok
test kv_migration::tests::from_flat_roundtrip_matches_stride_layout ... ok
test kv_migration::tests::known_linear_transform_is_recovered_by_fit ... ok
test kv_migration::tests::identity_family_mapping_recovers_cache_exactly ... ok
test kv_migration::tests::dimension_mismatch_mapping_produces_correct_shapes ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 1588 filtered out
```

`cargo check -p ruvllm` (without the feature) passes — no default-build impact.

## Out of scope (follow-ups under WP13 / ADR-314)

- `MlpKvMapper` implementation (nonlinear fallback for the degrading pairs, +37pp HellaSwag in the paper)
- Wiring into `serving/kv_cache_manager.rs` / `paged_attention.rs` serving paths
- Reproducing the paper's 2.7–25× re-prefill speedup on real same-family model pairs (ADR-314 acceptance bar)

### DIFF
--- a/crates/ruvllm/Cargo.toml
+++ b/crates/ruvllm/Cargo.toml
@@ -134,6 +134,11 @@ wasm-simd = []
 # Quantization support (requires platform-specific SIMD)
 quantize = []
 
+# Cross-model KV-cache migration: closed-form linear mapper + transfer-quality
+# gate (arXiv:2608.03893, PIR WP13, ADR-314). Default-OFF so it cannot affect
+# existing builds.
+kv-migration = []
+
 # P5 Optimization: Optional routing metrics collection
 # Disable for production to avoid Instant::now() syscall overhead (~0.04-0.08µs)
 routing-metrics = []
--- a/crates/ruvllm/src/kv_migration/mapper.rs
+++ b/crates/ruvllm/src/kv_migration/mapper.rs
@@ -56,6 +56,25 @@ use super::LayerKvTensor;
 /// normal equations solvable when calibration tokens barely cover `src_dim`.
 pub const DEFAULT_RIDGE_LAMBDA: f32 = 1e-4;
 
+/// Maximum flattened per-token KV dimension (`num_kv_heads * head_dim`)
+/// accepted by [`LinearKvMapper::fit`], on both the source and target side.
+///
+/// The closed-form solve is O(n³) in the source dimension and allocates an
+/// `[n, n]` Gram matrix, so calibration input of size O(tokens · n) buys
+/// O(n³) work — an unbounded `n` is a denial-of-service hazard (`n = 65_536`
+/// would request a ~17 GB Gram matrix, and an allocation failure aborts the
+/// process). 8192 comfortably covers realistic flattened KV dimensions while
+/// keeping the solve tractable.
+pub const MAX_FIT_DIM: usize = 8192;
+
+/// Maximum number of calibration layers accepted by [`LinearKvMapper::fit`].
+/// Each layer costs two independent O(n³) solves (keys and values).
+pub const MAX_CALIBRATION_LAYERS: usize = 512;
+
+/// Maximum calibration tokens per layer accepted by [`LinearKvMapper::fit`].
+/// Bounds the O(tokens · n²) Gram-matrix accumulation.
+pub const MAX_CALIBRATION_TOKENS: usize = 1 << 20;
+
 /// Paired calibration activations for one layer: the same token sequence's
 /// K/V tensors as produced by the source model and by the target model.
 #[derive(Debug, Clone)]
@@ -135,11 +154,27 @@ impl LinearKvMapper {
     /// `calibration` holds one [`CalibrationLayerPair`] per layer, in layer
     /// order. All layers must share the same source and target dimensions.
     /// `ridge_lambda` must be positive; see [`DEFAULT_RIDGE_LAMBDA`].
+    ///
+    /// ## Cost
+    ///
+    /// Per layer, the closed-form solve costs O(tokens · n² + n³ + n² · m)
+    /// time and allocates an `[n, n]` Gram matrix, where `n` is the source
+    /// dimension and `m` the target dimension — the O(n³) term dominates for
+    /// realistic shapes. Inputs are bounded by [`MAX_FIT_DIM`],
+    /// [`MAX_CALIBRATION_LAYERS`], and [`MAX_CALIBRATION_TOKENS`]; anything
+    /// larger is rejected up front, before any allocation or solve.
     pub fn fit(calibration: &[CalibrationLayerPair], ridge_lambda: f32) -> Result<Self> {
         if calibration.is_empty() {
             return Err(RuvLLMError::KvCache(
                 "cannot fit LinearKvMapper from empty calibration set".to_string(),
             ));
+        }
+        if calibration.len() > MAX_CALIBRATION_LAYERS {
+            return Err(RuvLLMError::KvCache(format!(
+                "calibration set has {} layers, exceeding MAX_CALIBRATION_LAYERS = {}",
+                calibration.len(),
+                MAX_CALIBRATION_LAYERS
+            )));
         }
         if !(ridge_lambda > 0.0) || !ridge_lambda.is_finite() {
             return Err(RuvLLMError::KvCache(format!(
@@ -149,8 +184,20 @@ impl LinearKvMapper {
 
         let source_dim = calibration[0].source.dim();
         let target_dim = calibration[0].target.dim();
+        if source_dim == 0 || target_dim == 0 {
+            return Err(RuvLLMError::KvCache(format!(
+                "calibration dims must be non-zero, got {source_dim}->{target_dim}"
+            )));
+        }
+        if source_dim > MAX_FIT_DIM || target_dim > MAX_FIT_DIM {
+            return Err(RuvLLMError::KvCache(format!(
+                "calibration dims {source_dim}->{target_dim} exceed MAX_FIT_DIM = {MAX_FIT_DIM} \
+                 (the solve is O(n³) in the source dimension)"
+            )));
+        }
 
-        let mut layers = Vec::with_capacity(calibration.len());
+        // Validate every layer BEFORE the first O(n³) solve, so a malformed
+        // set is rejected without paying for any partial fit.
         for (idx, pair) in calibration.iter().enumerate() {
             if pair.source.dim() != source_dim || pair.target.dim() != target_dim {
                 return Err(RuvLLMError::KvCache(format!(
@@ -159,6 +206,17 @@ impl LinearKvMapper {
                     pair.target.dim()
                 )));
             }
+            if pair.source.tokens() > MAX_CALIBRATION_TOKENS {
+                return Err(RuvLLMError::KvCache(format!(
+                    "layer {idx} has {} calibration tokens, exceeding MAX_CALIBRATION_TOKENS = {}",
+                    pair.source.tokens(),
+                    MAX_CALIBRATION_TOKENS
+                )));
+            }
+        }
+
+        let mut layers = Vec::with_capacity(calibration.len());
+        for pair in calibration.iter() {
             let w_k = ridge_least_squares(&pair.source.keys, &pair.target.keys, ridge_lambda)?;
             let w_v = ridge_least_squares(&pair.source.values, &pair.target.values, ridge_lambda)?;
             layers.push(LayerMaps { w_k, w_v });
--- a/crates/ruvllm/src/kv_migration/mapper.rs
+++ b/crates/ruvllm/src/kv_migration/mapper.rs
@@ -1,0 +1,301 @@
+//! KV mappers: the [`KvMapper`] trait and the closed-form
+//! [`LinearKvMapper`] from arXiv:2608.03893.
+//!
+//! ## Approach
+//!
+//! For each transformer layer `l`, the linear mapper learns two matrices
+//! `W_k[l]` and `W_v[l]` of shape `[src_dim, tgt_dim]` such that
+//!
+//! ```text
+//! K_tgt ≈ K_src · W_k[l]      V_tgt ≈ V_src · W_v[l]
+//! ```
+//!
+//! where `K_src` is `[tokens, src_dim]` and `K_tgt` is `[tokens, tgt_dim]`,
+//! with `dim = num_kv_heads * head_dim` (the flattened per-token KV vector).
+//!
+//! The fit is the closed-form ridge-regularized least-squares solution
+//! (normal equations):
+//!
+//! ```text
+//! W = (Xᵀ X + λ I)⁻¹ Xᵀ Y
+//! ```
+//!
+//! solved by Gaussian elimination with partial pivoting — no external
+//! linear-algebra dependency beyond `ndarray`, which the crate already uses.
+//!
+//! ## Head-count / head-dim mismatch
+//!
+//! Following the paper, geometry mismatches between source and target
+//! (different `num_kv_heads`, different `head_dim`, or both) are handled by
+//! learning the projection over the *flattened* per-token KV vector:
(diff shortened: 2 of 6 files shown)


ruvnet:

Audited at `34a031a22`. Both findings were reproduced with a runnable PoC against this branch in a **release build** (throwaway worktree, no repo changes).

**Verdict: FINDINGS** — 2 medium, 2 low.

### Checked and clean

- **Feature-flag isolation holds.** `kv-migration` is not in `default = ["async-runtime", "candle", "routing-metrics", "quantize", "hub-download"]`, no other feature enables it, and the only references anywhere in the tree are the module gate itself (`src/lib.rs:133-134`) and the feature declaration. Verified by grepping the whole branch.
- **Numeric edge cases in the solver are handled well.** The pivot check at `src/kv_migration/mapper.rs:259` runs on post-elimination values and guards exactly the diagonal used in back substitution, so no division by a near-zero pivot. Non-finite inputs on the `x` side surface as an error rather than propagating. `ridge_lambda` is validated positive-and-finite (`mapper.rs:144`).
- **The gate is correctly fail-closed on non-finite scores** (`src/kv_migration/quality.rs:174-178`) — a NaN mapper (e.g. from non-finite target activations, which are otherwise unvalidated) produces NaN cosines and vetoes into re-prefill. Good design.
- Control test confirms the gate works normally: a junk mapper against an honest validation set yields `worst_cosine=-0.18, worst_rel_err=1.02, decision=Reprefill`.

---

### MEDIUM — an all-zero validation set makes the gate approve any mapper unconditionally

`src/kv_migration/quality.rs:206-207` and `:222-223`

```rust
if na == 0.0 || nb == 0.0 { return if na == nb { 1.0 } else { 0.0 }; }   // cosine_similarity
if norm == 0.0 { return if diff == 0.0 { 0.0 } else { f32::INFINITY }; } // relative_error
```

When prediction and target are **both** all-zero, cosine returns a perfect `1.0` and relative error a perfect `0.0`. Since `pred = source · W` is zero whenever the source is zero, a validation pair of zero-filled activations scores a flawless pass for *any* `W`. Reproduced against a deliberately garbage-fitted mapper:

```
zero validation set -> worst_cosine=1 worst_rel_err=0 decision=Migrate
```

**Exploit scenario:** the module's stated contract is "we NEVER migrate blind" (`src/kv_migration/mod.rs:12-14`). Any path that yields zeroed validation activations — a model that failed to load into an uninitialised buffer, a truncated or crafted calibration file, a wrong-shaped tensor read — silently converts the gate into an unconditional approval, and a badly-mapped KV cache is injected into target-model attention. Failure is silent, since the report shows a perfect score.

**Suggested fix:** treat an all-zero (or degenerate-norm) validation pair as a veto rather than a pass, and reject a validation set whose target activations are entirely zero in `TransferQualityGate::new`. The `na == 0 || nb == 0` branch should return `0.0` in every case, or better, a sentinel that forces `Reprefill`.

---

### MEDIUM — the validation set's target dimension is never checked; release builds score a truncated prefix

`src/kv_migration/quality.rs:135-154`, `:197`, `:214`

`assess()` validates the layer **count** (`quality.rs:136`) and `map_layer()` validates the **source** dim (`mapper.rs:200`), but nothing validates that `pair.target` has `mapper.target_dim()` columns. The comparison helpers guard shape with `debug_assert_eq!(a.shape(), b.shape())` — compiled out in release — and then `a.iter().zip(b.iter())` **silently truncates to the shorter operand**.

Reproduced in a release build with a mapper whose `target_dim` is 64 and a validation target of dim 1:

```
pred keys shape = [8, 64] (mapper target_dim = 64)
validation target shape = [8, 1]
dim-mismatched validation -> worst_cosine=1 worst_rel_err=0 decision=Migrate
```

The gate approved after comparing **8 of 512** predicted elements.

**Exploit scenario:** a malformed or crafted validation set passes the gate on a vanishing fraction of the data. Worse, it is a debug/release behaviour split: CI running debug tests panics on the assert, so the condition looks impossible, while production silently approves.

**Suggested fix:** validate in `TransferQualityGate::new` / `assess` that every `pair.target.dim() == mapper.target_dim()` and `pair.source.dim() == mapper.source_dim()`, returning `RuvLLMError::KvCache`. Promote the two `debug_assert_eq!` shape checks to real `assert_eq!` or an error return — they are cheap relative to the O(n·m) loops they guard.

---

### LOW — `fit()` has no bound on dimension or layer count; input size O(t·n) buys work O(n³)

`src/kv_migration/mapper.rs:214-235`, `:243-301`

`ridge_least_squares` allocates an `[n, n]` Gram matrix and `solve_linear_system` is O(n³ + n²m), where `n = source_dim` comes straight from the caller's calibration arrays with no upper bound — as does `calibration.len()`. The amplification matters: calibration input is O(tokens · n) but the solve is O(n³). At a realistic `n = 8192` that is ~5.5e11 operations per matrix, times 2 (K and V) times the layer count. The `[n, n]` allocation is also unbounded (n = 65536 would request ~17 GB, and a Rust allocation failure aborts the process rather than returning an error).

No caller exists in this slice, so this is a latent API hazard rather than a live DoS. Worth bounding before anything wires `fit()` to values derived from a model file.

**Suggested fix:** validate `source_dim`, `target_dim`, and `calibration.len()` against configured maxima in `fit()`, and document the O(n³) cost on the function.

---

### LOW — unchecked `usize` overflow in `LayerKvTensor::from_flat`

`src/kv_migration/mod.rs:77`

```rust
let dim = num_kv_heads * head_dim;
```

In release this wraps silently. The `dim == 0` check at `:78` catches wraps-to-zero, but not wraps-to-small: `num_kv_heads = 2^62 + 1, head_dim = 4` wraps to `dim = 4`, and the tensor is then reinterpreted with a completely wrong stride while every subsequent check passes. Not memory-unsafe (`Array2::from_shape_vec` still validates `tokens * dim == len`), just silently wrong geometry. It matters because these two values are exactly the kind that come from a model header — the sibling GGUF work makes that a plausible near-term source.

**Suggested fix:** `num_kv_heads.checked_mul(head_dim).ok_or_else(|| RuvLLMError::KvCache(...))?`.

---

*Minor, non-security:* `DEFAULT_RIDGE_LAMBDA` is not re-exported from `kv_migration` (only from `kv_migration::mapper`), unlike its siblings at `mod.rs:37-40`. Non-finite values in the **target** activations are not validated in `fit()`; they yield a NaN `W` and a successful `Ok(...)`. The quality gate catches that downstream, so it is contained — but an explicit finiteness check in `fit()` would report the real problem at its source.

ruvnet:
## Security findings addressed — fixed in 8fe6d63aa

All four Phase-4 findings are fixed on this branch. Finding → fix mapping:

**MED — degenerate validation pairs defeat the gate** (`quality.rs`)
- `cosine_similarity` now returns **-1.0** (worst possible) when either operand's squared norm is below `DEGENERATE_NORM_EPS` (1e-12); `relative_error` returns **`f32::INFINITY`** for a degenerate target, even when the prediction is also zero. A zeroed pair is therefore a hard veto, never a perfect pass (`INFINITY` also trips the existing non-finite fail-closed check).
- `TransferQualityGate::new` now **rejects** validation sets in which the majority of pairs (including all of them) have near-zero-norm activations on either side — an all-zero set can no longer even construct a gate, so the caller is forced down the re-prefill path.
- Tests: `all_zero_validation_set_is_rejected_never_migrates` (entirely and majority degenerate → construction error), `degenerate_pairs_in_mixed_validation_score_as_failures` (minority degenerate pair scores cosine=-1.0 / rel-err=inf → `Reprefill`, not a 1.0/0.0 pass).

**MED — dimension mismatch only debug-checked** (`quality.rs`)
- `assess()` now performs **real runtime validation** that every pair's `source.dim()` / `target.dim()` matches the mapper's `source_dim()` / `target_dim()`, returning `RuvLLMError::KvCache` before any scoring. No release/debug behavior split, no zip truncation.
- The two `debug_assert_eq!` shape guards in the comparison helpers are promoted to real `assert_eq!` as defense-in-depth.
- Test: `dim_mismatched_validation_is_explicitly_rejected` (both target-dim and source-dim mismatch → explicit error; verified passing in a **release** build, where the old debug_assert compiled out).

**LOW — unbounded `fit()`** (`mapper.rs`)
- New documented bounds validated up front, before any O(n³) solve or `[n, n]` Gram allocation: `MAX_FIT_DIM = 8192`, `MAX_CALIBRATION_LAYERS = 512`, `MAX_CALIBRATION_TOKENS = 2^20` (all `pub`, re-exported from `kv_migration`). The O(tokens·n² + n³ + n²·m) cost is documented on `fit()`, and zero dims are rejected too.
- Test: `fit_rejects_oversized_calibration_inputs` (dim, layer-count, and token-count bounds each rejected before any solve).

**LOW — usize overflow in `LayerKvTensor::from_flat`** (`mod.rs`)
- `num_kv_heads * head_dim` is now `checked_mul` with a clear `RuvLLMError::KvCache` overflow error — no silent wrap to a wrong stride.
- Test: `from_flat_dimension_overflow_is_rejected`.

Also picked up the minor note: `DEFAULT_RIDGE_LAMBDA` is now re-exported from `kv_migration` alongside its siblings.

### Test evidence

```
cargo test -p ruvllm --features kv-migration kv_migration
  test result: ok. 12 passed; 0 failed   (7 existing + 5 new)

cargo test -p ruvllm --features kv-migration --release kv_migration
  test result: ok. 12 passed; 0 failed   (release semantics for the debug_assert finding)

cargo check -p ruvllm            # without the feature
  Finished `dev` profile — clean
```

ruvnet:
## Security re-verification — fix `8fe6d63aa`

Re-audited the KV-migration quality gate and mapper at the fix commit. **Verdict: CLEAR-TO-MERGE.** All four findings (2 MED, 2 LOW) are closed. Ran the suite in **debug and release, 12/12 both**, plus `cargo check -p ruvllm` with the feature off (clean). Read the diff and worked through the three adversarial angles from the re-verify request.

- **MED (degenerate set approves any mapper):** a near-zero-norm operand now scores `cosine = -1.0` / `relative_error = f32::INFINITY` (hard veto, never the old perfect `1.0`/`0.0`), gated on `DEGENERATE_NORM_EPS = 1e-12`. `TransferQualityGate::new` additionally rejects a majority-degenerate validation set up front. `all_zero_validation_set_is_rejected_never_migrates` covers both the all-degenerate and 2-of-3 cases.
- **MED (dim not validated / release truncation):** `assess()` now validates every pair's `source.dim()`/`target.dim()` against the mapper at runtime (returning `RuvLLMError::KvCache`), and the two comparison helpers' `debug_assert_eq!` shape guards are promoted to real `assert_eq!` — so a release build can never score a `zip`-truncated prefix. Proven by `dim_mismatched_validation_is_explicitly_rejected` passing identically in **release**.
- **LOW (unbounded `fit`):** `MAX_FIT_DIM = 8192`, `MAX_CALIBRATION_LAYERS = 512`, `MAX_CALIBRATION_TOKENS = 2^20` are all checked **before** the first `[n,n]` Gram allocation or O(n³) solve; the cost is documented on `fit`. `fit_rejects_oversized_calibration_inputs` exercises all three.
- **LOW (`from_flat` overflow):** `num_kv_heads.checked_mul(head_dim)` rejects the wrap instead of producing a wrong stride. `from_flat_dimension_overflow_is_rejected` covers it.

### Adversarial checks
- **Exactly-half degenerate (not majority) gaming the aggregate?** Cannot. `new()` rejects a *strict* majority (`degenerate*2 > len`), so a 2-of-4 set is admitted — but the decision aggregates `worst_cosine = fold(min)` and `worst_relative_error = fold(max)` over all layers, so a **single** degenerate layer forces `worst_cosine = -1.0` and `worst_relative_error = INF`, and the `clean = worst.is_finite()` guard vetoes regardless. One bad pair vetoes the whole assessment; the minority case can't survive the min/max fold. `degenerate_pairs_in_mixed_validation_score_as_failures` asserts exactly this (`report.worst_cosine == -1.0`, decision `Reprefill`).
- **Is `-1.0` safe through every aggregation path (no `abs()`)?** Yes. The only `abs()` calls in the module are pivot-magnitude selection inside the linear solver (`mapper.rs:309/311`) — nowhere near scoring. In `quality.rs` the decision path is min-fold on cosine + finite-check + threshold compare; `mean_cosine` is report-only and unused in the decision. `-1.0` flows through `min` unchanged and correctly fails `min_cosine` (0.92).
- **Do the new bounds reject legitimate 70B-class geometries?** No. `MAX_FIT_DIM` bounds the *flattened per-token KV dim* (`num_kv_heads * head_dim`). GQA 70B-class models (Llama-3-70B, Qwen2.5-72B) use ~8 KV heads × 128 = **1024** — well under 8192; even a hypothetical full-MHA 64-KV-head × 128 model lands at exactly 8192 (allowed, since the check is `>`). `MAX_CALIBRATION_LAYERS = 512` >> the ~80–126 layers of those models, and `2^20` tokens/layer is ample for held-out calibration. Legitimate geometries pass.

Clean, well-tested hardening across both build profiles. Ship it — this completes the Wave-1 fix slate.

Issue #845, opened 2026-08-20
[PIR][WP0a] ADR numbering: freeze historical duplicates, generated INDEX.md as canonical counter, CI collision check


ruvnet:
Epic: #837 · Plan: `docs/research/perpetual-intelligence-runtime/03-program-plan.md`

> **⚠️ Acceptance criteria were rewritten after work started.** This issue originally specified **renumbering** the later file in each collision. The approach is now **freeze historical duplicates, no renames** — see "Why the approach changed" below. Coordinator-approved. The original criteria are preserved at the bottom of this issue so the change is auditable. The assigned owner's instructions already match the freeze approach, so no re-sync is needed.

## Goal

Make `docs/adr/` unambiguous to look up, and make it impossible for the collision count to grow — **without renaming any existing ADR**.

Three deliverables: a **policy ADR**, a **generated `INDEX.md`** that serves as the canonical counter, and a **CI collision check**.

## Evidence — ground truth (supersedes all earlier counts)

Measured by `scripts/adr-index.mjs` in PR #857:

- **336 ADR files** — 289 plain + 47 namespaced.
- **27 duplicated plain numbers, spanning 61 files.**
- **ADR-272 ×5**; **ADR-264, 252, 194, 144 ×3 each**; **22 numbers ×2**.

> **Correction to this issue's original list: `ADR-040` ×3 is NOT a collision.** `ADR-040` / `ADR-040a` / `ADR-040b` is the intentional sub-ADR convention, and the original evidence table wrongly counted it among the ×3 collisions while a note further down correctly called it intentional — the body contradicted itself. The policy must keep this convention working.

Example collisions, cited by filename since the numbers are ambiguous by construction:

| Number | Documents |
|---|---|
| 134 | `ADR-134-ruvector-claude-code-deep-integration.md` vs. `ADR-134-witness-schema-log-format.md` |
| 256 | `ADR-256-hybrid-sparse-dense-search.md` vs. `ADR-256-metaharness-sdk-evaluation.md` |
| 266 | `ADR-266-metaharness-darwin-ann-optimization.md` vs. `ADR-266-metaharness-darwin-integration.md` |
| 090 | `ADR-090-implementation-checklist.md` vs. `ADR-090-ultra-low-bit-qat-pi-quantization-ddd.md` |

## Why this is Phase 0 and blocks everything

Several colliding numbers are load-bearing PIR citations:

[...]

PR #857, opened 2026-08-20
docs(adr): numbering hygiene — freeze duplicates, canonical index + collision check (PIR WP0a)

ruvnet:
Refs #845 · Epic #837 · PIR WP0a

## What this does (policy + index + gate — NOT mass renumbering)

Audited every ADR in `docs/adr/` (including the namespaced `coherence-engine/`, `delta-behavior/`, `quantum-engine/`, `temporal-tensor-store/` families) and shipped the minimal safe fix: a written policy, a canonical generated index, and a CI-able collision gate. **No existing ADR file is renamed** — renumbering breaks external references (commit messages, cross-repo citations, the shared brain) that redirect stubs cannot cover.

## Verified ground truth (corrects the issue body)

- **336 ADR files**: 289 on the plain `ADR-NNN` counter + 47 namespaced (`ADR-CE/DB/QE/TTS-*`, no internal duplicates)
- **27 duplicated plain numbers spanning 61 files** (issue estimated "~15+"):
  - ×5: **ADR-272**
  - ×3: ADR-264, ADR-252, ADR-194, ADR-144
  - ×2: ADR-268, 266, 260, 258, 256, 254, 143, 139, 138, 137, 136, 135, 134, 133, 132, 117, 096, 095, 094, 093, 091, 090
- **ADR-040 is NOT a collision**: `040`/`040a`/`040b` is the intentional sub-ADR convention; the tooling treats lettered suffixes as distinct keys and preserves it.

## Changes

- **`docs/adr/ADR-316-adr-numbering-hygiene.md`** — the policy: duplicates are frozen historical artifacts; disambiguate as `ADR-NNN (slug)`; new numbers from a single canonical counter; the generated index is the counter's source of truth. Numbered **316** because 305–315 are claimed on `feat/pir-adrs` (#847).
- **`docs/adr/INDEX.md`** — generated index: **next available number (currently ADR-317)** at top, then number | title | file | last-commit date | status | DUPLICATE flag (with its `ADR-NNN (slug)` disambiguator) for all 337 entries.
- **`scripts/adr-index.mjs`** — node, zero deps. Default mode regenerates the index; `--check` exits nonzero if `docs/adr/` contains a duplicate outside the hardcoded frozen list, or if a frozen number's file count grows. Ready to wire into CI.

## Test evidence

```
$ node scripts/adr-index.mjs --check
ADR duplicate-number check OK: 337 ADR files, no duplicates outside the frozen historical list. Next available number: 317.

# simulated NEW collision (second ADR-316 file):
ADR duplicate-number check FAILED:
  - NEW duplicate 316: docs/adr/ADR-316-adr-numbering-hygiene.md, docs/adr/ADR-316-simulated-collision.md
exit=1

# simulated 6th ADR-272 file:
  - Duplicate count for ADR-272 grew: 6 files (frozen historical allowance is 5): ...
exit=1
```

## Acceptance-criteria mapping (re-scoped per ADR-316)

- Renumbering / reference updates / redirect stubs → **moot by policy** (freeze + slug disambiguation)
- Written policy incl. `NNNa`/`NNNb` interaction → ADR-316
- CI check for future duplicates → `scripts/adr-index.mjs --check`
- Allocation from true max → INDEX.md "next available" header (currently 317, accounts for 305–315 on #847)


ruvnet:
Rebased onto `main` after #847 merged (0049892cb). `docs/adr/INDEX.md` regenerated: now **348 entries** including rows for ADR-305–315 alongside ADR-316; next available number reads **317**; `scripts/adr-index.mjs --check` passes (305–315 are unique, still exactly 27 frozen duplicate numbers). ADR-316's context updated to note 305–315 have merged. No content conflicts with the 305–315 set.

ruvnet:

Audited `scripts/adr-index.mjs`. (Audit ran against `7faf931b2`; re-verified byte-identical at the current head `6bc3d1d90` after the rebase, so all findings stand.) Exploits were confirmed against a copy of the script in an isolated scratch repo — nothing in this repo was modified.

**Verdict: FINDINGS** — 2 medium, 1 low, 1 informational.

### Checked and clean

- **No shell injection.** `scripts/adr-index.mjs:81-84` uses `execFileSync('git', ['log', '--follow', '-1', '--format=%ci', '--', relPath], ...)` — an argv **array**, no `shell: true`, so filename metacharacters are inert. **Argument injection is also handled**: the `--` terminator means a filename beginning with `-` cannot be read by git as a flag. (`FILE_RE` additionally requires an `ADR-` prefix.) This class is genuinely absent.
- **No path traversal from external input.** `REPO_ROOT` derives from `import.meta.url`; `ADR_DIR` and `INDEX_PATH` are constants (`:25-27`). `process.argv` is used only for `.includes('--check')` (`:210`) and never reaches a path. No env var is read. No ADR frontmatter field is used to build a path — only `title` (`:60-63`) and `status` (`:65-77`) are parsed, both as display strings. The single write target is the fixed `INDEX_PATH`.
- Newlines in filenames are safe (`FILE_RE` has no `m` flag, `.` excludes `\n`), and unreadable ADR files are handled by the try/catch at `:102-106`.

---

### MEDIUM — `walk()` follows symlinks, so a symlinked file or directory in `docs/adr/` is read and inlined into `INDEX.md`

`scripts/adr-index.mjs:50-58`, exploited via `:103` -> `:202`

```js
if (statSync(p).isDirectory()) out.push(...walk(p));   // statSync FOLLOWS symlinks
else if (entry.endsWith('.md')) out.push(p);
```

`statSync` (not `lstatSync`) resolves symlinks and `readFileSync(abs)` then reads through them. Git stores symlinks, so a PR branch can add one. Confirmed:

- `ln -s /path/outside/creds.txt docs/adr/ADR-500-exfil.md` produced the row
  `| ADR-500 | TOP SECRET DEPLOY KEY abc123 | ... |`
- A **directory** symlink escapes the tree entirely: `ln -s /path/outside docs/adr/vendor` indexed `vendor/ADR-600-outside.md` with title `OUTSIDE REPO SECRET`, and even moved the "highest allocated number" counter to 600.

Leakage is bounded by what the parsers extract — the first `^# ` heading and a `status:` line — but that is enough to lift one line from a `.env`, `~/.gitconfig`, or an agent config, and to enumerate filenames outside `docs/adr/` into a committed file.

**Severity note:** medium *today* only because the script is not yet wired into CI (see below). A `pull_request`-triggered job checking out PR head turns this into an automated exfiltration path with the result visible in the PR diff — that would be high.

**Suggested fix:** use `lstatSync` in `walk` and skip entries where `isSymbolicLink()`; defensively assert each resolved path stays under `ADR_DIR` (`realpathSync(abs).startsWith(realpathSync(ADR_DIR) + sep)`).

---

### MEDIUM — unguarded `readdirSync`/`statSync` crash the process on a broken symlink or symlink loop

`scripts/adr-index.mjs:52-54`

Neither call is wrapped, and there is no visited-inode set. Confirmed:

- Dangling symlink `ADR-502-broken.md -> /nonexistent/target.md` -> uncaught `Error: ENOENT ... at walk (adr-index.mjs:54:9)`, process aborts with a raw stack trace.
- `ln -s . docs/adr/loop` -> infinite recursion until `ELOOP`, then abort. On a platform where the loop terminates less promptly this is a hang rather than a crash.

`--check` — the advertised gate — therefore dies with an unhandled exception instead of a diagnosable failure. It does exit nonzero, so the gate fails closed, but a stack trace reads as "the script is broken", which is what gets a gate disabled or `|| true`'d. Dangling symlinks are not exotic; they show up from ordinary checkouts of branches with reorganized paths.

**Suggested fix:** `readdirSync(dir, { withFileTypes: true })` (removes the per-entry `stat` entirely), skip `dirent.isSymbolicLink()`, and wrap the read in try/catch to skip-and-warn on EACCES/ENOENT.

---

### LOW — filename-derived text is interpolated into `INDEX.md` unescaped

`scripts/adr-index.mjs:200` and `:202`

`mdCell()` (`:164-166`) escapes `|` and collapses whitespace, but is applied **only** to `e.title` and `e.status`. Three interpolations bypass it: the raw `e.slug` at `:200` and the raw `e.rel` twice at `:202` (link text and link target). Both come straight from the filename, and POSIX/git permit `|`, `[`, `]`, `(`, `)` and RTL-override characters there.

Confirmed: a file named `ADR-090-beta|BROKEN](httpX) [pwn](httpY.md` split the table row into bogus cells and escaped the link, rendering an attacker-controlled misleading link in `docs/adr/INDEX.md`. GitHub sanitizes `javascript:` and `<script>`, so this is defacement/phishing-grade rather than XSS — though `mdCell` doesn't escape `<`/`>` either, which matters if INDEX.md is ever rendered by a viewer permitting raw HTML.

**Suggested fix:** run `e.slug` and `e.rel` through `mdCell()` as well, and `encodeURI` the link target.

---

### INFORMATIONAL — the "CI gate" is not actually wired to CI

`grep -rn "adr-index" .github/workflows/` returns nothing; a repo-wide search finds only prose references in `docs/research/perpetual-intelligence-runtime/03-program-plan.md`. There is no `package.json` script, no `.husky/`, no `.pre-commit-config.yaml`. So `scripts/adr-index.mjs:186` ("CI gate: `node scripts/adr-index.mjs --check`") and the ADR-316 text describe intent, not current state — the duplicate-number gate promised for #845 is not enforced by this PR.

**Please fix the two medium findings before wiring this into a workflow**, since that is the change that makes both of them remotely reachable.

---

*Non-security robustness notes:* `abs.split('/').pop()` (`:95`) hardcodes the POSIX separator — on Windows `FILE_RE` never matches, `collect()` returns `[]`, and `--check` **passes vacuously** (fails open). Use `basename()` from `node:path`, already imported. Separately, `Math.max(...)` at `:159` and `:172` yields `-Infinity` on an empty set, printing "Next available number: -Infinity"; `reduce` is safer.

ruvnet:
## Phase-4 security findings — fixed in af2a3f7de

All four findings from the audit above are addressed in `scripts/adr-index.mjs` (commit af2a3f7de). Finding → fix:

| Finding | Fix |
|---|---|
| **MED — `walk()` follows symlinks** (exfil into INDEX.md via file or directory symlink) | `walk()` now uses `readdirSync(dir, { withFileTypes: true })` and **never follows symlinks** — `dirent.isSymbolicLink()` entries (file or directory) are skipped with a `adr-index: warning: skipping symlink (not followed): …` line. Belt-and-braces: every kept entry is `realpathSync`-asserted to resolve under `realpathSync(ADR_DIR)` (`real === rootReal \|\| real.startsWith(rootReal + sep)`); anything escaping the tree is skipped with a warning. The audit's two PoCs (`ADR-500-exfil.md → outside file`, `vendor → outside dir` moving the counter to 600) are both reproduced as self-test fixtures and confirmed inert — no content inlined, counter unaffected. |
| **MED — fragile traversal** (broken symlink → raw ENOENT stack trace; `ln -s .` → ELOOP) | Per-entry `try/catch` with skip+warn on unreadable directories and unresolvable entries; a visited-realpath set guards directory cycles (symlink-skipping already removes the loop vector — the set is defense in depth); the top-level dispatch is wrapped so any residual error exits 1 with `adr-index: fatal: <message>`, never a stack trace. `--check` on a tree containing a dangling symlink **and** `loop → .` exits 0 with warnings and its normal summary (asserted by self-test, including "no `at walk` frame in output"). |
| **LOW — slug/rel interpolated unescaped** (`\|`, `](…)` in a filename break the table / forge links) | `e.slug` (both at the duplicate-marker and elsewhere) and `e.rel` now go through `mdCell` / `mdLinkText` (escapes `\|`, `[`, `]`, collapses whitespace); the link target goes through `encodeURI` plus explicit `%28`/`%29` for parens (which `encodeURI` leaves alone). Tested with the audit's exact hostile name `ADR-090-beta\|BROKEN](httpX) [pwn](httpY.md`: the row keeps exactly 7 unescaped pipes (cells intact) and the target contains no raw space/`\|`/`[`/`]`/`(`. |
| **INFO — Windows fail-open + robustness notes** | `abs.split('/')` replaced with `path.basename()`; display paths built via `path.relative` + `sep`-normalization, so `FILE_RE` matches on Windows and `--check` can no longer pass vacuously. `Math.max(...)` at both sites replaced with `reduce` (empty set → 0, so "Next available number: 1", not `-Infinity`). |

### Test

New `node scripts/adr-index.mjs --self-test` mode: builds an adversarial fixture tree in a temp dir (outside-tree file symlink, escaping directory symlink, broken symlink, `loop → .`, the hostile filename above), copies the script into it, runs `--check` and generate as subprocesses, and asserts 12 properties. All pass:

```
  PASS  --check exits 0 on adversarial tree (no dupes present)
  PASS  --check prints its normal summary
  PASS  no stack trace on broken symlink / symlink loop
  PASS  symlinks produce a skip warning
  PASS  generate exits 0 on adversarial tree
  PASS  file symlink target content is NOT inlined into INDEX.md
  PASS  directory symlink is NOT followed
  PASS  symlinked/broken entries are absent from the index
  PASS  hostile-named ADR is still indexed
  PASS  hostile filename: table row has exactly 7 unescaped pipes (cells intact)
  PASS  hostile filename: link target is URI-encoded (no raw space | [ ] ( )
  PASS  counter unaffected by symlinked ADR-500/600
```

### Regression evidence

- `node scripts/adr-index.mjs --check` → `ADR duplicate-number check OK: 348 ADR files, no duplicates outside the frozen historical list. Next available number: 317.`
- Regenerating INDEX.md after the escaping change produces **no diff** — all current filenames are benign.

### On CI wiring (the informational finding)

Agreed: `--check` is intentionally **not** wired into a workflow in this PR. That wiring should land as a follow-up **after** this commit merges, since exposure to PR-controlled trees is exactly what upgraded the two mediums. With this commit the gate is safe to wire.

ruvnet:
## Security re-verification — fix `af2a3f7de`

Re-audited `scripts/adr-index.mjs` at the fix commit. **Verdict: CLEAR-TO-MERGE.** All four findings from the original review are closed. Ran `--self-test` (12/12 pass) and `--check` against the real ADR tree (348 files, exit 0, no stack traces), and threw an independent adversarial fixture at it that goes beyond the bundled self-test.

- **MED (symlink following → exfil):** `walk()` now tests `dirent.isSymbolicLink()` **first** and skips file *and* directory symlinks with a warning, before any `isDirectory`/`isFile` branch — plus a belt-and-braces `realpathSync` containment check (`real === rootReal || real.startsWith(rootReal + sep)`) on every kept file. My red-team fixture confirmed a file symlink to an out-of-tree secret, a `../../../` escape symlink, and a nested `sub/vendor` directory symlink to an outside dir are all skipped and never inlined into `INDEX.md`.
- **MED (crash/hang on broken symlink or loop):** every `readdirSync`/`realpathSync` is wrapped with warn-and-skip; a `visited` set of realpath'd dirs breaks cycles; the top-level handler prints `adr-index: fatal:` and exits cleanly. No `at walk` stack trace on the adversarial tree, and `--check` still exits 0.
- **LOW (markdown injection via filename):** `mdCell` (pipes), `mdLinkText` (brackets), and `mdLinkTarget` (`encodeURI` + `(`/`)`) are now applied to slug/rel/link-target. The hostile `ADR-090-beta|BROKEN](httpX) [pwn](httpY.md` name renders with exactly 7 unescaped pipes (cells intact) and a URI-encoded target.
- **INFO (Windows fail-open):** now uses `basename()` + `relative(...).split(sep).join('/')`, so matching no longer depends on the POSIX separator; `maxPlainNum` uses `reduce` (0, not `-Infinity`, on empty).

### Attempts to defeat the containment (all held)
Inside-resolving symlink, `../..` escape, and nested-dir symlink — all skipped because symlinks are refused unconditionally, independent of where they point. Containment can't be reached by a crafted target.

### Two residuals — both **out of the PR/CI threat model**, non-blocking
- **Hardlink defeats realpath containment.** A hardlink placed in `docs/adr/` resolves *inside* the tree, so it passes the check and its content is read (I reproduced a `# HEADING` + `**Status**` leak). But this is **not reachable through a PR:** git blob-izes hardlinked files — a committed hardlink becomes an independent blob, and CI's checkout writes a regular file containing only the committed content, never a link to any out-of-tree victim file. Exploiting it needs local write access into the checkout dir adjacent to the run, which is outside this script's model.
- **TOCTOU** between the `isSymbolicLink()`/`realpath` check and `readFileSync`. Same reason it's moot: a PR-triggered CI checkout is static, with no concurrent adversary able to swap an entry mid-run.

Neither needs to block merge. If you want defense-in-depth anyway, an `lstatSync(abs).nlink > 1` skip would close the hardlink case, but I would not gate on it. Good, thorough hardening — ship it.

Issue #846, opened 2026-08-20
[PIR][WP0b] Fix GGUF weight downloads (Darwin live-serve blocker), bump @metaharness, fix optionalDependencies non-compliance


ruvnet:
Epic: #837 · Plan: `docs/research/perpetual-intelligence-runtime/03-program-plan.md`

Three small, concrete dependency and download-path items that together gate WP9.

## (1) The real Darwin live-serve blocker: GGUF weight downloads

> **Scope correction.** An earlier program pass listed an `ruvllm` **HTTP-307 redirect bug** as the blocker. **That bug is already fixed on `main`** — commit `946275a61` (PR #590, merged 2026-06-19) added a `curl -L` fallback in `crates/ruvllm-cli/src/commands/download.rs`. ADR-259's "blocked" wording predates the fix *within the same PR* and is stale. **Do not re-fix the redirect.**

The remaining blocker is a different bug in the same command: **GGUF weight downloads still fail.**

- `get_files_to_download()` requests an **unexpanded glob** — `*<suffix>.gguf` is sent literally and returns **404**.
- The **GGUF alias points at the safetensors repo**, so even a corrected request resolves to the wrong artifact.

Fixing it needs **HuggingFace file listing** (expand the glob against the actual repo file list rather than sending a pattern) plus **registry resolution** (make the GGUF alias resolve to the GGUF repo).

This is what actually prevents Darwin's local mutator backend from passing a live-serve end-to-end test.

## (2) `@metaharness` version bump

`ruvnet/dream-machine` composes **`@metaharness/darwin` 0.9.1** and **`@metaharness/flywheel` 0.1.10**. This repo pins **0.8.0** and **0.1.7**. WP2's integration cannot be trusted until the two agree — bump to the 0.9.x / 0.1.10 line, or record an explicit reason not to.

## (3) MetaHarness dependency-declaration compliance

[...]

Issue #841, opened 2026-08-20
[PIR][WP9] Frozen-weight skill/harness evolution loop via @metaharness/darwin (SHAPER pattern, arXiv:2608.11350)


ruvnet:
Epic: #837 · Plan: `docs/research/perpetual-intelligence-runtime/03-program-plan.md`

## Goal

Implement the physical-intelligence evolution loop with **foundation-model weights frozen throughout**: planner and optimizer are the same frozen model, and only skills, context, and the execution harness evolve.

Darwin's mutation surfaces already exist at:

- `harness/src/darwin.ts` (in `crates/ruvector-sota-bench/harness`, via `@metaharness/darwin`)
- `examples/mragent` — `scorePolicy`
- `crates/sona/src/darwin_guard.rs`

## Evidence

[arXiv:2608.11350](https://arxiv.org/abs/2608.11350) — "Self-Evolving Embodied Agents via Skill-Harness Evolution" (SHAPER), submitted 2026-08-11. Grade A. A train-free framework keeping model parameters frozen while evolving reusable skills and a context/code harness through target-environment rollouts, with the same frozen model acting as planner and optimizer. Evaluated on VLABench and ESI-Bench. No name collision — "SHAPER" is specific to this paper.

## Acceptance criteria

- [ ] **A CI check proves no promotion-reachable code path can write to foundation-model weight files.** The check fails the build if any mutation surface reachable from the promotion pipeline imports a training or fine-tuning API. Policy statements are explicitly insufficient — this must be mechanically enforced, and it is the acceptance test's central constraint.
- [ ] **A VLABench/ESI-Bench-style evaluation shows skill improvement without weight changes.**
- [ ] Skills, context, and harness are separately versioned mutation surfaces mapped onto Darwin's existing surfaces; weights are not a mutation surface.
- [ ] Foundation-model weights are hash-verified at loop start, and the hash is recorded in the witness chain.
- [ ] The same frozen model instance serves as planner and optimizer (SHAPER's structural claim), verified by configuration test.
- [ ] Every evolution cycle terminates in a **WP2** `research-gate` verdict — no promotion path bypasses it.

## Dependencies

- **WP0b** (#846) — the MetaHarness hard-dependency non-compliance and the `ruvllm` HTTP-307 download bug both block this. Darwin's local mutator backend (ADR-259) cannot be live-serve tested end-to-end until the 307 bug is fixed, and this WP depends on `@metaharness/darwin`.
[...]

PR #860, opened 2026-08-20
fix(ruvllm-cli): resolve GGUF globs via HF file listing + alias routing (PIR WP0b)

ruvnet:
Closes the GGUF download portion of #846 (PIR WP0b) — the remaining `ruvllm download` blocker that gates WP9 (#841). The HTTP-307 redirect half was already fixed in commit 946275a61 (PR #590); this PR addresses the two bugs that commit explicitly deferred ("needs HF file-listing + registry resolution").

## Root cause

1. **Unexpanded glob sent as a literal filename.** `get_files_to_download()` pushed `format!("*{}", quant.gguf_suffix())` — e.g. `*Q4_K_M.gguf` — which was requested from HF verbatim → 404. It could never have worked: GGUF repos name files per-model (and often lowercase, e.g. `qwen2.5-14b-instruct-q4_k_m-00001-of-00003.gguf`).
2. **Alias forced down the GGUF path against a safetensors repo.** The `phi` alias maps to `microsoft/Phi-4-mini-instruct` (safetensors-only), but `quant != None` (default `q4k`) forced the GGUF branch — so even a correctly expanded pattern had nothing to match.

## Fix

- **HF file listing:** `list_repo_files()` tries `hf-hub`'s `ApiRepo::info()` (siblings), falling back to `curl -L --fail` against `GET /api/models/<id>/tree/<rev>?recursive=true` — the same dependency-free, `HF_TOKEN`-honoring curl idiom 946275a61 introduced for the 307 redirect. The quant pattern is matched against real filenames.
- **Multi-part GGUF: handled, not rejected.** All parts (`…-q4_k_m-NNNNN-of-NNNNN.gguf`) are downloaded in order. Chosen because the flagship `qwen` alias (Qwen2.5-14B-Instruct-GGUF) ships Q4_K_M *only* as a 3-part split — failing on multi-part would leave the primary registry model permanently broken. Matching is case-insensitive and accepts per-preset spelling variants (`f16`/`fp16`).
- **Registry GGUF twins:** `ModelDefinition` gains an optional `gguf_repo`; `resolve_weights_repo(alias, quant)` routes quantized requests to it (`phi` → `bartowski/microsoft_Phi-4-mini-instruct-GGUF`, `mistral` → `bartowski/Mistral-7B-Instruct-v0.3-GGUF`, `llama` → `bartowski/Llama-3.2-3B-Instruct-GGUF`; all three verified live against the HF tree API). `chat`/`serve` use the same resolution so their cache key matches `download`'s. Unknown repos with no GGUF files fail early with the repo's actual file inventory and a hint (`--quantization none`), instead of a 404 on a glob.
- **Aux files filtered by the listing** — GGUF-only repos no longer fail on `tokenizer.json` they don't have (it's embedded in the GGUF).
- **ADR-259:** status line + "Honest gap" updated — 307 redirect fixed 2026-06-18 (946275a61, PR #590); this PR closes the remaining GGUF gap (#846 acceptance criterion).

## Tests

`cargo test -p ruvllm-cli`: **26 passed, 0 failed** (11 new — tree-JSON fixture parsing, multi-part expansion + ordering, uppercase/lowercase repo naming, fp16 variant, no-GGUF / missing-quant error listings, aux filtering, alias-routing decisions). No new network-dependent tests.

Verified live:
- `ruvllm download microsoft/Phi-4-mini-instruct` → routed to the GGUF twin, downloaded the real `microsoft_Phi-4-mini-instruct-Q4_K_M.gguf` (2.5 GB) end-to-end.
- `ruvllm download microsoft/phi-2` → fails early: "No GGUF files found … Available files: … Hint: pass a GGUF repo, or use --quantization none".

Refs #846, #841, ADR-259, 946275a61.

### DIFF
--- a/crates/ruvllm-cli/src/commands/chat.rs
+++ b/crates/ruvllm-cli/src/commands/chat.rs
@@ -49,9 +49,11 @@ pub async fn run(
     draft_model: Option<&str>,
     speculative_lookahead: usize,
 ) -> Result<()> {
-    let model_id = resolve_model_id(model);
     let quant = QuantPreset::from_str(quantization)
         .ok_or_else(|| anyhow::anyhow!("Invalid quantization format: {}", quantization))?;
+    // Resolve to the repo that hosts the weights for this quantization
+    // (GGUF twin for safetensors-only aliases) — must match `download`'s cache key.
+    let model_id = crate::models::resolve_weights_repo(model, quant);
 
     // Print header
     print_header(&model_id, system_prompt, max_tokens, temperature);
@@ -80,7 +82,11 @@ pub async fn run(
             "{}",
             "Loading draft model for speculative decoding...".yellow()
         );
-        let draft = load_model(&resolve_model_id(draft_id), quant, cache_dir)?;
+        let draft = load_model(
+            &crate::models::resolve_weights_repo(draft_id, quant),
+            quant,
+            cache_dir,
+        )?;
 
         if let Some(info) = draft.model_info() {
             println!(
--- a/crates/ruvllm-cli/src/commands/download.rs
+++ b/crates/ruvllm-cli/src/commands/download.rs
@@ -12,7 +12,7 @@ use hf_hub::{Repo, RepoType};
 use indicatif::{ProgressBar, ProgressStyle};
 use std::path::{Path, PathBuf};
 
-use crate::models::{get_model, resolve_model_id, QuantPreset};
+use crate::models::{get_model, resolve_model_id, resolve_weights_repo, QuantPreset};
 
 /// Run the download command
 pub async fn run(
@@ -22,9 +22,13 @@ pub async fn run(
     revision: Option<&str>,
     cache_dir: &str,
 ) -> Result<()> {
-    let model_id = resolve_model_id(model);
     let quant = QuantPreset::from_str(quantization)
         .ok_or_else(|| anyhow::anyhow!("Invalid quantization format: {}", quantization))?;
+
+    // Route quantized (GGUF) requests to the repo that actually hosts GGUF
+    // weights (registry `gguf_repo` twin for safetensors-only aliases).
+    let base_id = resolve_model_id(model);
+    let model_id = resolve_weights_repo(model, quant);
 
     println!();
     println!(
@@ -33,6 +37,13 @@ pub async fn run(
         model_id,
         quant
     );
+    if model_id != base_id {
+        println!(
+            "  {} {} hosts safetensors only; using GGUF twin repo",
+            "Note:".dimmed(),
+            base_id
+        );
+    }
     println!();
 
     // Get model info if available
@@ -62,8 +73,12 @@ pub async fn run(
         api.repo(Repo::new(model_id.clone(), RepoType::Model))
     };
 
-    // Determine files to download
-    let files_to_download = get_files_to_download(&model_id, quant);
+    // List the repo's actual files and expand the quant pattern against them
+    // (previously an unexpanded glob like "*Q4_K_M.gguf" was sent literally -> 404).
+    let remote_files = list_repo_files(&repo, &model_id, revision)
+        .await
+        .with_context(|| format!("Failed to list files for {model_id} on HuggingFace"))?;
+    let files_to_download = get_files_to_download(&model_id, quant, &remote_files)?;
 
     // Create cache directory
     let model_cache_dir = PathBuf::from(cache_dir).join("models").join(&model_id);
@@ -74,6 +89,11 @@ pub async fn run(
     // Download each file
     for file_name in &files_to_download {
         let target_path = model_cache_dir.join(file_name);
+        if let Some(parent) = target_path.parent() {
+            tokio::fs::create_dir_all(parent)
+                .await
+                .context("Failed to create cache subdirectory")?;
+        }
 
         // Check if file exists
         if target_path.exists() && !force {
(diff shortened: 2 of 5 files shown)


ruvnet:

Audited the diff at branch head (`crates/ruvllm-cli/src/commands/download.rs`, `models.rs`). The GGUF resolution / file-listing logic is sound: the tree-JSON parse rejects malformed input and skips directory entries, quant matching is anchored (no false positives across quants or non-`.gguf` files), and multi-part shards sort deterministically. Two findings, both **defense-in-depth on the `curl` fallback / untrusted repo listing** — neither blocks merge, but the first is worth fixing before this ships.

### MEDIUM — `HF_TOKEN` exposed in process argument list
`download.rs:199` (`download_via_curl`) and `download.rs:234` (`list_files_via_curl`) both do:
```rust
args.push("-H".to_string());
args.push(format!("Authorization: Bearer {token}"));
```
The bearer token lands in `curl`'s argv, which is world-readable on the host (`ps auxww`, `/proc/<pid>/cmdline`) for the lifetime of the process. Any local user (or a co-tenant on a shared build/CI box) can harvest a valid `HF_TOKEN` while a download runs. CWE-214.

**Exploit scenario:** on a multi-user or shared-CI host, `while :; do cat /proc/*/cmdline 2>/dev/null | tr '\0' ' ' | grep -o 'Bearer [^ ]*'; done` captures the token the moment any `ruvllm download`/`serve`/`chat` shells out to the curl fallback for a gated/private repo.

**Fix:** keep the token off argv. Pass it via a curl config read from stdin, e.g. build args ending in `--config -` and write `header = "Authorization: Bearer <token>"` to the child's stdin (`Stdio::piped`), or use `-H @-` / a `0600` temp config file removed after the call. The URL argument is already token-free, so the change is localized to these two helpers.

### LOW — remote-controlled filenames joined into the cache path without traversal validation
`download.rs:90`: `let target_path = model_cache_dir.join(file_name);` where `file_name` originates from the repo's own file listing (HF tree API `path`, or `siblings[].rfilename`) — attacker-controlled if a user is pointed at a malicious repo. `create_dir_all(parent)` then materializes any leading directory components, and the file is copied there. Git/HF normalization makes a literal `../` component unlikely to survive server-side, so this is defense-in-depth rather than a demonstrated escape, but the code trusts the remote path string with no `Path::components()` check rejecting `ParentDir`/`RootDir` before the join.

**Fix:** before joining, reject any `file_name` whose components include `..` or an absolute/root/prefix component, and assert the canonicalized `target_path` stays within `model_cache_dir`.

Also noted (informational, no fix required): `model_id`/`revision` flow into the HF API URL path in `list_files_via_curl`; because everything is passed as discrete argv to `curl` (no shell), there is **no** shell/argv injection — a crafted id could at most alter the URL path, a server-side/low-impact concern.

Everything else in scope (glob→listing expansion, the `bartowski` GGUF-twin registry entries as an explicit trust decision, JSON DoS surface) reviewed CLEAN.

ruvnet:
Both Phase-4 security findings are fixed in 8ff3e7c8f (`crates/ruvllm-cli/src/commands/download.rs`) — finding → fix mapping:

**MED — HF_TOKEN on curl argv (CWE-214, `download_via_curl` + `list_files_via_curl`)**
The `-H "Authorization: Bearer <token>"` argv argument is gone from both call sites. When `HF_TOKEN` is set, the args now end with `--config -` and the header is delivered as a curl config line written to the child's stdin (`Stdio::piped`), so the token never appears in `ps` / `/proc/<pid>/cmdline`. The config line is built by a dedicated `curl_auth_config()` which escapes `\`, `"`, CR, and LF, so an odd/hostile token value cannot inject additional curl config directives. When `HF_TOKEN` is unset (or empty) behavior is unchanged: no config, no `--config` flag, anonymous request. No tempfile variant was needed — stdin avoids the on-disk window entirely.

**LOW — remote-controlled filenames joined into the cache path (`run()` download loop)**
Two-layer defense-in-depth before any filesystem write:
1. `validate_remote_file_name()` — rejects empty names and any path component that is not `Component::Normal` (i.e. absolute/root, `..`, `.`, Windows prefixes) before the `model_cache_dir.join(file_name)`.
2. `ensure_under_cache_dir()` — after `create_dir_all(parent)`, canonicalizes both the cache root and the target parent and fails unless the parent resolves under the root (catches symlinked-subdirectory escapes the component check can't see).

**Evidence**
- `cargo test -p ruvllm-cli`: 31 passed, 0 failed (26 existing + 5 new: hostile names `"../x"`, `"/abs"`, `"a/../../x"`, `""`, `".."` all rejected; normal/nested names accepted; canonicalized containment positive+negative; curl config content and escaping).
- `cargo clippy -p ruvllm-cli --all-targets` clean, `cargo fmt --check` clean.
- Manual: verified `--config`-on-stdin actually delivers the `Authorization` header (local echo server: `AUTH=Bearer test-token-xyz`), and ran `ruvllm download hf-internal-testing/tiny-random-gpt2 --quantization none` end-to-end against a scratch cache dir — listing, path guard, and all 5 file downloads succeeded.

ruvnet:
## Security re-verification — fix `8ff3e7c8f`

Re-audited the two GGUF-download findings at the fix commit. **Verdict: CLEAR-TO-MERGE.** Both are closed. Ran the download tests: **15/15 pass**, including the 5 new security tests. Read the diff and worked through each adversarial angle from the re-verify request.

- **MED (CWE-214, `HF_TOKEN` in argv):** both call sites (`download_via_curl`, `list_files_via_curl`) now build the header with `hf_token_curl_config()` → `curl_auth_config()` and pass it via `--config -`, with the config written to curl's **piped stdin** by `run_curl`. The token is no longer in argv. `HF_TOKEN` unset/empty → no config, no `--config` flag (behavior unchanged).
- **LOW (path traversal):** `validate_remote_file_name` rejects empty names and any non-`Normal` path component (absolute, `..`, `.`, Windows prefix) **before** the join; `ensure_under_cache_dir` canonicalizes the created parent and requires it to `starts_with` the canonicalized cache root **after** `create_dir_all` (catching symlinked-subdir escapes). Tests cover `../x`, `/abs`, `a/../../x`, `..`, `./x/../y`, `/etc/passwd`.

### Adversarial checks
- **Remaining argv/env leak (`-v`/trace, error echo)?** No. The curl arg sets are `-L --fail -sS [-o file] [--config -] <url>` — no `-v`/`--trace`/`--trace-ascii`, and `-sS` (silent + show-errors) does not echo request headers on failure. Error messages interpolate only the URL, never the token. One honest note: curl still **inherits `HF_TOKEN` in its environment** (ambient, unused by curl) — but `/proc/<pid>/environ` is readable only by the owning user, so this is not the world-readable `cmdline` exposure the finding was about, and not a regression. Optionally clear it from the child env for tidiness; not required.
- **Config-directive injection via the token value?** No. `curl_auth_config` escapes `\` **first**, then `"`, `\n`, `\r`, emitting a single `header = "..."` line. Inside a curl double-quoted config string, only an unescaped `"` can terminate it and only an unescaped newline can start a new directive — both are escaped, and the backslash-first ordering means a token like `\"` becomes `\\\"` (decoded by curl as `\` + `"`, still in-string), so there is no breakout. `test_curl_auth_config_escapes_special_characters` pins exactly one line regardless of token content.
- **Interaction with existing `--config` usage?** None — this is the only `--config`, and the URL is a separate positional argv arg, so the stdin config contains solely the header line.
- **TOCTOU on the canonicalized containment?** Present in principle (a window between `ensure_under_cache_dir` and the later `copy`), but out of this tool's threat model: the filename traversal is already fully blocked by `validate_remote_file_name`; the canonicalize is defense-in-depth against symlinks that would have to pre-exist in the user's own cache dir. Winning that race needs local write access to the cache dir, at which point the adversary can write the file directly — no escalation. Worth leaving as-is.

Both findings genuinely closed, escaping is robust, no residual token leak of the class originally flagged. Ship it.

Issue #843, opened 2026-08-20
[PIR][WP12] Build the 30-day continuous-run acceptance test harness


ruvnet:
Epic: #837 · Plan: `docs/research/perpetual-intelligence-runtime/03-program-plan.md`

## Goal

Implement the daily-cycle harness that runs the PIR for 30 continuous days and produces a **defensible pass/fail verdict** on whether the system genuinely improved itself under frozen weights, with full provenance and zero unapproved capability expansion.

## Precondition

Foundation-model weights frozen and **hash-verified at day 0**. WP9's structural gate must confirm no write path to weight files exists for the duration — checked before the clock starts, not after.

## Daily cycle

1. Run the full physical/latent/memory loop over a fixed batch of tasks from a **held-out task set never seen during any prior day's mutation testing** — otherwise the system overfits the evaluation set and the day-30 numbers mean nothing.
2. Log per task: latency, task success (binary plus graded where applicable), sensing accuracy (RuView/mmWave ground-truth comparison where available), token/compute cost.
3. Each day's proposed mutations run: Darwin propose → `research-gate` evaluate → proof-gate check → constitutional gate if capability-expanding → promote or reject. **Every step witness-logged.**
4. Nightly Cloud Scheduler sweeps: causal-audit re-runs (WP6) and witness-chain integrity verification (WP8) against that day's promoted mutations.

## Day-30 evaluation criteria

- [ ] **Statistical comparison, not mean comparison.** Day-30 vs. day-1 per metric (latency, task success, sensing accuracy, cost) via paired bootstrap or equivalent, against a **pre-registered significance threshold and minimum effect size**. This is the guard against the exact "gains don't prove causation" failure mode that [arXiv:2607.26773](https://arxiv.org/abs/2607.26773) (2026-07-29) documents.
- [ ] **Full mutation provenance.** Every promoted mutation across all 30 days is traceable through the witness chain from proposal → evaluation → promotion, queryable end-to-end in **one traversal** — which is what the WP8 shared record schema exists to make possible.
- [ ] **Zero unapproved capability expansion.** Diff the capability/tool/action tables at day 0 vs. day 30. Any expansion without a matching WP11 constitutional-gate approval record **fails the acceptance test outright, independent of the performance numbers.**
- [ ] **Frozen-weights verification.** Re-hash foundation-model weights at day 30; must be bit-identical to day 0.
[...]

Issue #842, opened 2026-08-20
[PIR][WP10] Reversible-action verification stage (WorldCycle pattern, arXiv:2608.04964)


ruvnet:
Epic: #837 · Plan: `docs/research/perpetual-intelligence-runtime/03-program-plan.md`

## Goal

Add reversible-action-sequence verification between "observe consequences" and the promotion-gate evaluation in the physical loop.

The mechanism: an action sequence composed with its inverse should return the world to the initial state. Divergence from that closed cycle is an **annotation-free supervision signal** — no human labels required. This is how the program satisfies **invariant 6** ("every physical action produces new evidence") with evidence that is self-verifying rather than self-reported.

## Evidence

[arXiv:2608.04964](https://arxiv.org/abs/2608.04964) — "WorldCycle: Self-Verifiable Reinforcement Learning for Long-Horizon Video World Models," submitted 2026-08-05. Grade A, exact numeric match. The abstract states verbatim: *"WorldCycle reduces state returning drift by up to 44% and lifts composite-action accuracy nearly 4x over the base model."* A diagnostic benchmark, CycleBench, ships alongside it. No name collision.

## Acceptance criteria

- [ ] **A CycleBench-equivalent internal benchmark shows the target improvement over a no-WorldCycle baseline** — up to 44% long-horizon drift reduction and ~4x composite-action accuracy as the internal acceptance bar. **Report our own measured numbers; never restate the paper's figures as if they were ours.**
- [ ] A reversible-action-cycle verifier runs on every physical action sequence before its results reach the promotion gate.
- [ ] Cycle-closure error is computed and recorded per action sequence, and is available as an input signal to the promotion verdict.
- [ ] Verification results are witness-logged so the acceptance test can attribute drift reduction to this stage specifically.
- [ ] Sequences whose inverse is not well-defined are handled explicitly — either excluded with a recorded reason, or verified by a documented alternative. Never silently passed.
- [ ] Proposed crate: `ruvector-worldcycle-verify` under `crates/`.

## Dependencies

**WP9** (#841) — this stage sits inside that loop. Feeds **WP2**'s gate as an evaluation input. Blocks **WP12** (#843).

PR #869, opened 2026-08-20
feat(sota-harness): SHAPER-pattern frozen-weight evolution loop skeleton (PIR WP9, ADR-313)

ruvnet:
First shippable slice of **PIR WP9** (#841, epic #837): the structurally-enforced SHAPER-pattern loop skeleton per **ADR-313** (arXiv:2608.11350 — frozen model serves as both planner and optimizer; only skills, context, and the execution harness evolve).

### 1. Structural frozen-weights enforcement (`scripts/frozen-weights-check.mjs`)
ADR-313's central constraint, enforced **structurally, not by policy**: a zero-dependency, CI-runnable gate that fails the build if any file under the evolution loop's mutation surfaces (`crates/ruvector-sota-bench/harness/src`, `examples/mragent`, `crates/sona/src/darwin_guard.rs`) imports/invokes a fine-tuning or weight-writing API — or references a model weight file at all.

**Deny-list (9 entries, each documented in-script with why + in-repo anchor):**
| id | denies | anchor |
|---|---|---|
| `ruvllm-training-module` | `ruvllm::training` | `real_trainer.rs` (`RealTrainer::train`), `grpo.rs`, `contrastive.rs`, `mcp_tools.rs` |
| `ruvllm-qat-module` | `ruvllm::qat`, `lora_qat`, `training_loop` | `qat/{training_loop,lora_qat}.rs` |
| `ruvllm-lora-module` | `ruvllm::lora`, `micro(_)lora` | `lora/{training.rs,adapters/trainer.rs,micro_lora.rs}` — adapters are weight deltas |
| `training-entry-points` | `train_step`, `train_epoch`, `train_on_trajectories`, `train_buffered`, `RealTrainer` | grep of `pub fn train*` over ruvllm |
| `generic-finetune` | any `fine[-_]?tune` token | ADR-313 Decision §2 |
| `pretrain-pipelines` | `ruvltra_pretrain`, `pretrain_pipeline` | `sona/ruvltra_pretrain.rs`, `claude_flow/pretrain_pipeline.rs` |
| `weight-writing` | `save_checkpoint`, `save_adapter`, `save_weights`, `export_weights` | qat/lora/training writers |
| `model-file-reference` | `.gguf`, `.safetensors` | mutation surfaces may not name model files (weights path simply absent) |
| `mcp-weight-mutation-tools` | `ruvllm_microlora_{create,adapt}`, `ruvllm_sona_{create,adapt}` | TS-side MCP weight mutation |

Hardening matches the merged `adr-index.mjs`/`workspace-check.mjs` lessons (PR #857): symlinks never followed, realpath containment, per-entry error wrapping, and **a missing surface directory is a hard failure** (renaming a dir can't silently disable the gate). Comments are stripped before matching so `darwin_guard.rs`'s train/eval-contamination *prose* doesn't false-positive. `--self-test` builds adversarial positive/negative fixtures (violations per category, comment-only mentions, out-of-tree symlinked violations, broken symlink, symlink loop, deleted surface). Wired into CI as its own `frozen-weights` job.

### 2. Loop skeleton (`harness/src/shaperLoop.ts`)
One generation: **propose** (reusing Darwin's existing `@metaharness/darwin` 0.9.x surfaces — `gepaHarnessProposer` wraps `runRuvectorGepa`, nothing reinvented) → **evaluate** through the existing WP2 `dreamMachine.ts` adapter → **verdict** → **promotion recommendation** via the existing `vetoesFromVerdict`/vetoes/flywheel path. The same `FrozenModelRef` instance is handed to both the planner and optimizer roles (SHAPER's structural claim, configuration-tested), its loop-start sha256 is embedded in the witness-stamped report, and no weight-file path exists in any option or result type. Constitutional boundary identical to WP2's: recommendations only, no promote/merge export, frozen data result — test-asserted.

### 3. Genome types (`harness/src/genome.ts`)
ADR-313 §1's enumeration of what MAY evolve as a **closed type union** (`skills | context | harness` — a weights surface is unrepresentable; runtime guard throws for untyped callers). Capability-expansion boundary per **ADR-315**: any candidate adding tools/action classes/communication peers routes through `constitutionalGateStub`, which models autogenous `promoteAuthorized`'s `authorized` conjunct and **blocks by default** — no approval record can exist until WP11 wires PIR's capability tables in (documented as the WP11 integration point).

### 4. Tests
9 new under the harness's `node --test` setup; full suite **26/26 passing** (17 existing + 9 new):
- full synthetic generation (mutate → evaluate → verdict → recommendation, ACCEPT path)
- regressed candidate blocked via `dream_machine_reject`
- capability-expanding candidate routed to the constitutional stub and BLOCKED (evaluator provably never runs)
- weights-surface and malformed-model-hash rejection
- no-promotion-export + frozen-result boundary assertion (same contract as WP2's)
- GEPA proposer reuse of Darwin's search (injected benchmark)
- frozen-weights check: real-tree run clean + `--self-test` green

## What this slice does NOT include (honest scoping)
- **No real environment rollouts** — evaluation here is the WP2 statistical/dream-machine path over injected samples; VLABench/ESI-Bench-style skill-improvement evidence is future WP9 work.
- **No WorldCycle verification stage** — that is WP10 (#842).
- **No 30-day acceptance harness / day-0-vs-day-30 re-hash** — that is WP12 (#843); this slice records the loop-start hash in the witness-stamped report but does not run the longitudinal check.
- **No live ruvllm mutator end-to-end** — Darwin's local mutator backend (ADR-259) is exercised only through the injected-benchmark GEPA path here; live-serve e2e comes after WP0b's fix is validated in that flow.
- **The ADR-315 gate is a stub** — deliberately blocks everything; real autogenous `promoteAuthorized` integration is WP11.

Refs #841, #837. ADRs: 313, 315, 306.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,25 @@ jobs:
       - name: Check crates/ for orphan crates (PIR #859)
         run: node scripts/workspace-check.mjs
 
+  frozen-weights:
+    name: Frozen-weights structural gate
+    runs-on: ubuntu-22.04
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
+        with:
+          submodules: recursive
+
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020
+        with:
+          node-version: '20'
+
+      - name: Self-test the gate (PIR #841, ADR-313)
+        run: node scripts/frozen-weights-check.mjs --self-test
+
+      - name: Scan evolution-loop mutation surfaces for training/weight APIs
+        run: node scripts/frozen-weights-check.mjs
+
   fmt:
     name: Rustfmt
     runs-on: ubuntu-22.04
--- a/crates/ruvector-sota-bench/harness/src/darwin.ts
+++ b/crates/ruvector-sota-bench/harness/src/darwin.ts
@@ -76,7 +76,7 @@ function reflectParameter(prompt: string): string {
   return `${target}:value=${next}`;
 }
 
-function policyFromGenome(genome: Genome): Record<string, string> {
+export function policyFromGenome(genome: Genome): Record<string, string> {
   return Object.fromEntries(Object.entries(genome.components).map(([name, encoded]) => {
     const prefix = `${name}:value=`;
     if (!encoded.startsWith(prefix)) throw new Error(`invalid Darwin component encoding for ${name}`);
--- a/crates/ruvector-sota-bench/harness/src/genome.ts
+++ b/crates/ruvector-sota-bench/harness/src/genome.ts
@@ -1,0 +1,158 @@
+/**
+ * Skill / context / harness genome types (PIR WP9, ADR-313, issue #841).
+ *
+ * ADR-313 §1 enumerates exactly what MAY evolve in the SHAPER-pattern loop:
+ * skills, context, and the execution harness. Weights are NOT a mutation
+ * surface — `MutationSurface` has no "weights" member, so a weight-mutating
+ * candidate is unrepresentable in this type system. The structural backstop
+ * for JS callers (and for this module's own source) is
+ * scripts/frozen-weights-check.mjs, which scans this directory as an enforced
+ * mutation surface.
+ *
+ * CAPABILITY-EXPANSION BOUNDARY (ADR-315): a mutation that adds tools, action
+ * classes, or communication peers requires the constitutional gate — a
+ * separately-authorized approval DISTINCT from ordinary behavioral promotion.
+ * That gate is stubbed here against autogenous's `promoteAuthorized` concept
+ * (ADR-315 Decision §2: `Promote = Better ∧ Safe ∧ Authorized ∧ Reversible`)
+ * and BLOCKS by default: until WP11 wires PIR's capability/tool/action tables
+ * into the `authorized` conjunct, no approval record can exist, so no
+ * capability expansion can be recommended. WP11 integration point.
+ */
+
+/**
+ * What may evolve (ADR-313 §1). Deliberately closed: adding a member to this
+ * union is an ADR-level decision, not a code change.
+ */
+export type MutationSurface = "skills" | "context" | "harness";
+
+/** A reusable skill the frozen model can invoke — SHAPER's evolving skill set. */
+export interface SkillDefinition {
+  readonly name: string;
+  /** The skill body (prompt/program text the frozen model executes against). */
+  readonly body: string;
+}
+
+/** The evolving skill library. */
+export interface SkillGenome {
+  readonly surface: "skills";
+  readonly version: number;
+  readonly skills: readonly SkillDefinition[];
+}
+
+/** The evolving context/configuration handed to the frozen model. */
+export interface ContextGenome {
+  readonly surface: "context";
+  readonly version: number;
+  readonly entries: Readonly<Record<string, string>>;
+}
+
+/**
+ * The evolving execution-harness parameters. Maps onto Darwin's existing
+ * `Policy` surface (flywheel.ts PARAMETER_CANDIDATES / darwin.ts GEPA).
+ */
+export interface HarnessGenome {
+  readonly surface: "harness";
+  readonly version: number;
+  readonly parameters: Readonly<Record<string, string>>;
+}
+
+export type EvolvableGenome = SkillGenome | ContextGenome | HarnessGenome;
+
+const SURFACES: readonly MutationSurface[] = ["skills", "context", "harness"];
+
+/**
+ * Runtime guard for untyped callers: the only admissible surfaces are
+ * ADR-313's three. Anything else — "weights" above all — throws.
+ */
+export function assertEvolvableGenome(genome: EvolvableGenome): void {
+  if (!SURFACES.includes(genome.surface)) {
+    throw new Error(
+      `inadmissible mutation surface "${String(genome.surface)}" — ` +
+      `ADR-313 permits only ${SURFACES.join(", ")}; weights never evolve`,
+    );
+  }
+  if (!Number.isInteger(genome.version) || genome.version < 0) {
+    throw new Error("genome.version must be a non-negative integer (surfaces are separately versioned)");
+  }
+}
+
+/**
+ * A reference to the frozen foundation model. The sha256 is the loop-start
+ * hash recorded for the witness chain (ADR-313: day-30 re-hash must be
+ * bit-identical to day 0). There is deliberately no weight-file path here.
+ */
+export interface FrozenModelRef {
+  readonly id: string;
+  readonly sha256: string;
+}
+
+export function assertFrozenModelRef(model: FrozenModelRef): void {
+  if (!/^[0-9a-f]{64}$/.test(model.sha256)) {
+    throw new Error("frozen model sha256 must be 64 lowercase hex chars (loop-start hash, witness-recorded)");
+  }
+  if (!model.id) throw new Error("frozen model ref requires an id");
+}
+
+/**
+ * What a candidate mutation would ADD to the agent's capability set.
+ * Any non-empty field crosses the ADR-315 constitutional boundary.
+ */
+export interface CapabilityDelta {
+  readonly addsTools: readonly string[];
+  readonly addsActionClasses: readonly string[];
+  readonly addsCommunicationPeers: readonly string[];
+}
+
+/** A proposed mutation: parent genome, candidate genome, and its capability delta. */
+export interface CandidateMutation {
+  readonly parent: EvolvableGenome;
+  readonly candidate: EvolvableGenome;
+  /** One-line human-auditable description of the mutation. */
+  readonly describe: string;
+  /** Omitted or all-empty means: no capability expansion. */
+  readonly capabilityDelta?: CapabilityDelta;
+}
+
+export function expandsCapabilities(delta?: CapabilityDelta): boolean {
+  if (!delta) return false;
+  return delta.addsTools.length > 0
+    || delta.addsActionClasses.length > 0
+    || delta.addsCommunicationPeers.length > 0;
+}
+
+/**
+ * The constitutional gate's decision shape, mirroring the conjuncts of
+ * autogenous `promoteAuthorized(candidate, champion, { authorized, reversible })`
+ * (ADR-315 Decision §2 — implemented and DONE upstream).
+ */
+export interface ConstitutionalDecision {
+  readonly allowed: boolean;
+  readonly reasons: readonly string[];
+}
+
+/**
+ * ADR-315 constitutional gate — STUB (WP11 integration point).
+ *
+ * Models the `authorized` conjunct of autogenous's `promoteAuthorized`
+ * predicate. This program's residual scope (ADR-315 Decision §2a/2b) is to
+ * wire PIR's capability/tool/action tables into that conjunct and to enforce
+ * the proposer≠approver distinctness rule. None of that wiring exists yet, so
+ * NO approval record can exist, so the gate blocks — capability expansion is
+ * denied by default, never rounded up. WP11 replaces this stub with the real
+ * autogenous integration; the test suite pins the blocked-by-default contract.
+ */
+export function constitutionalGateStub(mutation: CandidateMutation): ConstitutionalDecision {
+  if (!expandsCapabilities(mutation.capabilityDelta)) {
+    return Object.freeze({
+      allowed: false,
+      reasons: Object.freeze(["not_a_capability_expansion — this gate only adjudicates capability expansion (ADR-315 §1)"]),
+    });
+  }
+  return Object.freeze({
+    allowed: false,
+    reasons: Object.freeze([
+      "constitutional_gate_stub_unwired — WP11 has not wired PIR capability tables into promoteAuthorized's authorized conjunct",
+      "no_approval_record — ADR-315 requires an explicit, witness-anchored approval distinct from behavioral promotion",
+    ]),
+  });
+}
--- a/crates/ruvector-sota-bench/harness/src/shaperLoop.ts
+++ b/crates/ruvector-sota-bench/harness/src/shaperLoop.ts
@@ -27,6 +27,19 @@
  * test/shaperLoop.test.ts asserts both. A capability-expanding candidate is
  * additionally routed through the ADR-315 constitutional gate (stubbed in
  * genome.ts, WP11 integration point) and is BLOCKED by default.
+ *
+ * ACCEPTED RESIDUALS (PR #869 security audit, on record — not closed here):
+ *   - The capability gate fires on the proposer's SELF-DECLARED
+ *     `capabilityDelta`. A mutation that adds a tool via its skill body while
+ *     declaring no delta would skip the gate and reach evaluation — but not
+ *     promotion: the human + veto/flywheel path still own that. Independent
+ *     extraction of the capability delta from genome content is exactly the
+ *     WP11 wiring named in genome.ts.
+ *   - The model-hash binding is RECORD-ONLY in this slice: sha256 is
+ *     validated as 64-hex and witness-stamped into the report (tamper-evident
+ *     about which hash was claimed), but nothing here recomputes it from real
+ *     weights. The day-0/day-30 re-hash that ENFORCES the freeze is WP12's
+ *     30-day acceptance harness.
  */
 import { runRuvectorGepa, policyFromGenome, type RuvectorGepaOptions } from "./darwin.js";
 import {
--- a/crates/ruvector-sota-bench/harness/src/shaperLoop.ts
+++ b/crates/ruvector-sota-bench/harness/src/shaperLoop.ts
@@ -1,0 +1,193 @@
+/**
+ * SHAPER-pattern frozen-weight evolution loop — one-generation skeleton
+ * (PIR WP9, ADR-313, issue #841; SHAPER = arXiv:2608.11350).
+ *
+ * Orchestrates ONE generation:
+ *
+ *   propose (Darwin/GEPA surfaces) ──► evaluate (dreamMachine.ts adapter)
+ *        ──► verdict ──► promotion RECOMMENDATION (vetoes/flywheel path)
+ *
+ * FROZEN WEIGHTS ARE STRUCTURAL HERE, not policy:
+ *   - The model-weights path is simply absent: the loop takes a
+ *     `FrozenModelRef` (id + loop-start sha256, witness-recorded via the
+ *     dream-machine report stamp) and no weight-file path exists anywhere in
+ *     the option or result types. `MutationSurface` (genome.ts) has no
+ *     "weights" member, so a weight-mutating candidate is unrepresentable.
+ *   - The SAME FrozenModelRef instance is handed to both the proposer
+ *     (planner role) and the evaluator (optimizer role) — SHAPER's structural
+ *     claim that one frozen model serves as planner and optimizer. The result
+ *     records both roles' model identity for the configuration test.
+ *   - This file lives inside a mutation-surface directory enforced by
+ *     scripts/frozen-weights-check.mjs: importing a training / weight-writing
+ *     API here fails CI.
+ *
+ * CONSTITUTIONAL BOUNDARY — identical to WP2's (dreamMachine.ts, ADR-306):
+ * the loop emits recommendations only. It exports no promote/merge function,
+ * the result is frozen data with no callable members, and
+ * test/shaperLoop.test.ts asserts both. A capability-expanding candidate is
+ * additionally routed through the ADR-315 constitutional gate (stubbed in
+ * genome.ts, WP11 integration point) and is BLOCKED by default.
+ */
+import { runRuvectorGepa, policyFromGenome, type RuvectorGepaOptions } from "./darwin.js";
+import {
+  evaluateCandidate,
+  vetoesFromVerdict,
+  type DreamMachineVerdict,
+} from "./dreamMachine.js";
+import { pairedBootstrapDecision } from "./statistics.js";
+import {
+  assertEvolvableGenome,
+  assertFrozenModelRef,
+  constitutionalGateStub,
+  expandsCapabilities,
+  type CandidateMutation,
+  type ConstitutionalDecision,
+  type EvolvableGenome,
+  type FrozenModelRef,
+} from "./genome.js";
+
+/** Propose one candidate mutation. Receives the frozen model in its PLANNER role. */
+export type CandidateProposer = (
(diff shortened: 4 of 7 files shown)


ruvnet:
## Security audit — Phase 4 pre-merge review (WP9 SHAPER loop)

Audited the four surfaces at branch head. **Verdict: FINDINGS — 1 MEDIUM (plus 2 low/informational).** Ran `frozen-weights-check.mjs --self-test` (14/14 pass), the real-tree scan (25 files, 0 hits — clean today), and probed the deny-matcher and the TS contracts directly. The CI job, `darwin.ts`, `genome.ts`, and `shaperLoop.ts` are clean; the finding is a correctness gap in the gate's comment-stripping that weakens it in the false-negative direction.

### MEDIUM — `stripComments` eats denied tokens that follow `//` inside a string literal
`scripts/frozen-weights-check.mjs` (`stripComments`, applied in `scanFile`)

The stripper runs `source.replace(/\/\/[^\n]*/g, ' ')` with no awareness of string context, so a `//` **inside a quoted string** is treated as a line comment and everything after it on that line is deleted before matching. Any denied token sitting after a `//` — i.e. the most natural way to name a remote model file or a training endpoint, a URL — is silently dropped. Reproduced directly against the shipped regex:

```
"const w = \"https://hf.co/repo/model.safetensors\";"
   stripped = 'const w = "https: '     model-file flagged: FALSE   (.safetensors gone)
"loadWeights(\"file://models/x.gguf\");"
   stripped = 'loadWeights("file: '    model-file flagged: FALSE   (.gguf gone)
"const u = \"http://internal/finetune\";"
   stripped = 'const u = "http: '      finetune  flagged: FALSE
"const w = \"models/llama.gguf\";"      model-file flagged: TRUE    (no // → still caught)
"import { RealTrainer } from \"ruvllm/training\";"  training flagged: TRUE  (single / → still caught)
```

So the `model-file-reference` and `generic-finetune` deny rules (2 of the 9) are defeated for any URL-form reference, while a plain relative path is still caught. This is a **false negative** — the dangerous direction for a gate — and it exceeds the script's own documented "does not catch a sufficiently obfuscated or dynamically-assembled path" limitation: a bare `https://…/model.safetensors` string is ordinary code, neither obfuscated nor dynamic.

**Impact is bounded** and this is not a merge-blocker for correctness of what exists today (the real-tree scan is clean, and the training-import rules — `ruvllm::training`, `RealTrainer`, `save_checkpoint`, `train_step`, the MCP tool names — contain no `//` and remain intact). But this is the program's centerpiece structural gate, so the model-file rule should actually bite on the natural remote-reference form before the loop is relied on.

**Fix:** make comment removal string-aware. Either (a) a small character-scan that tracks `"`/`'`/`` ` `` string state and only strips `//` and `/* */` when *not* inside a string (preserves `.safetensors` in a URL, still strips genuine prose), or (b) match the deny-list against the raw source too and take the union — but (a) is cleaner since (b) reintroduces the prose false-positives (`darwin_guard.rs` "train/eval") that stripping exists to avoid.

### LOW / informational (no fix required to merge)
- **Scan coverage** (`SCAN_EXT` / `SKIP_DIRS`): the scan only reads `.ts/.mts/.cts/.js/.mjs/.cjs/.rs` and skips `fixtures`, `dist`, `node_modules`, `target`, `pkg`. A mutation surface that reached a training API via a `.py`/`.sh` helper, or hid a weight-writing `.mjs` under `fixtures/` or a committed `dist/`, would be missed. This is the documented "static check / move-code-to-dodge" scoping boundary (the gate makes the honest path loud, not every dishonest path impossible) — worth stating explicitly since `dist/` is the form that actually executes.
- **Capability gate depends on a self-declared delta** (`shaperLoop.ts` → `expandsCapabilities(mutation.capabilityDelta)`): the constitutional gate only fires when the candidate *declares* a `capabilityDelta`. A mutation that adds a tool via its skill body while declaring no delta would skip the gate and reach evaluation. Independent extraction of the capability delta from genome content is exactly the WP11 wiring named in `genome.ts`; noting it so the self-declared-delta assumption is on record. (Reaching evaluation is not reaching promotion — a human + the veto/flywheel path still own that.)
- **Model-hash binding is record-only:** `FrozenModelRef.sha256` is validated as 64-hex and witness-stamped into the dream-machine report, but nothing in this slice recomputes the hash from the actual model to *enforce* the freeze — the day-0/day-30 re-hash ADR-313 describes is operational and out of this skeleton's scope. The binding is tamper-evident about *which* hash was claimed, not that it matches real weights.

### Checked and clean
- **CI job** (`.github/workflows/ci.yml` `frozen-weights`): inherits the workflow's plain `pull_request` trigger (not `pull_request_target`), references no `secrets.*`, pins both actions to full 40-char commit SHAs, and its two `run:` steps are static `node scripts/frozen-weights-check.mjs [--self-test]` with no `${{ }}` interpolation. No injection or secret-exposure surface.
- **Guard-script hardening** inherits the PR #857 lessons and holds: symlinks (file and directory) are never followed and are skipped with a warning; every kept entry is realpath-contained under its surface; per-entry try/catch means broken symlinks / cycles warn-and-skip with no stack-trace abort; and a **missing or symlinked surface directory is a hard fail** (`lstatSync` → treated as missing → exit 1), so renaming a surface can't silently disable the gate. `--self-test` exercises all of this (14/14).
- **`genome.ts`:** `MutationSurface` is a closed union with no `weights` member, and `assertEvolvableGenome` is a runtime backstop that throws on any surface outside `skills|context|harness` — a deserialized `{surface:"weights"}` is rejected (test-covered). `constitutionalGateStub` blocks by default (frozen `allowed:false`), never rounds up.
- **`shaperLoop.ts`:** the capability gate is ordered **before** evaluation in code, not merely tested — a capability-expanding candidate sets `blockedByGate`, `candidateSamples` short-circuits to `[]`, and `options.evaluate` is never invoked (the test asserts `evaluated === false`). The result is `Object.freeze`d frozen data; no `promote`/`merge`/`apply` export exists (asserted by regex over exports + `result.promote === undefined`). Same frozen instance serves planner and optimizer.
- **`darwin.ts`:** the sole change is `policyFromGenome` gaining `export` — visibility only, no behavior change.

ruvnet:
## Audit response — finding → fix mapping (commit 677490aae)

**MEDIUM — `stripComments` eats denied tokens after `//` inside a string → FIXED.**
Replaced the regex strip with a per-language character scan that tracks quote state and strips `//` and `/* */` (and `#` for py/sh) **only outside strings**:
- js-like: `'` / `"` / template literals, including `${}` re-entry with counted brace nesting and backslash escapes;
- rust: `"` only — `'` is deliberately not a string delimiter (lifetimes like `&'a str` would desynchronize the scan; a char literal cannot hide a multi-char token), plus Rust's nested block comments;
- python: `#` comments, `'`/`"` and triple-quoted strings;
- shell: `#` only at line start/after whitespace (so `$#`/`${#x}` survive), `'` without escapes, `"` with.

All three of your reproduction fixtures are now self-test cases and **flag**: `"https://hf.co/repo/model.safetensors"` → `model-file-reference`, `"file://models/x.gguf"` → `model-file-reference`, `"http://internal/finetune"` → `generic-finetune`. Your regression case `import { t } from "ruvllm/training"` (single slash) **still flags**, and two new negative cases — a comment containing `http://internal/finetune` and a `.sh` comment mentioning `fine_tune` — are asserted **NOT** flagged. Residual stripper edges (a JS regex literal containing `//`; a Rust char-literal `"` or raw string) are documented in the function doc and are all in the false-positive direction — they can over-strip prose but can never hide a token in code or a string.

**LOW (scan coverage) → HARDENED.** `SCAN_EXT` now includes `.py` and `.sh`; `dist` is removed from `SKIP_DIRS` (committed build output is what executes). Honest result of the widened real-tree scan: **25 files across the 3 surfaces, 0 hits** — no committed `dist/` or vendored code exists under the surfaces today, so no scoping carve-out was needed; if one ever appears it will be scanned like source. Self-test now covers a `.py` weight-writer call, a `.sh` finetune invocation, and a violation under `dist/` (all flag). `fixtures/` and `node_modules/` remain skipped by convention, stated explicitly in the header.

**LOW (self-declared capabilityDelta) → DOCUMENTED as accepted residual** in both the script header and the `shaperLoop.ts` module docs: the gate fires on the proposer's self-declared delta until WP11 wires independent delta extraction from genome content; reaching evaluation is not reaching promotion — the human + veto/flywheel path still own that.

**LOW (model-hash record-only) → DOCUMENTED as accepted residual** in the same two places: `FrozenModelRef.sha256` is validated and witness-stamped (tamper-evident about the claimed hash), but the day-0/day-30 re-hash that enforces the freeze is WP12's operational harness.

**Verification at branch head:** `--self-test` 23/23 assertions pass; real-tree scan `OK — 25 files … 0 training/weight-API references`; harness `npm test` 26/26 green.

Ready for re-verification.

ruvnet:
## Security re-verification — fix `677490aae`

The original MEDIUM (comment-strip eating denied tokens after `//` in a string) is **genuinely fixed and verified**. But hunting specifically for false-negatives as requested, I produced two against the new scanner — both in the **JS regex-literal path that the fix documents as "false-positive-direction only."** That residual claim is incorrect. **Verdict: REMAINING ISSUE (1 residual false-negative, MEDIUM-narrow) — the original finding is closed, but the documented residual needs to be either fixed or re-classified before this is relied on.**

### Confirmed fixed
- The three original URL-form fixtures now flag: `"https://hf.co/repo/model.safetensors"` → `.safetensors`, `"file://models/x.gguf"` → `.gguf`, `"http://internal/finetune"` → `finetune`. Comment-only prose (`// … model.safetensors …`) is still correctly NOT flagged. The char scanner keeps string contents while dropping true comments — correct.
- `--self-test` passes (23/23), real-tree scan clean (25 files, 0 hits with `.py`/`.sh` now in `SCAN_EXT` and `dist` removed from `SKIP_DIRS`).
- The two low residuals are on record in `shaperLoop.ts` (self-declared `capabilityDelta`; record-only model-hash → WP12). Good.

### REMAINING — MEDIUM (narrow): JS regex literal containing `//` strips a trailing denied token
The scanner has no regex-literal awareness, so inside a JS regex a `//` that is **not** a comment in real JS is misread as a line comment, stripping the rest of the line. Two runnable PoCs, both flagged by the deny-list before this fix's semantics and **silently passed now**:

```js
// (a) '//' inside a regex character class — class contains '/', regex continues
const re = /[a//]/; const w = "evil-model.gguf"; export {re,w};
// (b) escaped slashes adjacent to the closing slash — a normal URL-matching regex
const re = /https:\/\//; loadWeights("m.safetensors");
```
Both are **live code** in real JS (confirmed with node): in (a) `re` is a RegExp and `w === "evil-model.gguf"`; in (b) the regex matches `https://` and `loadWeights` actually runs with `"m.safetensors"`. Yet `frozen-weights-check.mjs` reports **0 hits** for both — the `.gguf`/`.safetensors` after the regex `//` is comment-stripped. So the `model-file-reference` and `generic-finetune` rules are defeated for any denied token sharing a line with a regex literal that contains `//`.

This is the same class as the original finding (silent miss → false-negative, the dangerous direction), and it directly contradicts the header's "residuals … false-positive-direction only." Form (b) is not exotic: `if (/https?:\/\//.test(u)) loadModel("x.gguf")` is ordinary code, so the accidental-miss path is real, not just a crafted evasion.

Rust raw strings (`r#"…/model.safetensors"#`), template literals with `${}`, and division-then-`//` were all tested and correctly flag (the raw-string / template desyncs land in string-content, which is preserved → over-flag, the safe direction). So the false-negative is specific to **JS regex literals**, not the other documented residual.

**Fix options:**
1. Add minimal regex-literal recognition to the JS branch: when a `/` appears in value position (after `=`, `(`, `,`, `return`, `[`, an operator, or line start), consume to the matching unescaped `/` — respecting `[...]` character classes where `/` is literal — as a string-like (content-preserving) token, so its interior is never treated as a comment. This closes both PoCs.
2. If full regex lexing is judged too costly for a skeleton gate, at minimum **re-document the residual honestly** as a KNOWN false-negative (a regex literal containing `//` on the same line as a denied token), not "false-positive-direction only" — otherwise a reviewer trusting the header will not scrutinize regex-adjacent lines, which is exactly where the miss hides.

Given the gate is defense-in-depth (human + veto/flywheel own promotion) and the real tree is clean, this is not a hard blocker if you choose to merge and fast-follow — but the "false-positive-only" claim must be corrected either way, and option 1 is a small, well-bounded change I'd prefer before the loop is load-bearing.

ruvnet:
## Re-verify response — residual JS-regex false-negative FIXED (commit ae7d875cb)

You were right on both counts: it was a genuine false-negative, and the "false-positive-direction only" header claim was wrong for the JS regex path. Fixed the residual, corrected the claim.

**Fix — regex-literal awareness in the JS branch.** A single `/` that is not `//` or `/*` (both stay *always* comments — an empty `//` regex is illegal JS, so comment recognition still wins first) now opens a **regex literal** when it appears in **value position**; in **operator position** it is division. Value vs. operator is tracked by the last significant token:
- operand → division: identifier/number, `)`, `]`, or a just-closed string/regex/template;
- value → regex: `=` `(` `,` `[` `{` `;` `:`, operators, line start, and value-keywords (`return`, `typeof`, `new`, `throw`, `yield`, `await`, `case`, …).

A regex body — including any internal `//` and `[...]` char classes (where `/` is literal) — is preserved as content and never comment-stripped, then trailing flags are consumed. When the value/operator guess is uncertain the scan **prefers the regex reading**, which preserves content — the false-POSITIVE (safe) direction.

**Both your PoCs now flag** (added as self-test cases):
- `const re = /[a//]/; const w = "evil-model.gguf";` → `.gguf` flagged (`model-file-reference`)
- `const re = /https:\/\//; loadWeights("m.safetensors");` → `.safetensors` flagged
- plus the ordinary-code form `if (/https?:\/\//.test(u)) loadModel("x.gguf")` → `.gguf` flagged

**Negative cases added** (must NOT flag): a division-then-real-comment line (`a / b / c; // … models/x.gguf` — division must not derail comment stripping) and a regex-then-real-comment line (`/ab+c/; // … model.safetensors` — the closing `/` ends the regex, the `//` after is a real comment). Both correctly leave the comment token stripped/unflagged.

**Corrected the header:** the residual note now states the JS-regex false-negative is *closed*, and the only remaining stripper edges (Rust char-literal `"` / raw strings; a division mis-guessed as regex) are genuinely over-flag/safe-direction — a reviewer trusting the header will not be misled onto regex-adjacent lines.

**Anticipating more regex-evasion attempts**, I stress-probed 11 further variants beyond the PoCs — regex after `return`/`,`/`[`/`(`, in a ternary branch, with `gi` flags, nested-class slashes `/[/a-z//]+/`, paren-division `(a) / b / c`, and assign-division chains `a/2/2`. In every case the denied token in code/strings is preserved (flagged); only denied tokens inside genuine trailing comments are stripped. Happy to have you push on it further.

**Verification at branch head (ae7d875cb):** `--self-test` 27/27 assertions pass (incl. both PoCs + the ordinary-code guard + division/regex negative cases); real-tree scan `OK — 25 files … 0 hits`; harness `npm test` 26/26 green.

Ready for re-verification.

ruvnet:
## Security re-verification — fix `ae7d875cb`

The position heuristic **closes the original false-negatives** — the URL-string and escaped-slash-regex forms from the last round now flag correctly (e.g. `if (cond) /a\/\//.test("x"); loadWeights("real.safetensors")` is caught). `--self-test` 27/27, real-tree 0 hits. But you asked me to keep hunting, and I found **one narrow residual false-negative** in the `)`/`]`→operand rule. **Verdict: REMAINING ISSUE (LOW, contrived) — your call whether to block or accept-with-honest-doc; the "closes the JS false-negative" claim isn't 100%.**

### The residual: statement-position regex after `)` + `//` inside a char class
`)` and `]` unconditionally set `expectOperand = true`. But after a control-flow header — `if (…)`, `while (…)`, `for (…)` — the `)` is followed by a **statement**, where `/` opens a regex, not division. The heuristic confidently guesses division, then a `//` inside the regex's character class trips the always-on `//`→line-comment rule and swallows the rest of the line — including a live denied token. Two runnable PoCs, both scan as **0 hits**, both confirmed live via node:

```js
// PoC 1 — loadWeights runs with "evil.gguf" (node-confirmed), gate reports 0 hits
function f(cond, s) { if (cond) /[//]x/.test(s); loadWeights("evil.gguf"); }
// PoC 2 — realistic-ish: URL-ish char class, model load via return on the same line
export function chk(u, s) { if (isUrl(u)) /[a-z//]/.test(s); return loadWeights("m.safetensors"); }
```
Stripped output for PoC 1 is `function f(cond,s){ if (cond) /[` — everything from the `//` in the char class onward is treated as a line comment, so the live `loadWeights("evil.gguf")` is dropped and `.gguf` is never seen.

**Root cause / why it's the confident-wrong case:** the heuristic's "uncertain → regex (safe)" fallback doesn't apply here — the `)`→operand rule is *certain* it's division, which is exactly wrong for `if/while/for (...)` statement position. It's the classic JS regex/division ambiguity: a `)` closing a control-flow header vs. closing a grouping/call are indistinguishable without tracking what the matching `(` opened.

**Severity — LOW / contrived:** the trigger needs a `//` literally inside a regex character class, in statement position after `)`, on the same line as a denied token. Accidental occurrence is very unlikely; as deliberate evasion it's possible but conspicuous (`/[//]/` next to a vanished line would read as suspicious in review), and the gate is defense-in-depth (human + veto own promotion). The training-import rules (`ruvllm::training`, `RealTrainer`, `save_checkpoint`, MCP tool names) contain no `//` and are unaffected; the exposure is the model-file/finetune rules for a same-line token after this specific construct.

**Fix options:** (1) handle the control-flow case — when a `)` closes a `if/while/for/switch/catch (…)` header, set `expectOperand = false` (needs tracking the matching `(`'s keyword); or (2) the pragmatic move: when the (mis)classified division region reaches a `[` before any `;`/newline, or a `//` while an unclosed `[` is open on the line, treat as regex content; or (3) accept it and correct the header note to name this as a KNOWN narrow false-negative rather than implying the JS path is fully closed. Given it's LOW and the realistic surface (strings/imports) is fully covered now, merge-and-fast-follow is defensible — but the residual should be documented honestly, not described as closed.

ruvnet:
## 3rd re-verify response — took the PRINCIPLED path (bias inversion), whole FN class closed (commit 654e603d4)

You're right that three progressively-narrower false-negatives means the approach was wrong, not the individual patches. A sound JS regex-vs-division classifier needs full control-flow paren-matching — a real JS lexer, out of scope for a shell gate. So I did **option 1 (bias inversion)**, not another PoC patch.

**The inversion.** Instead of strip-comments-then-match (which drops a real token whenever it misjudges comment context), the stripper now blanks out **only spans it is conservatively CERTAIN are comments**, and the deny-list matches everything else. A denied token is reported **unless it provably sits in a genuine comment**; any ambiguity — regex-, string-, or statement-position-adjacent — resolves to **FLAG**. Failure mode is over-flagging (safe), never under-flagging (dangerous).

**The sound rule needs no regex/division classification at all.** A JS `//` is treated as a comment **only when no code `/` precedes it on the line**. Rationale: a regex literal must open with a `/`, and division is a `/` too — so if any code `/` came first, the `//` could be regex-interior (`/[a//]/`) or division-adjacent, and we conservatively do **not** call it a comment. With zero preceding code `/`, a `//` cannot be inside a regex (nothing opened one) → it's a genuine comment. Rust `//` is always a comment (no regex literal, no `//` operator); strings (all forms, incl. URLs) and `/* */` are lexed as before. **Provably sound: suppression happens only for real comments, so no denied token in code/string/regex is ever dropped** — regardless of what precedes the regex (`if`/`while`/`for`/ternary/nested `)`), because the rule never inspects that context.

**Your 3rd PoC now flags** (added as a self-test case): `if (cond) /[//]x/.test(s); loadWeights("evil.gguf")` → `.gguf` flagged (`model-file-reference`). All prior PoCs still flag (URL strings, char-class/escaped/guard regexes).

**Honest about the trade** — two ACCEPTED over-flags are now on record as self-test cases and in the header: a denied token in a *real* trailing comment that shares its line with a code `/` (division or regex) **is flagged** (annoying, safe). A provable comment — no prior code `/` on the line, e.g. a leading or standalone `// … model.gguf` — is still correctly suppressed (negative case retained). Corrected the header: the JS path is now genuinely closed; the only remaining edges (regex-body tokens, Rust raw-string desync, comment-after-slash) are all over-flag/safe.

**Anticipating the next FN hunt**, I stress-ran 13 code/string variants — regex after `while`/`for`/ternary/nested-call `)`, division chains `a/2/2`, template and single-quote strings with `//`, a `//` inside a string literal — **zero false-negatives**; and 3 provable-comment cases suppress correctly.

**Verification at branch head (654e603d4):** `--self-test` 27/27 pass (all three PoCs + provable-comment negative + accepted-over-flag cases); real-tree scan `OK — 25 files … 0 hits`; harness `npm test` 26/26 green.

This is the version that converges: the gate no longer guesses, so there is no narrower case to find. Ready for re-verification.

ruvnet:
## Security re-verification — fix `654e603d4` — CLEAR-TO-MERGE

The bias-inversion converges. I could not produce a false-negative, and I can now argue there isn't one to find. **Verdict: CLEAR-TO-MERGE.**

### Why the rule is sound (not just "no FN found")
The stripper drops a `//` as a comment only when, in code mode, **no code `/` has appeared earlier on the line**. A false-negative requires a denied token in **live code** to be dropped — i.e. a `//` the stripper treats as a comment that JS does **not**. In JS the only non-comment `//` is:
- **inside a string** → the stripper is in string mode and never reaches the code-mode `//` branch, so the interior (and any denied token) is preserved; or
- **inside a regex** → a regex literal always opens with `/`, and that opening slash sets `sawCodeSlashThisLine`, so the later `//` is kept as code (not stripped) and the interior is preserved.

Every other `//` in code position with no preceding slash is a genuine comment in JS too, so stripping it matches JS semantics. The failure mode is therefore strictly **over-flagging** (a real comment kept because a division `/` preceded it on the line → its prose gets flagged), which is the safe direction. The gate no longer classifies regex-vs-division at all — it just refuses to call anything a comment once the line is ambiguous.

### Empirical confirmation
- All three prior FN PoCs now **flag**: the URL-string `"https://…/model.safetensors"`, `if (cond) /[//]x/.test(s); loadWeights("evil.gguf")`, and `if (isUrl(u)) /[a-z//]/.test(s); return loadWeights("m.safetensors")`.
- Fresh FN-hunt variants all flag: division-then-string-token, regex-then-string-token-same-line, regex-inside a template `${…}` expression, and two-strings-with-`//`-in-the-first. Zero false-negatives.
- The safe-direction check holds: `a / b // comment mentioning model.gguf` **over-flags** (the comment mention is kept) — acceptable per the design.
- `--self-test` 27/27.

That closes the whole JS-regex FN class. Nothing further gating from the security side — good to merge (and to unblock #870's rebase onto this).

Issue #865, opened 2026-08-20
[PIR][WP18] Cross-source causal-graph memory fusion layer (MemFuse pattern, ADR-320)


ruvnet:
Part of the **RuV Perpetual Intelligence Runtime (PIR), Wave 2**. Epic: #837. Plan: `docs/research/perpetual-intelligence-runtime/07-wave2-program-plan.md` · evidence: `06-wave2-evidence-review.md` (both land on `main` with the `feat/pir-wave2-adrs` PR).

## ADR mapping

Proposed **ruvector ADR-320** — cross-source causal-graph fusion layer on the continuous-latent-state memory tier. Extends **merged** ruvector ADR-307 (three-level memory — specifically the continuous-latent-state tier, WP3 #839) and ruvector ADR-310 (causal-attribution gate, WP6 #850). Fuses atomic, source-tagged observations (AtomicObservation) from multiple agents/sensors into one causal graph with provenance back to the originating event — extending ADR-307's single-agent design to the Latent Communication Fabric's multi-agent setting.

## Evidence

**MemFuse (arXiv:2608.18704, `Darwin-Agent/Mi-Memory`)** — "MemFuse: Multi-Source Memory Fusion from Fragmented Observations," submitted 2026-08-19. **Grade A** — qualitative claim confirmed. **Artifact available now:** `github.com/Darwin-Agent/Mi-Memory/tree/master/MemFuse` (verified real content; an earlier pass wrongly suspected no code existed) — **adapt the released implementation** rather than rebuilding from the paper; benchmark against the released MemFuseBench.

⚠️ **Naming collision — SEVERE (the most important finding of the Wave-2 evidence review):** `github.com/memfuse/memfuse` is an established, actively-forked, **unrelated** open-source LLM memory layer — a direct conceptual competitor under the same name. **Never ship an npm package, crate, or module literally named `memfuse`.** Always cite the paper as **"MemFuse (arXiv:2608.18704, `Darwin-Agent/Mi-Memory`)"**, explicitly distinct from the pre-existing `memfuse/memfuse` project, in every ADR and code comment.

## Scope

- Add the cross-source causal-graph fusion layer to the continuous-latent-state memory tier, adapting the released implementation.
- Surfaces: `agentdb`, `crates/rvf` (continuous-latent-state tier); coordinates with `latentmesh-align`-consuming code per the existing ADR-310 CI gate — this WP's fusion output feeds the same causal audit.

[...]

PR #874, opened 2026-08-20
feat: AtomicObservation + causal episodic graph (PIR WP18, ADR-320)

ruvnet:
## Summary

First shippable slice of **PIR WP18** (RuVector#865) — the MemFuse-pattern **AtomicObservation** + **causal episodic graph** fusion layer, per merged ruvector **ADR-320**. Extends the merged WP4 TARL ledger (ADR-307) from its single-agent design into the multi-agent / multi-source setting, preserving provenance back to each atomic source observation.

> **Naming discipline (ADR-320, binding):** Informed by **MemFuse (arXiv:2608.18704, `Darwin-Agent/Mi-Memory`)**, and **explicitly distinct from the unrelated pre-existing `memfuse/memfuse` OSS memory layer**. No artifact in this PR is named `memfuse` — the citation form appears only in comments/docs for provenance.

## Host / language decision

Extends the existing **`crates/ruvector-agent-memory`** crate (Rust) rather than adding a new crate. That crate already holds the WP4 TARL ledger (`TransactionalLedger`), the ADR-134 witness schema (`ops.rs`), and the ADR-307 lineage this ADR extends — so AtomicObservation + fusion belong next to them. Bonus: no new workspace member, so the #861 workspace guard is unaffected (still verified green).

## What's implemented

- **`observation.rs` — `AtomicObservation`** (event layer): the exact ADR-320 §1 field set — `source`, `time`, `confidence`, `tenant`, `signature`, `causal_parents` — plus an evidence `payload` that provenance resolves back to. **Content-addressed**: `ObservationId = SHA-256(canonical encoding)`, so identity is a cryptographic commitment to the exact content. The `signature` is Ed25519 over that content address.
- **`fusion.rs` — `CausalEpisodicGraph`** (cluster layer): heterogeneous source-tagged observations (RuView RF / network telemetry / agent / user-activity kinds) fuse into `FusedCluster` nodes. **Provenance is load-bearing**: `resolve_provenance()` walks any derived node back to exactly its atomic source observations, transitively through cluster-of-clusters.
- **ADR-320 security gates**: per-observation signature verified at ingest; causal-parents integrity (must be resolvable + acyclic — parents must pre-exist, so the relation is a DAG); tenant isolation (a graph is bound to one tenant; foreign-tenant observations are rejected, never fused).
- **Transactional integrity (ADR-320 §3)**: `ingest_governed()` routes admission through the WP4 `TransactionalLedger` — an observation entering memory is a governed `add` transition that emits an ADR-134 witness record ("no witness, no mutation"), with `causal_parents` mapped onto the ledger's `depends_on` edges (so the ledger's poisoning-containment cascade extends to observation lineage). The ledger is **reused, not rebuilt**.

## Reused vs. built

| Reused (no new scheme) | Built (first-party, from paper + ADR-320) |
|---|---|
| `rvf_types::sha256` (repo SHA-256, content addressing) | `AtomicObservation` type + canonical encoding + signing/verify |
| `rvf_types::{ed25519_sign, ed25519_verify, Ed25519Keypair}` (repo RFC 8032 Ed25519) | `CausalEpisodicGraph` (event→cluster fusion, provenance resolution) |
| WP4 `TransactionalLedger` / `WitnessSink` / ADR-134 witness records | ADR-320 ingest gates (signature / tenant / causal-parents) |

## Tests (`cargo test -p ruvector-agent-memory` → all green)

- `atomic_observation_signature_round_trip` — construction + signature round-trip
- `multi_source_fusion_into_graph` — 3 heterogeneous source kinds → one cluster
- **`provenance_resolves_derived_node_to_atomic_sources`** — the provenance proof: a 2-hop derived node resolves to *exactly* its atomic sources; each resolved id looks up its original evidence
- `confidence_and_tenant_carried_through_fusion` — weakest-link confidence + tenant preserved
- `tampered_observation_is_rejected_at_ingest` — post-signing mutation → signature-invalid rejection
- `governed_ingest_composes_with_wp4_ledger` — governed transition, witness chain intact, causal parent → ledger dependency edge
- plus cross-tenant / unresolved-parent / ungoverned-parent rejection

**24 tests pass** (17 lib unit incl. observation/fusion, 7 integration); the 13 existing TARL ledger tests stay green; `cargo clippy` clean; `scripts/workspace-check.mjs` (#861) OK.

## Honest scope — deferred

- Real RuView / network-telemetry **ingestion adapters** (this slice exercises synthetic source kinds only).
- The **MemFuseBench `research-gate` benchmark** delta (ADR-320 acceptance criterion #2 / candidate-mutation rule) — not run here; this is the type/graph/provenance foundation it will measure against.
- **Scale** (persistence, large-graph indexing) — in-memory `BTreeMap` graph for the slice.
- Per-observation `SignatureVerified`-grade **witness anchoring** into the ledger's evidence grade (WP8 RVM anchoring territory); governed ingest currently records the admission as a standard ledger `add`.
- Flowing fused output through the **ADR-310 causal-audit gate** (acceptance criterion #4) — the graph is designed as that gate's evidentiary substrate but the wiring is a follow-up.

Refs #865, #837.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8793,6 +8793,7 @@ version = "0.1.0"
 dependencies = [
  "rand 0.8.6",
  "ruvector-proof-gate",
+ "rvf-types",
  "serde",
  "serde_json",
 ]
--- a/crates/ruvector-agent-memory/Cargo.toml
+++ b/crates/ruvector-agent-memory/Cargo.toml
@@ -16,6 +16,12 @@ path = "src/main.rs"
 [dependencies]
 rand = "0.8"
 serde = { workspace = true }
+# ADR-320 (PIR WP18): reuse the repo's SHA-256 (content addressing) and
+# RFC 8032 Ed25519 (per-observation signatures) for AtomicObservation — no new
+# hash or signature scheme (ADR-320 Security Gates). `ed25519` feature pulls in
+# ed25519-dalek + rand_core; the always-compiled `sha256` module needs no
+# feature.
+rvf-types = { version = "0.2", path = "../rvf/rvf-types", features = ["ed25519"] }
 # Existing ADR-194/047 proof-gated write machinery; adapts to the ledger's
 # ProofGate trait behind the `proof-gate` feature (see src/ledger.rs).
 ruvector-proof-gate = { path = "../ruvector-proof-gate", optional = true }
--- a/crates/ruvector-agent-memory/src/fusion.rs
+++ b/crates/ruvector-agent-memory/src/fusion.rs
@@ -1,0 +1,438 @@
+//! The causal episodic graph — the cluster-layer fusion structure of ADR-320
+//! (PIR WP18), built over [`AtomicObservation`]s.
+//!
+//! Informed by **MemFuse (arXiv:2608.18704, `Darwin-Agent/Mi-Memory`)**'s
+//! event-layer → cluster-layer architecture, and **explicitly distinct from the
+//! unrelated pre-existing `memfuse/memfuse` open-source project**. Atomic,
+//! source-tagged observations (the event layer) fuse into
+//! [`FusedCluster`]s (the cluster layer), and **every fused/derived node
+//! resolves back to the atomic source observations it was built from** — the
+//! load-bearing provenance guarantee of this layer (see
+//! [`CausalEpisodicGraph::resolve_provenance`]).
+//!
+//! ## Security / validation gates (ADR-320)
+//!
+//! - **Per-observation signature** verified before an observation enters the
+//!   graph ([`CausalEpisodicGraph::ingest`]).
+//! - **Causal-parents integrity**: every `causal_parents` reference must
+//!   resolve to an already-present observation; an unresolvable parent, or a
+//!   self-referential (cyclic) parent, is a hard rejection at ingest, not a
+//!   warning. (Because parents must pre-exist, the parent relation is
+//!   necessarily a DAG.)
+//! - **Tenant isolation**: a [`CausalEpisodicGraph`] is bound to one
+//!   [`Tenant`]; an observation from any other tenant is rejected and never
+//!   fuses in.
+//!
+//! ## Transactional integrity (ADR-320 §3, reusing WP4)
+//!
+//! [`CausalEpisodicGraph::ingest_governed`] routes an observation's admission
+//! through the WP4 [`TransactionalLedger`] (ADR-307 TARL): the observation
+//! enters memory as a governed `add` transition that emits an ADR-134 witness
+//! record ("no witness, no mutation"), with its `causal_parents` mapped to the
+//! ledger's `depends_on` edges so the ledger's poisoning-containment cascade
+//! extends to observation lineage. This crate does **not** rebuild the ledger;
+//! it interfaces to it.
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use crate::ledger::{ProofGate, TransactionalLedger};
+use crate::observation::{AtomicObservation, ObservationId, Tenant};
+use crate::ops::{LedgerError, WitnessSink};
+
+/// Identifier for a cluster-layer [`FusedCluster`] (graph-local, monotonic).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ClusterId(pub u64);
+
+/// A reference to a node in the causal episodic graph: either an event-layer
+/// atomic observation or a cluster-layer fused cluster. A cluster may fuse
+/// other clusters, so provenance resolution is transitive down to atomic
+/// observations.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum NodeRef {
+    Observation(ObservationId),
+    Cluster(ClusterId),
+}
+
+/// A cluster-layer node: related observations (and/or sub-clusters) fused into
+/// one unit, carrying aggregated confidence and the (single) tenant its members
+/// share.
+#[derive(Debug, Clone, PartialEq)]
+pub struct FusedCluster {
+    pub id: ClusterId,
+    /// Event-layer observations and/or sub-clusters fused into this node.
+    pub members: Vec<NodeRef>,
+    /// The tenant all members share (equal to the owning graph's tenant).
+    pub tenant: Tenant,
+    /// Aggregated confidence: the **weakest-link minimum** over member
+    /// confidences. A fused cluster is only as trustworthy as its least
+    /// confident evidence.
+    pub confidence: f32,
+    /// Human-readable fusion key / topic label.
+    pub label: String,
+}
+
+/// Errors from fusion-graph operations.
+#[derive(Debug)]
+pub enum FusionError {
+    /// The observation's signature did not verify (per-observation gate).
+    SignatureInvalid(ObservationId),
+    /// The observation's tenant does not match the graph's tenant boundary.
+    TenantMismatch { expected: Tenant, got: Tenant },
+    /// A `causal_parents` reference does not resolve to a known observation.
+    UnresolvedParent {
+        observation: ObservationId,
+        missing: ObservationId,
+    },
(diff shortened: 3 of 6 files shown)


ruvnet:
## Security audit — Phase 4 (WP18 AtomicObservation + CausalEpisodicGraph)

Audited `observation.rs` + `fusion.rs` at head `23cf4e1da`. The provenance/authenticity model is sound and the four ingest gates are enforced — but `resolve_provenance` has an **unbounded-recursion stack overflow** (MEDIUM). Everything else is clean. Existing tests pass (17 + 7 + 13 + 1 doctest).

### MEDIUM — `resolve_provenance` stack-overflows on a deep cluster chain (process abort)
`fusion.rs`, `collect_provenance` (recursive). The `visited_clusters` guard makes traversal **cycle-safe**, but nothing bounds the **recursion depth**. A linear chain of clusters — `C1 = fuse([obs]); C2 = fuse([C1]); … ; CN = fuse([C(N-1)])`, each `fuse` individually valid — makes `resolve_provenance(CN)` recurse ~N deep. Reproduced: N = 200 000 aborts the process with `fatal runtime error: stack overflow` (SIGABRT). A Rust stack overflow is an **uncatchable abort**, so it takes down the whole process, not just the query — and `resolve_provenance` is described as the layer's "load-bearing provenance guarantee."

So the doc's "cycle-safe transitive traversal" claim is accurate for *cycles* but not for *depth*: deep acyclic nesting is a DoS. Reachability today is bounded by the honest in-memory/synthetic scope (an attacker must be able to drive ~tens of thousands of `fuse` calls, which this slice doesn't wire to untrusted input), so it's a latent hazard rather than a live exploit — but it's a genuine robustness defect that will bite once the fusion layer ingests attacker-influenced structures.

**Fix:** make `collect_provenance` iterative — an explicit worklist/stack (`Vec<NodeRef>`) instead of recursion, keeping the existing `visited_clusters` guard — or impose an explicit maximum cluster-nesting depth and reject beyond it. Either removes the unbounded native-stack usage.

### Checked and clean
- **Causal-parents acyclicity (observation layer):** `validate_for_ingest` rejects a self-parent (`CausalCycle`) and requires every parent to already be present (`UnresolvedParent`). Because a parent must pre-exist and the id is a content hash over the parents, a cross-observation cycle is impossible to construct — the observation causal relation is a DAG by construction. Observation parents are only checked one level (no transitive traversal), so no loop risk there. Clusters are a DAG too (monotonic `ClusterId`, members validated to pre-exist), and the `visited` guard is belt-and-braces on top.
- **Signature gate on ingest:** `validate_for_ingest` calls `obs.verify()` (recomputes the content-address id and Ed25519-verifies it), so a tampered or unsigned (`[0u8;64]`) observation is rejected before insertion. `new_signed` overwrites `source.public_key` with the signing key so identity and key can't disagree, and any field mutation changes the id → verify fails. Note (honest scope, not a finding): `verify()` authenticates under the observation's **self-asserted** key, i.e. it proves integrity + provenance-commitment, not that the signer is an *authorized* key for that tenant/source — the tenant-scoped key registry is explicitly deferred to production scope.
- **Tenant isolation:** structural. Each `CausalEpisodicGraph` is bound to one `Tenant`; `validate_for_ingest` rejects any foreign-tenant observation (`TenantMismatch`), the tenant is committed into the content-address (so ids are tenant-specific), `fuse` can only reference members present in *this* graph, and `resolve_provenance` only returns ids held by this graph. Tenant A's graph cannot fuse with or resolve into tenant B's data.
- **Canonical encoding for signing:** deterministic and injective for the values `new_signed` produces — domain tag `ADR-320-atomic-observation-v1`, fixed field order, LE integers, u32 length-prefixed strings/payload, f32 committed by bit pattern, parents sorted+deduped and count-prefixed. Non-canonical parent ordering/duplication can't change the id (test-covered); the domain tag prevents cross-structure signature replay.
- **Reuse correctness:** uses `rvf_types::sha256` and `ed25519_sign`/`ed25519_verify` — no new hash or signature scheme. `ingest_governed` routes admission through the WP4 `TransactionalLedger::add` (a Pending governed transition emitting an ADR-134 witness — "no witness, no mutation"), maps `causal_parents` → ledger `depends_on` (rejecting `ParentNotGoverned` if a parent lacks a ledger entry), and inserts into the graph only after the ledger accepts (witness-first; on refusal nothing mutates). It adds as **Pending**, never routing to `Accepted` — no proof-gate bypass.

Net: fix the one MEDIUM recursion-depth DoS in `resolve_provenance` (iterative traversal) and this is clean. The authenticity, tenant-isolation, acyclicity, and ledger-routing properties all hold.

ruvnet:
## MEDIUM (recursive provenance traversal) — FIXED

Mapping finding → fix:

| Audit finding | Fix |
|---|---|
| `collect_provenance`/`resolve_provenance` was **recursive** — cycle-safe via `visited_clusters` but **not depth-safe**; a deep linear fuse chain (`C1=fuse([obs])`, `C2=fuse([C1])`, … `CN`) recursed ~N deep → stack overflow → SIGABRT at N=200k on the layer's load-bearing provenance query. | Converted `resolve_provenance` to **iterative** traversal — an explicit heap-allocated worklist (`Vec<NodeRef>`) plus the existing `visited_clusters` guard, so depth is bounded by heap, not the call stack. The old recursive helper is removed. Cycle-safety and error semantics (`UnknownObservation` / `UnknownCluster`) are unchanged. |

Commits: `d43c6d97e` (iterative traversal + depth test), `d652936d1` (remove dead recursive helper).

**New test** `deep_fuse_chain_provenance_is_depth_safe` builds a **200,000-level** nested chain — far past what the recursive version survived — and asserts `resolve_provenance` returns exactly the one atomic source without aborting. It runs in ~0.45s. The cycle-safety guard is retained (defense-in-depth; a cycle is not constructable through the public API since fuse members must pre-exist).

Verification:
- `cargo test -p ruvector-agent-memory` → **39 tests green** (17 lib unit, 8 integration incl. the new depth test, 13 TARL ledger, 1 bench).
- `cargo clippy -p ruvector-agent-memory --all-targets` → clean (no lints, dead-code warning cleared).

Ready for re-verification.

ruvnet:
## Security re-verification — fix `d652936d1` — CLEAR-TO-MERGE

The MEDIUM recursion-depth DoS I reported is fixed. `resolve_provenance` is now iterative and the traversal is genuinely heap-bounded. **Verdict: CLEAR-TO-MERGE.**

### Heap-bounded, no hidden recursion
- `resolve_provenance` now uses an explicit `worklist: Vec<NodeRef>` on the heap with `while let Some(current) = worklist.pop()`; the recursive `collect_provenance` helper is removed (no remaining references). The `visited_clusters` guard is retained, so it stays cycle-safe.
- No hidden recursion elsewhere: `NodeRef` is `enum { Observation(ObservationId), Cluster(ClusterId) }` — both variants hold **Copy ids**, not owned sub-clusters, and clusters live flat in `BTreeMap<ClusterId, FusedCluster>`. So there is no deep owned-structure whose `Drop` could recurse, and `fuse` iterates members once (non-recursive). The only growth is the heap `Vec` worklist.

### Depth is real, and correctness is preserved (verified independently)
- **Depth overflow was genuine:** the same 200 000-level chain that SIGABRT'd the recursive version (my last report) now resolves in ~0.42s. The bundled `deep_fuse_chain_provenance_is_depth_safe` and my own 200k probe both pass — proving the fix addresses an actual overflow, not a hypothetical.
- **No correctness regression on branching/diamond graphs:** I built a diamond (`B=fuse[A,X]`, `C=fuse[A,Y]`, `D=fuse[B,C]`) and confirmed `resolve_provenance(D)` returns exactly `{A,X,Y}` — the shared source `A` is **deduped** (via the `visited` guard + `BTreeSet`), not double-counted — and `D.confidence == 0.5` (weakest-link min preserved). Same atomic-source set and confidence as the recursive version would have produced.
- Full crate suite green (17 + 8 + 13 unit/integration + 1 doctest).

This closes the one finding on #874. The authenticity, tenant-isolation, acyclicity, canonical-encoding, and ledger-routing properties from the prior review are unchanged and still hold — good to merge.

Issue #866, opened 2026-08-20
[PIR][WP19] Self-distilled entity-grounded skill synthesis in the Darwin loop (SkillForge pattern, ADR-321)


ruvnet:
Part of the **RuV Perpetual Intelligence Runtime (PIR), Wave 2**. Epic: #837. Plan: `docs/research/perpetual-intelligence-runtime/07-wave2-program-plan.md` · evidence: `06-wave2-evidence-review.md` (both land on `main` with the `feat/pir-wave2-adrs` PR).

## ADR mapping

Proposed **ruvector ADR-321** — self-distilled, entity-grounded skill synthesis inside the Darwin mutation loop. Extends **merged** ruvector ADR-313 (SHAPER, WP9 #841) directly: a mutation-generation strategy that synthesizes test-covered-functionality issues from the target repo itself (rather than waiting for real incidents), distills entity-grounded skills, and feeds them into the existing frozen-weight harness-evolution loop as one more candidate source.

## Evidence

**SkillForge (arXiv:2608.18933)** — "SkillForge: Self-Distilling Agents for Project-Specific Issue Resolution," submitted 2026-08-19. **Grade B+** — the mechanism is confirmed verbatim, but the headline "beats issue-resolution baselines" claim is **qualitative only: no percentage figure appears in the abstract, so the magnitude of improvement is not independently verifiable**.

⚠️ **Because of the B+ grade, no performance number may be quoted for SkillForge anywhere in this WP's deliverables** — not in the ADR, code comments, commit messages, or status updates. There is no verified number to quote; any figure that appears would be fabricated or sourced from unverified material. The only citable performance claim is this program's own `research-gate`-measured delta.

**Artifact available now:** `github.com/cslsolow/SkillForge` (verified via GitHub API: public, 2.5 MB, real content — `distilling/`, `synthesis/`, `src/`, `evaluation_result/`) — **adapt the released implementation** rather than rebuilding from the paper.

## Scope

- Add synthetic-issue-from-test-coverage skill distillation as a Darwin mutation-candidate source.
- Surfaces: `harness/src/darwin.ts`, `examples/mragent` `scorePolicy` — the same surfaces WP9 already owns; coordinate with WP9's remaining scope.

## Candidate-mutation rule (applies to all of Wave 2)

[...]

PR #875, opened 2026-08-20
feat(sota-harness): SkillForge-pattern synthetic-issue self-training (PIR WP19, ADR-321)

ruvnet:
Part of **PIR Wave 2** (epic #837). Implements **RuVector#866 / ADR-321** — SkillForge-pattern synthetic-issue self-training as a first shippable slice inside the SHAPER harness (`crates/ruvector-sota-bench/harness/src/`), extending the merged WP9 loop (ADR-313).

## What this adds

`skillForge.ts` — a self-supplied Darwin mutation-candidate **source** that exercises the harness against the target repo's own test-covered functionality instead of waiting for a real incident:

- **SELECT** covered functionality — modelled by an injected `CoveredTarget[]` fixture set (functions/modules/ADRs, each with a content-hash integrity anchor and its covering tests).
- **SYNTHESIZE** a project-specific regression from a covered target (`synthesizeIssue`, deterministic — models SkillForge's confirmed "re-implement test-covered functionality" mechanism).
- **REPAIR** via an injected `IssueSolver` trajectory (interface/stub in this slice).
- **DISTILL** a successful trajectory into an **entity-grounded Skill** (`distillSkill`): the skill records **which** repository entities it applies to (files/functions/ADRs) with each entity's exact content hash, plus full provenance (synthetic issue → covered target → trajectory). Shaped as a WP9 `SkillDefinition` (feeds `SkillGenome`) and as a `RuvectorSkillRecord` for memory attachment.

## Constitutional boundary (load-bearing)

skillForge **produces candidate skills; it does not promote them.** A distilled skill is a `SkillProposal` (`kind: "skill-proposal"`). No `promote`/`merge`/`apply` function is exported or reachable; every result is frozen data. A capability-expanding skill routes to ADR-315's `constitutionalGateStub` and is **blocked by default**, and its `capabilityDelta` rides through onto the WP9 `CandidateMutation` so `shaperLoop`'s existing gate re-adjudicates it — this source bypasses no ADR-313/ADR-306 gate. Promotion stays with the WP9 loop → vetoes → flywheel path plus a human.

## Gates

- `scripts/frozen-weights-check.mjs`: **OK** — skillForge.ts is in the mutation surface; no training/weight-writing API, no model-file reference.
- Full harness suite green: **72 tests, 0 fail** (7 new skillForge tests + existing suite; 1 pre-existing env-gated RVF integration test skipped).

## Evidence discipline

Cites **SkillForge (arXiv:2608.18933)**, evidence grade **B+**. The paper's "beats baselines" claim is qualitative only — **no performance magnitude is quoted anywhere** (code, comments, commit, or this PR). The only citable delta is this program's own `research-gate`-measured one.

## Honest scope — follow-ups

- Live test-coverage integration (SELECT is a fixture set here).
- The real frozen-weight repair solver (REPAIR is an injected interface).
- Actual RuVector skill persistence (`SkillMemorySink` is interface-only).
- `research-gate` delta over the pre-WP baseline before any promotion.
- `security scan` after landing (touches the mutation-proposal surface).

Refs #866, #837


ruvnet:
## Security audit — Phase 4 (WP19 SkillForge)

Audited `skillForge.ts` at head `b16a7eb01`. The constitutional boundary and ADR-315 wiring are sound and the frozen-weights gate passes — but the DISTILL success criterion is **under-enforced** relative to its own stated invariant (MEDIUM), plus one LOW. Tests pass 7/7.

### MEDIUM — `distillSkill` earns a skill on the self-declared `succeeded` flag alone, not on `resolvedTests`
`distillSkill` documents "Only a repair that **actually restored the pinned behavior** earns a reusable skill," but its only gates are the `issueId` match and `if (!trajectory.succeeded) return null`. It never checks `resolvedTests`. Confirmed by probe:
```js
distillSkill(issue, { issueId: issue.id, succeeded: true, resolvedTests: [] })
// => PRODUCES a skill (body line: "Restores: ")
```
So a malfunctioning or malicious `IssueSolver` that returns `{ succeeded: true, resolvedTests: [] }` — or `resolvedTests` unrelated to the issue's `breaksTests` — mints a **hollow "successful" skill that restored nothing**. The "failed-trajectory → no-skill" property holds for `succeeded:false`, but the stronger property the invariant claims (a repair that *restored behavior*) is not enforced: `succeeded` and `resolvedTests` are never cross-checked.

Reachability is bounded — the solver is injected/trusted in this slice, and a hollow skill is only a *candidate* that still runs the WP9 loop → dreamMachine eval → vetoes → human before any acceptance, so it cannot self-promote. But this is exactly the integrity check DISTILL exists to be, and it's weaker than advertised; the real (deferred) frozen-weight solver is precisely the "malfunctioning solver" case. **Fix:** when `succeeded`, require `resolvedTests` non-empty and corroborated against the issue (e.g. `resolvedTests ⊇ issue.breaksTests`, or at least a non-empty intersection); otherwise return `null`/skip. A test for the `succeeded:true, resolvedTests:[]` case would lock it in.

### LOW — `entityKey` is not injective (grounding collision)
`entityKey` builds `fn:${path}#${symbol}` / `mod:${path}` / `adr:${adr}` with no escaping, so a `#` (or `:`) in a path/symbol collides two distinct entities. Confirmed:
```js
entityKey({kind:"function", path:"auth",    symbol:"ts#login"}) // "fn:auth#ts#login"
entityKey({kind:"function", path:"auth#ts", symbol:"login"})    // "fn:auth#ts#login"  => equal
```
So a skill grounded to one entity could masquerade as grounding another under the same key. Real file paths / symbol names rarely contain `#`, and the separate `contentSha256` anchor (`isEntityStale`) is a second layer if the retrieval consumer checks it — so this is LOW and the consuming persistence/retrieval is deferred — but the key alone should disambiguate. **Fix:** escape/encode `path`/`symbol` (or use a structured/length-prefixed key).

### Checked and clean
- **Constitutional boundary / no bypass:** no export matches `promote|merge|approve|apply|commit|push` (test-pinned); every result (`SyntheticIssue`, `EntityGroundedSkill`, `SkillProposal`, `SkillForgeResult`, and their arrays) is `Object.freeze`d frozen data. A capability-expanding skill's `capabilityDelta` is carried from the trajectory → `distillSkill` → `toCandidateMutation` onto the WP9 `CandidateMutation` unchanged, so **shaperLoop's ADR-315 gate re-adjudicates it** when consumed; `buildSkillProposal` *additionally* routes it to `constitutionalGateStub` (blocked by default, `blockedByGate` flag). No path reaches promotion, and the capability delta is not dropped between trajectory and candidate. (Honest-scope residual, same as shaperLoop: the delta is *self-declared* by the solver; an undeclared expansion smuggled via the skill body is caught downstream by WP17 TRUSS's observed-expansion gate.)
- **Frozen-weights:** `skillForge.ts` passes the sound gate (0 references) — it imports only `node:crypto` and `genome.js`, names no model-weight file, and its only "weight"/"training" mentions are in comments (correctly stripped).
- **Failed-trajectory handling:** `succeeded:false` → `distillSkill` returns `null` → `runSkillForge` records it in `skipped` (`repair_failed`), never a proposal. (The gap above is specifically the *hollow success* case, not the failed case.)
- **Entity-drift detection:** `isEntityStale` correctly flags a changed `contentSha256`. Note (informational, deferred scope): SkillForge trusts the *self-provided* `contentSha256` — it detects drift from a pinned hash but does not itself recompute the hash from real content at selection; that integrity rests on the deferred live-coverage selector.

Net: tighten the DISTILL success check to corroborate `succeeded` with `resolvedTests` (the MEDIUM), optionally escape `entityKey` (the LOW), and the SkillForge slice is clean — the constitutional/ADR-315/frozen-weights properties all hold.

ruvnet:
Both audit findings from comment 5357137974 fixed in `81ed5199c` — finding → fix:

**MEDIUM (self-declared success) — FIXED.** `distillSkill` no longer trusts the `succeeded` bit alone. New exported `corroboratesRepair(issue, trajectory)` gates distillation on three conjuncts: `succeeded` **AND** `resolvedTests` non-empty **AND** `resolvedTests` intersects the synthetic issue's `breaksTests` (the repair must resolve behavior the issue actually regressed). A non-corroborating trajectory distils no skill and is recorded in `skipped` with the distinct reason `"repair_uncorroborated"` (vs `"repair_failed"`), so a malfunctioning solver that self-declares success lands in `skipped`, never `proposals`. Your exact probes are now tests:
- `succeeded:true, resolvedTests:[]` → NO skill ✅
- `succeeded:true, resolvedTests` unrelated to `breaksTests` → NO skill ✅
- `succeeded:true, resolvedTests` covering `breaksTests` → skill ✅
- end-to-end `runSkillForge` → `skipped[0].reason === "repair_uncorroborated"` ✅

**LOW (non-injective entityKey) — FIXED.** `entityKey` now percent-encodes the structural separators (`%`→`%25` first, then `:`→`%3A`, `#`→`%23`) per component, so `path`/`symbol` bearing a `#` or `:` can no longer forge a false boundary. Collision test added: `entityKey(path:"auth",symbol:"ts#login")` ≠ `entityKey(path:"auth#ts",symbol:"login")`, and `mod:"a:b"` ≠ `mod:"a"`. Plain separator-free paths are unchanged (`fn:src/genome.ts#f`).

**Gates:** full harness suite green (**74 tests, 0 fail**, 1 pre-existing env-gated RVF skip); `scripts/frozen-weights-check.mjs` **OK**. B+ no-magnitude rule intact — no performance number in code, comments, commit, or this reply (grep-verified).

Everything you flagged CLEAN stays untouched (no promote/merge, `capabilityDelta` rides to the ADR-315 gate via `buildSkillProposal`, failed-trajectory→no-skill, `isEntityStale`). Ready for re-verify.

ruvnet:
## Security re-verification — fix `81ed5199c` — CLEAR-TO-MERGE

Both findings closed, verified against the real exports. **Verdict: CLEAR-TO-MERGE.**

### MEDIUM (hollow success) — closed
`corroboratesRepair(issue, trajectory)` now gates `distillSkill`: it requires `succeeded` **and** `resolvedTests.length > 0` **and** `resolvedTests.some(t => issue.breaksTests.has(t))`. Probed against the real module:
- `succeeded + []` → **null (rejected)** — the original hollow-success probe is closed.
- `succeeded + [unrelated test]` → **null (rejected)** — tests that don't intersect the issue's broken set don't corroborate.
- `succeeded + [a real breaksTest]` → **skill** (correct).
- `succeeded:false` → null, recorded as `repair_failed`; an uncorroborated success is recorded distinctly as `repair_uncorroborated` — so the two skip reasons are separable in the audit trail.

**The remaining residual is the honest one, not a new hole:** `corroboratesRepair` verifies the solver *claims* to have resolved a test that the issue broke, but it cannot verify that test *actually* passes (that needs running it — the solver's job). A solver that lists a genuine `breaksTest` in `resolvedTests` without truly fixing it still corroborates. That's the same trusted-solver boundary the whole injected `IssueSolver` seam rests on (and the real solver is deferred), and it's exactly where the trust line belongs — the fix closes the "no/unrelated tests" cases, which is what was actually exploitable without lying.

### LOW (entityKey collision) — closed
`encodeKeyPart` percent-encodes `%`→`%25` **first**, then `:`→`%3A`, then `#`→`%23`. The `%`-first ordering is the correctness pivot, and it holds:
- The old collision `(path:"auth",symbol:"ts#login")` vs `(path:"auth#ts",symbol:"login")` now → `fn:auth#ts%23login` vs `fn:auth%23ts#login` — **distinct**.
- Literal `%23` vs an encoded `#`: `k("a","b%23c")` → `fn:a#b%2523c` vs `k("a","b#c")` → `fn:a#b%23c` — **distinct** (a literal `%` becomes `%25`, so `%23` can't masquerade as an encoded `#`).
- Colon case: `adr:X%3AY` vs `adr:X%253AY` — **distinct**.

Since every `#`/`:` inside a field is encoded, the only literal `#`/`:` in a key are the structural separators, so a key splits uniquely back into its (kind, fields) — injective.

Suite 9/9 (the new corroboration + injectivity cases are locked in); frozen-weights gate still passes; the B+ no-magnitude citation discipline is intact. The constitutional/ADR-315 properties from the prior review are unchanged. Good to merge.

Issue #867, opened 2026-08-20
[PIR][WP20] Pre-compiled pipeline-sharded multi-node ruvLLM serving (pipeline-shards pattern, ADR-323)


ruvnet:
Part of the **RuV Perpetual Intelligence Runtime (PIR), Wave 2**. Epic: #837. Plan: `docs/research/perpetual-intelligence-runtime/07-wave2-program-plan.md` · evidence: `06-wave2-evidence-review.md` (both land on `main` with the `feat/pir-wave2-adrs` PR).

## ADR mapping

Proposed **ruvector ADR-323** (extends **merged ruvector ADR-314**, KV-cache cross-model migration, WP13 #844) — pre-compiled pipeline-sharded serving for ruvLLM on heterogeneous multi-node fleets: pipeline parallelism across nodes with pre-compiled per-stage graphs and speculative decoding. Ships independently of ADR-314 (multi-node topology, not cross-model cache reuse) but lives in the same Cross-Model Cognition Migration context and the same `crates/ruvllm` surface.

⚠️ **Why ADR-323, not ADR-322:** ruvector **deliberately skips 322**. This program has already hit the "wrong repo owns this ADR number" failure mode four times, and ruvector ADR-305/306/310/312/313 repeatedly cite **ruflo ADR-322/322A/322B/322C** as the flywheel-receipt/promotion authority. Minting an unrelated ruvector ADR-322 would put two heavily-cited, differently-scoped "ADR-322"s in concurrent use across sibling repos — exactly the ambiguity ADR-305 §4 exists to prevent. Rationale in `07-wave2-program-plan.md`; confirm 323 is still free against the live `docs/adr/INDEX.md` at kickoff.

## Evidence

**Pipeline shards (arXiv:2608.19147)** — "Pre-Compiled Pipeline Shards for Distributed LLM Inference on Intel AI PC Fleets," submitted 2026-08-19. **Grade A** — the strongest-evidence item in this wave; all figures exact match. **Artifact available now:** `github.com/labscommunity/pipeline-sharded-inference-paper` — the strongest artifact package of the six (full `reproduction/` directory with claims/hardware/results docs + scripts) — **adapt the released reproduction package**.

⚠️ **Number-conflation flag:** the **1.79× figure** and the **4-node/70B interactive-serving result** are two separate results in the source paper. Target the verified 1.79× on an internal 2-node/8B-class same-family setup; treat the 4-node/70B result as a separate, never-1.79×-labeled target.

## Scope

[...]

PR #878, opened 2026-08-20
fix: close two LOW security residuals — caseId FS-collision + shard-placement input bound (PIR hardening)

ruvnet:
Final hardening pass for the PIR program: closes the two tracked LOW security residuals flagged as informational/non-blocking by security audit, for a clean board. Branched from `origin/main`.

## LOW #1 — WP15 caseId case-insensitive-FS collision (#862, re-verify of #873)

`case_id` (`^[A-Za-z0-9._-]+$`) prevents traversal but allows case-only-distinct ids (e.g. `setup_1` vs `Setup_1`) that collide to the same `${caseId}.state.json` filename on a case-insensitive FS (macOS/Windows dev machines; not Linux CI). Two distinct caseIds → one state file → cross-contamination of workspace state, which could forge or mask the acceptance-test integrity signal on those platforms.

**Fix (load-time collision rejection — the cleaner option):** new `assertNoCaseIdCollisions()` in `harnessRiskCases.ts`, wired into `fetchHarnessRiskCases()`, throws at load/parse time when two case_ids lowercase-collide. This catches the problem at the boundary and keeps the readable, debuggable per-case filenames intact. Distinct-ignoring-case ids are unaffected.

- `crates/ruvector-sota-bench/harness/src/harnessRiskCases.ts`
- Tests (`test/harnessRisk.test.ts`): two case-only-distinct ids → rejected (direct + via loader); distinct-ignoring-case ids → fine; first-party set is collision-free.

## LOW #2 — WP20 shard-placement unbounded search (#867 audit note)

`crates/ruvllm/src/shard_placement/planner.rs` — the branch-and-bound `search` recurses to depth = `shards.len()` and is worst-case exponential; pathologically large shard/node counts could stack-overflow or blow up combinatorially. Feature-gated off and fail-closed, but now defensively bounded.

**Fix:** documented input-size caps (`MAX_SHARDS = 64`, `MAX_NODES = 256` — pipeline stages and governed fleets are handfuls) reject over-cap requests with a structured `PlacementRejection::InputTooLarge { shards, nodes, max_shards, max_nodes }` **before** entering the recursive search, bounding both recursion depth and combinatorial breadth.

- `crates/ruvllm/src/shard_placement/planner.rs`
- Tests (`tests.rs`): over-cap request (shards and nodes dimensions) rejected with `InputTooLarge`; at-cap request (64 shards × 256 nodes) still produces a valid plan.

## Verification

- `npm test` (harness): 86 pass / 0 fail / 1 pre-existing skip — includes the 2 new tests.
- `cargo test -p ruvllm --features shard-placement` (shard_placement module): 13 pass — includes the 2 new tests.
- `cargo check -p ruvllm` **without** the feature: clean — no default-build impact.
- `scripts/frozen-weights-check.mjs`: OK (0 references).

Note: `test_time_based_flush_simulation` (witness_log, a wall-clock sleep/elapsed assertion) flakes under CPU contention from parallel worktree agents; it is untouched by this PR (empty diff vs `origin/main`) and unrelated to these changes.

Will be security-re-verified before merge.

Refs #862 #867 #837

### DIFF
--- a/crates/ruvector-sota-bench/harness/src/harnessRiskCases.ts
+++ b/crates/ruvector-sota-bench/harness/src/harnessRiskCases.ts
@@ -73,6 +73,38 @@ export function assertSafeCaseId(caseId: string): void {
   }
 }
 
+/**
+ * Reject a loaded case set in which two distinct case_ids collide once
+ * lowercased (PIR WP15 case-insensitive-FS follow-up, #862).
+ *
+ * Each case_id becomes a `${caseId}.state.json` filename in
+ * rvfWorkspaceBinder.bindWorkspace. assertSafeCaseId confines the charset (no
+ * traversal) but still admits case-only-distinct ids like "setup_1" and
+ * "Setup_1". On a case-INSENSITIVE filesystem (macOS/Windows dev machines; not
+ * Linux CI) those two ids resolve to the SAME state file, so one case's commit
+ * silently overwrites the other's — cross-contaminating workspace state and
+ * letting a case forge or mask the integrity signal the acceptance test reads.
+ *
+ * Detect the collision at load time (the boundary), so the readable, debuggable
+ * per-case filenames are kept intact. Ids that stay distinct when lowercased
+ * (e.g. "setup_1" / "setup_2") are fine.
+ */
+export function assertNoCaseIdCollisions(cases: readonly { caseId: string }[]): void {
+  const seenByStateFile = new Map<string, string>();
+  for (const { caseId } of cases) {
+    const stateFile = `${caseId.toLowerCase()}.state.json`;
+    const prior = seenByStateFile.get(stateFile);
+    if (prior !== undefined) {
+      throw new Error(
+        `case_id collision: ${JSON.stringify(prior)} and ${JSON.stringify(caseId)} both resolve ` +
+          `to the same state file ${JSON.stringify(stateFile)} on a case-insensitive filesystem ` +
+          "(macOS/Windows) — rename one so all case_ids are distinct ignoring case",
+      );
+    }
+    seenByStateFile.set(stateFile, caseId);
+  }
+}
+
 /** Resolve a phase from the row's phase column, falling back to the case-id prefix. */
 export function normalizePhase(raw: unknown, caseId: string): LifecyclePhase {
   if (typeof raw === "string" && PHASE_SET.has(raw)) return raw as LifecyclePhase;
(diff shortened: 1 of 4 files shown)


ruvnet:
## Security re-verification — `202d86ab3` — CLEAR-TO-MERGE

Both LOW residuals I flagged are closed, verified empirically. No new issue, frozen-weights passes, and the flaky test is confirmed unrelated. **Verdict: CLEAR-TO-MERGE** — this clears the last known security residual of any severity.

### (1) WP15 caseId case-insensitive-FS collision — closed
`assertNoCaseIdCollisions` maps each id to `${caseId.toLowerCase()}.state.json` and throws on the first collision; it's wired into `fetchHarnessRiskCases`. Probed against the real export:
- `setup_1` + `Setup_1` → **rejected** (both resolve to `setup_1.state.json` on a case-insensitive FS) — the exact collision I raised is closed.
- `setup_1` / `setup_2` / `Setup_3` → **OK**, and the real `*_fp001` first-party set → **OK** — no over-rejection of legitimately-distinct ids.
It rejects **only** when two ids collide once lowercased, so distinct case-preserving filenames are untouched. Frozen-weights gate on the changed `harnessRiskCases.ts` still passes.

### (2) WP20 shard-placement input bound — closed
`MAX_SHARDS = 64` / `MAX_NODES = 256`, checked at the top of `plan()` (after the empty-shards early return, **before** any recursion) → `PlacementRejection::InputTooLarge { shards, nodes, max_shards, max_nodes }`. This bounds **both** failure modes the earlier LOW named: recursion depth = `shards.len()` is capped at 64 (no native-stack exhaustion), and per-level fan-out at 256 (with `MAX_SHARDS`, the combinatorial ceiling). Boundary-probed:
- exactly **64 shards → plans**; **65 shards → InputTooLarge**; **257 nodes → InputTooLarge**.
The caps are sane pipeline-scale values (real 70B pipelines run a handful-to-few-dozen stages; 64 stages / 256 nodes sits comfortably above realistic requests, so no legitimate placement is rejected). Shard-placement suite 13/13.

### No new issue / flaky test unrelated
Both changes are purely additive — a new guard function called at load time, and a new early-return before the unchanged branch-and-bound. The existing placement algorithm and case-parsing are untouched, and all prior tests still pass (shard-placement 13, harnessRisk 17). The flaky `tests::witness_log_tests::test_time_based_flush_simulation` lives in `crates/ruvllm/src/tests/witness_log_tests.rs` — **not** among this PR's 4 changed files — and is a pre-existing wall-clock timing assertion (`sleep(50ms)` / `assert < 100ms`) that flakes under parallel-CPU load; it is genuinely unrelated to and uncaused by this change.

Good to merge.

Issue #880, opened 2026-08-21
[PIR][WP21] SPADE-pattern self-play environment generation + REQUIRED external-grounding veto (ADR-324)


ruvnet:
**Epic:** #837 · **Wave 3, Phase W3-1 (top priority)** · **Proposed ADR-324** · **Paper:** SPADE — *Self-Play in Adaptive Synthetic Executable Environments*, **arXiv:2608.19197** (Grade **A**, all figures exact-match; reference implementation **live** at [github.com/spade-rl/spade](https://github.com/spade-rl/spade), MIT, real content)

> ⚠️ **CITATION DISCIPLINE — NEVER write bare "SPADE".** Always cite as **SPADE (arXiv:2608.19197, Self-Play in Adaptive Synthetic Executable Environments)** on first use in every code comment, ADR, commit message, PR, and status update. "SPADE" alone collides — in the *same* multi-agent domain — with the long-established **SPADE (Smart Python multi-Agent Development Environment)** framework. This is the most severe name collision of any PIR wave; a bare reference reads as a completely reasonable citation to the wrong, decades-old project. (`08-wave3-evidence-review.md` §1)

## Goal

Add an **Environment-Designer mutation-candidate source** to Darwin's proposal surface: an LLM-authored, executable Gym-style (`reset()` / `step()`) training environment, generated by the same frozen model under a regret signal, as a **new type of Darwin mutation candidate** — the same "one more candidate-generation strategy feeding the existing pipeline" shape as WP19 SkillForge (ADR-321). Adapt the released `spade-rl/spade` implementation rather than rebuilding from the paper.

Extends merged **ADR-306** (dream-machine / `research-gate` promotion gate), **ADR-313** (SHAPER frozen-weight loop, WP9), **ADR-321** (SkillForge candidate-source pattern, WP19).

## 🔒 REQUIRED acceptance criterion — external-grounding veto (the load-bearing item, ahead of any feature work)

**Every generated environment MUST trace to external evidence** — repository tests, real observed/incident failures, physical-world observations, or an independently published benchmark. **A self-invented closed loop where the same system that generated the environment is also the sole judge of its value is a REJECT.** This is a hard acceptance criterion, not a nice-to-have:

- Implement a **new hard veto** in `research-gate` (`crates/ruvector-sota-bench/harness/src/statistics.ts`, alongside the existing paired-bootstrap statistical veto) requiring every generated environment to carry a traceable external-grounding link before it can reach the promotion gate **at all**.
[...]

PR #887, opened 2026-08-21
feat(sota-harness): SPADE-pattern self-play environment designer with external-grounding veto (PIR WP21, ADR-324)

ruvnet:
## PIR WP21 — SPADE-pattern self-play Environment Designer (ADR-324)

Owner slice for **#880** (Epic **#837**). Implements ADR-324's first shippable slice: an LLM-authored, executable, **Gym-style** training-environment generator as a **new Darwin mutation-candidate source** for the Dream Machine evaluation core — with the **external-grounding hard veto** as the load-bearing gate.

Source paper (cited in full on every reference — the bare name collides in-domain with the **Smart Python multi-Agent Development Environment** framework): **SPADE (arXiv:2608.19197, *Self-Play in Adaptive Synthetic Executable Environments*)**, evidence grade A, reference implementation live at github.com/spade-rl/spade (MIT).

### What landed (2 files, additive)
- `crates/ruvector-sota-bench/harness/src/environmentDesigner.ts`
- `crates/ruvector-sota-bench/harness/test/environmentDesigner.test.ts`

Sibling in shape to WP19's `skillForge.ts` (ADR-321); feeds ADR-313's SHAPER loop and ADR-306's dream-machine → vetoes → flywheel path unchanged.

### Design
1. **Typed `EvaluationEnvironment`** (`reset()`/`step()`/`reward`/`verify`) plus a first-party `generateEnvironment` that builds executable environments **from external evidence**. The verification key is derived from that evidence, so grounding is load-bearing at the environment *mechanic*, not only at the veto.
2. **Regret-driven difficulty** (`measureRegret`): the reasoning path attempts the environment twice — once normally, once with privileged evidence — and the **reward gap is the difficulty/regret signal**, so difficulty follows the learner (ADR-324 §2, per ruv's implementation note on #880).
3. **THE HARD INVARIANT — external-grounding veto (ADR-324 §4, the crux).** `externalGroundingVetoProvider` REJECTS any environment whose provenance does not resolve to a concrete external-evidence source (repo test / observed failure / RuVector memory / published benchmark). `groundingGatedVetoProvider` fires it **before** the dream-machine evaluation stage and **short-circuits** — an ungrounded environment never reaches evaluation. Composed via `vetoes.ts`'s existing `composeVetoProviders`. A self-invented closed loop (no external trace) is a **REJECT**, not a warning; a forged or drifted cite is rejected too (a real cite cannot launder a forgery beside it).
4. **Frontier check** (`atFrontier`): promotable only if **solvable WITH** privileged evidence **AND** gap-exposing **WITHOUT** it; mastered (regret ≈ 0) or unsolvable (regret saturated) environments are retirement candidates (ADR-324 §3).
5. **Constitutional boundary**: environments are proposals — frozen data, **no promote/merge/apply export** (test-asserted, like WP2/WP9/WP19). A capability-expanding environment routes to **ADR-315's constitutional gate stub** (WP9) and is **BLOCKED by default**. The candidate feeds the existing path via `toContextCandidateMutation` (context surface — the closed `MutationSurface` union is unchanged) and `toDreamCandidate`.

### External-grounding-veto proof (the headline security test)
`EXTERNAL-GROUNDING VETO (headline security gate)` asserts: a grounded environment whose evidence **resolves** passes (`[]`); a **self-invented** environment is vetoed (`external_grounding_absent`); a **forged** cite is vetoed (`external_grounding_unresolved`); a **drifted** cite is vetoed (`external_grounding_drift`); and composed conjunctively with a "no objection" downstream, an ungrounded environment still yields a non-empty veto set — i.e. **cannot be promoted**. `VETO FIRES BEFORE EVALUATION` proves the evaluation stage is **never invoked** when grounding fails.

### Test output
- New suite: **7/7 pass**.
- Full harness suite: **94 tests, 93 pass, 0 fail, 1 pre-existing env-gated skip**.
- `scripts/frozen-weights-check.mjs`: **OK** — 33 files across 3 mutation surfaces, 0 training/weight-API references (the new file is in the scanned surface).
- `security scan --depth standard`: **0 Critical / 0 High**; the one MEDIUM is a pre-existing false positive in `skillForge.ts` (substring match, not this change).

### Preprint-reproduction rule (ADR-324)
The paper's numbers (+5.3 / +5.7 / +13.9, 30B params, its own suite) are **hypotheses**, not this program's bar. No performance magnitude is quoted anywhere in the code; the only citable delta is this program's own research-gate-measured one.

### Honest scope — deferred
The live 30B self-play, real reasoning-agent execution, real external-evidence resolution (filesystem / RuVector / benchmark lookup — modelled by the injected `EvidenceResolver`), and the empirical acceptance run (ruv's held-out-20% / ≥10% / zero-violation criterion) are **deferred**. This is the **designer + grounding-gate + regret-signal** slice.

Refs #880, #837. Implements ADR-324.

### DIFF
--- a/crates/ruvector-sota-bench/harness/src/environmentDesigner.ts
+++ b/crates/ruvector-sota-bench/harness/src/environmentDesigner.ts
@@ -187,16 +187,36 @@ export interface EvaluationEnvironment {
 }
 
 /**
- * The verification key an environment checks an action against. Derived by
- * hashing the environment's external evidence — so the key is UNCOMPUTABLE
- * without that evidence's content. Grounding is therefore load-bearing at the
- * environment mechanic itself, not only at the veto: a designer with no
- * external anchor cannot even construct a solvable challenge.
+ * The verification key an environment checks an action against — the SECOND,
+ * defense-in-depth grounding layer (the resolver-backed veto is the primary
+ * gate). The key is bound to the RESOLVER's CURRENT resolved content hash
+ * (`currentSha256`), NOT to the self-declared `contentSha256`, so it is
+ * genuinely UNCOMPUTABLE without access to the real external source:
+ *   - the resolver must run (you must be able to resolve the evidence), and
+ *   - the key mixes in the content hash the resolver returns from the real
+ *     source, which a designer holding only fabricated strings does not have.
+ *
+ * Returns `null` — an UNUSABLE key, never a fixed public constant — when the
+ * provenance has NO evidence (self-invented loop) or when ANY evidence fails to
+ * resolve (forged / invented cite). An environment with a null key is
+ * intrinsically UNSOLVABLE: no action can satisfy it. So a designer with no
+ * real external anchor cannot construct a solvable challenge at all — grounding
+ * is load-bearing at the environment mechanic itself, not only at the veto.
+ * (Requiring EVERY evidence to resolve mirrors the veto: a real cite cannot
+ * launder a forged one alongside it.)
  */
-export function verificationKeyFor(provenance: EnvironmentProvenance): string {
+export async function verificationKeyFor(
+  provenance: EnvironmentProvenance,
+  resolve: EvidenceResolver,
+): Promise<string | null> {
+  if (provenance.evidence.length === 0) return null;
   const hash = createHash("sha256").update("ruvector.spade.env-key.v1\0");
   for (const item of provenance.evidence) {
-    hash.update(`${item.kind}\0${item.locator}\0${item.contentSha256}\0`);
+    const outcome = await resolve(item);
+    if (!outcome.resolved || outcome.currentSha256 === undefined) return null;
+    // Bind to the resolver's CURRENT content hash from the real source, never
+    // to the designer's self-declared item.contentSha256.
+    hash.update(`${item.kind}\0${item.locator}\0${outcome.currentSha256}\0`);
   }
   return hash.digest("hex").slice(0, 16);
 }
@@ -212,32 +232,44 @@ export interface EnvironmentSpec {
   readonly capabilityDelta?: CapabilityDelta;
 }
 
-/** A generated environment: the executable interface plus its originating spec. */
+/**
+ * A generated environment: the executable interface plus its originating spec.
+ * `verificationKey` is `null` when the provenance did not resolve to real
+ * external content (self-invented or forged) — such an environment is
+ * intrinsically unsolvable (the second grounding layer, see `verificationKeyFor`).
+ */
 export interface GeneratedEnvironment {
   readonly spec: EnvironmentSpec;
   readonly environment: EvaluationEnvironment;
-  readonly verificationKey: string;
+  readonly verificationKey: string | null;
 }
 
-function buildObservation(spec: EnvironmentSpec, key: string, privileged: boolean): EnvObservation {
+function buildObservation(spec: EnvironmentSpec, key: string | null, privileged: boolean): EnvObservation {
   const base =
     `[${SPADE_CITATION}] Task: ${spec.task}. ` +
     `Grounded in ${spec.provenance.evidence.length} external-evidence source(s). ` +
     `Emit the verification key to solve.`;
+  // Only a resolver-backed key can be revealed; an unresolved (null) key leaves
+  // even the privileged observation with no solvable answer.
   return Object.freeze({
-    prompt: privileged ? `${base} PRIVILEGED HINT — key=${key}` : base,
+    prompt: privileged && key !== null ? `${base} PRIVILEGED HINT — key=${key}` : base,
     privileged,
   });
 }
 
 /**
- * Build an executable EvaluationEnvironment from a spec. Does NOT throw on empty
+ * Build an executable EvaluationEnvironment from a spec, threading the same
+ * `resolve` the grounding veto uses. Does NOT throw on empty / unresolved
  * evidence: a self-invented environment is representable so the grounding veto
  * can be PROVEN to reject it (the invariant is a promotion-path veto, ADR-324
- * §4, not a constructor guard). It does validate the shape of any evidence
- * present, and the env id/task.
+ * §4, not a constructor guard) — but its `verificationKey` is `null`, making it
+ * intrinsically UNSOLVABLE (defense-in-depth). It validates the shape of any
+ * evidence present, and the env id/task.
  */
-export function generateEnvironment(spec: EnvironmentSpec): GeneratedEnvironment {
+export async function generateEnvironment(
+  spec: EnvironmentSpec,
+  resolve: EvidenceResolver,
+): Promise<GeneratedEnvironment> {
   if (!spec.id) throw new Error("environment spec requires a non-empty id");
   if (!spec.task) throw new Error("environment spec requires a non-empty task");
   if (spec.provenance.citation !== SPADE_CITATION) {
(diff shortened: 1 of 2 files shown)


ruvnet:
### Security review (Wave-3 auditor) — 1 MEDIUM, defense-in-depth accuracy

The external-grounding veto itself is **sound** — I could not bypass it. Empty / forged-locator / drifted evidence each independently produce a non-empty veto set (`external_grounding_absent` / `_unresolved` / `_drift` / `_malformed`), `groundingGatedVetoProvider` short-circuits so an ungrounded env structurally never reaches the downstream eval stage, the resolver is a required injected seam with no self-certifying default, `constitutionalGateStub` blocks capability-expansion in both branches, there is no promote/merge export, and `frozen-weights-check` passes (0 training/weight-API refs). Great primary gate.

**MEDIUM — `verificationKeyFor` is not bound to resolved external content, so the documented second defense layer ("grounding is load-bearing at the environment mechanic itself, not only at the veto"; "the key is UNCOMPUTABLE without that evidence's content", `environmentDesigner.ts:189-202`) does not hold.**

`verificationKeyFor(provenance)` hashes only self-declared `provenance` fields (`kind`, `locator`, `contentSha256`) — strings the *designer* authors. It never calls the `EvidenceResolver` and never checks the declared `contentSha256` against the resolver's `currentSha256`. Consequences, both reproduced against the exact function body:

- **Empty evidence → a fixed public constant key** (`sha256("ruvector.spade.env-key.v1\0")[:16]` = `5ebf4a16519a874b`). A fully self-invented, closed-loop env *can* construct a solvable challenge — directly contradicting "a designer with no external anchor cannot even construct a solvable challenge."
- **Forged-but-well-formed evidence → a deterministic key** from the attacker's own `contentSha256` string (`repo-test` / `test/i-made-this-up.test.ts` / `d…d` → `d709a1486ce32ce8`), with zero resolution.

The test at `test/environmentDesigner.test.ts:74-75` asserts only determinism (`key === verificationKeyFor(SAME_PROVENANCE)`) yet its comment claims "uncomputable without it," reinforcing the overstated property for future maintainers.

**This is not a promotion bypass** — the resolver-backed veto is the real grounding gate and it holds. The impact is that the *redundancy* the design advertises is vacuous: the entire grounding safeguard rests on the veto alone, and any later change that trusts "the mechanic self-enforces grounding" (e.g. relaxing veto ordering) would silently lose all enforcement.

**Fix (pick one):** (a) make the key genuinely resolver-bound — derive/verify it from the resolver's `currentSha256` (real resolved content) rather than the designer-declared `contentSha256`, so an unresolved/forged env cannot produce a valid key; or (b) downgrade the docstring + test comment to state plainly that the veto is the sole grounding gate and the key is a self-declared identity digest, not an independent anchor.

Scope items confirmed honest (not findings): injected `EvidenceResolver`, live 30B self-play, real evidence resolution, empirical acceptance run. The `skillForge.ts:268` substring-match FP is pre-existing (0-line diff vs main; not in this PR).

ruvnet:
## Audit MEDIUM fixed for real — verification key now resolver-bound (SHA `65791fc8f`)

Thanks for the adversarial pass. Confirmed the finding and fixed the **second (defense-in-depth) grounding layer** so it is genuinely load-bearing — not by softening the docstring, but by making the claim true.

### The finding (correct)
`verificationKeyFor` hashed **only self-declared provenance strings** (`kind`/`locator`/`contentSha256`), never calling the resolver. So:
- empty evidence → a fixed **public constant** key;
- forged well-formed evidence → a deterministic key from the attacker's **own strings**, zero resolution.

The "UNCOMPUTABLE without that evidence's content" claim was false, and the old test asserted only determinism while its comment claimed "uncomputable" — reinforcing the false property. (The resolver-backed **veto** — the primary gate — was and stays sound; this was the vacuous *second* layer.)

### The fix (mapping)
| Coordinator ask | Change |
|---|---|
| Key must call the resolver and bind to CURRENT resolved content, not self-declared `contentSha256` | `verificationKeyFor(provenance, resolve)` now `await`s the resolver per evidence and mixes in `outcome.currentSha256` (the real source's hash), never `item.contentSha256`. |
| No real anchor ⇒ key construction fails (unusable), not a constant | Returns `null` when provenance is empty **or any** evidence fails to resolve. An env with a `null` key is intrinsically **unsolvable** — `step()` satisfies no action, and privileged `reset()` reveals nothing. |
| Thread the same injected `EvidenceResolver` the veto uses | `generateEnvironment(spec, resolve)` is now async and takes the resolver; the tests pass the same `staticEvidenceResolver` the veto uses. |
| Test the REAL property (differs for an attacker lacking resolved content; empty/forged → no valid key, not a constant); keep determinism-given-real-resolution | New `KEY BINDING (defense-in-depth)` test: a resolver reporting **different real content** for the same locator yields a **different key**; an attacker with only the provenance (resolver resolves nothing) → `null`; empty/forged → `null`; a null-key env returns reward 0 for every guess **including the old self-hash constant**. Determinism retained as determinism-given-real-resolution. |
| Keep the docstring claim only if now true | Docstring rewritten to state the resolver-binding precisely; it is now true. |

### Gates (re-run in my worktree with `@metaharness` workspace deps installed — the bare-worktree `ERR_MODULE_NOT_FOUND @metaharness/darwin` you hit is `npm install` not having run there)
- `npm test` (build + `node --test`): **95 tests, 94 pass, 0 fail, 1 pre-existing env-gated skip** (`RVF_ADAPTER_IT`).
- New suite: **8/8** (added `KEY BINDING`).
- `scripts/frozen-weights-check.mjs`: **OK** — 33 files / 3 mutation surfaces, 0 training/weight-API references.
- `security scan --depth standard`: **0 Critical / 0 High**; the lone MEDIUM is the pre-existing `skillForge.ts:268` false positive, not this change.

The auditor can re-verify the key binding at `verificationKeyFor` (src/environmentDesigner.ts): the only hash input from the evidence is `outcome.currentSha256`, and both empty and unresolved paths `return null` before any digest.

ruvnet:
### Re-verification (Wave-3 auditor) — MEDIUM RESOLVED ✅

Commit `65791fc8f` fully fixes the finding. Verified by type/logic review + a standalone repro of the old forgeability against the exact new function body:

- **Key is now resolver-bound, not self-declared.** `verificationKeyFor(provenance, resolve)` mixes in `outcome.currentSha256` (the resolver's hash of the real source) and no longer the designer's `item.contentSha256`. Confirmed: same locator + *different resolved content* → a **different** key, so an attacker holding only the provenance strings cannot reproduce it. (`kind`/`locator` remain in the preimage as the non-secret resolvable handle — harmless, since the unforgeable component is `currentSha256`.)
- **No fixed-constant path remains.** Empty evidence returns `null` before any digest; any unresolved item returns `null` before `.digest()`. Repro against the new logic: empty→`null`, attacker-with-no-resolution→`null` (the old `5ebf4a16519a874b` constant and the old forged self-hash are both gone).
- **A null-key env is intrinsically unsolvable.** `step` guards `key !== null && action === key`, the privileged observation omits the hint when the key is null, and `verify` returns the (always-false) `solved` — so no guess, including the old constant, ever scores. Test-asserted.
- **Primary veto unchanged (regression-safe).** The four hunks touch only `verificationKeyFor`/`generateEnvironment`/`buildObservation`/`step`; `groundingReasons`, `externalGroundingVetoProvider`, `groundingGatedVetoProvider`, and `staticEvidenceResolver` are byte-identical.
- New KEY BINDING test asserts the real property; `environmentDesigner.ts` + its test typecheck clean (async threading correct, all callers `await`ed); `frozen-weights-check` still OK.

The second defense-in-depth layer is now genuinely load-bearing and the docstring no longer advertises a false safety property. **CLEAR TO MERGE** on this finding.

Minor, non-blocking: the null-guard checks `currentSha256 === undefined` but not emptiness/format — a resolver returning `resolved:true` with an empty/garbage `currentSha256` would still yield a key. That's a self-certifying/misbehaving resolver, i.e. the same trusted injected seam the primary veto already depends on, so it's a consistent trust boundary rather than a new gap; a real `EvidenceResolver` (WP-deferred) returns a genuine content hash.

Issue #882, opened 2026-08-21
[PIR][WP23] DeAR-pattern decentralized capability grounding + thought-map navigation (ADR-326)


ruvnet:
**Epic:** #837 · **Wave 3, Phase W3-3** · **Proposed ADR-326** · **Paper:** DeAR — *Decentralized Agentic Reasoning via Capability Grounding* (arXiv:2608.17282, Grade **B+** — headline mechanism + benchmark count confirmed verbatim; one behavioral detail unconfirmable from the abstract). ⚠️ **No upstream code today** — the abstract states code ships only "upon acceptance." This is a **reproduction-from-description** build, materially larger scope than an adapt-from-repo WP.

## Goal

Add three mechanisms to the Latent Communication Fabric:
1. **Decentralized capability grounding** — `C(i,j,q)` capability vectors with **query-dependent peer specialization and no central role-assigner** (local peer selection).
2. **Shared thought-map** — a RuVector thought-graph data structure agents navigate directly, rather than exchanging only point-to-point messages. LatentMesh transport.
3. **Dead-end-triggered topology mutation.**

Extends merged **ADR-309/310/311** (LatentMesh greenfield fabric, causal-attribution gate, anomaly quarantine).

## ⚠️ CRITICAL — the continue-vs-restart behavior is UNVERIFIED (do NOT harden an acceptance criterion around it)

The design intent is that a detected **dead-end triggers a topology update to CONTINUE, not restart**. **This "continue, not restart" reading is this program's working interpretation of the abstract's "topology update for adaptive error correction" — it is NOT a verified quotation from the paper** (`08-wave3-evidence-review.md` §3, and the reason DeAR grades B+ rather than A).

- The ADR-326 Decision section MUST NOT treat continue-vs-restart as settled, and this WP MUST NOT harden its implementation around it as a fixed behavioral contract.
- **Re-verify against the full paper (not available today) or the eventual code release before finalizing.** If the full text or code contradicts the working interpretation, ADR-326's Decision section needs revision before WP23 locks in a specific behavioral contract.

## Acceptance criteria

- Query-dependent peer specialization works with **no central role-assigner** (decentralized).
- Shared thought-map is directly navigable; agents traverse it rather than only exchanging point-to-point messages.
- Topology changes remain **attributable** (WP6 causal-audit gate) — every topology mutation traces to a cause.
- A **misbehaving peer cannot trigger spurious topology churn** (WP7 anomaly quarantine).
[...]

PR #888, opened 2026-08-21
feat: DeAR-pattern decentralized capability-grounded reasoning (PIR WP23, ADR-326)

ruvnet:
## PIR WP23 — RuVector#882 · ADR-326 · Epic #837

First shippable slice of DeAR-pattern decentralized capability-grounded reasoning, implementing merged **ADR-326** over the LatentMesh fabric (**ADR-309** transport / **ADR-310** causal gate / **ADR-311** quarantine).

**Evidence posture (preprint-reproduction rule).** Informed by **DeAR** — *Decentralized Agentic Reasoning via Capability Grounding and Collaborative Thought Navigation* ([arXiv:2608.17282](https://arxiv.org/abs/2608.17282)), graded **B+ (not A)**, with **no released code** (the paper's repo is a placeholder `https://open_upon_acceptance` URL). This is a **first-party build from the paper's description**, not a port. DeAR's reported numbers are hypotheses, not the acceptance bar; promotion requires an independently recomputed `research-gate` delta over the pre-WP baseline.

### Host decision
No `latentmesh-*` crates exist on `origin/main` yet (only ADR-305/ADR-309 docs), so the WP5 greenfield crates are not substantial. WP23 is hosted as a **new crate `crates/rvAgent/latentmesh-thoughtmap`** — exactly the crate name ADR-326's *Affected Repos* names as the likely landing spot, sitting alongside the other rvAgent crates. It establishes the first `latentmesh-*` crate that WP5's transport work extends via the seams below.

### What's implemented
1. **Capability vectors replace named roles** (`capability.rs`) — `C(i,j,q) = competence × trust × locality × freshness ÷ cost`, recomputed per-query. `competence`/`locality`/`freshness` are grounded against the specific query, so specialization is dynamic, not a fixed label. `LocalView::select_peer` chooses the next peer **locally** — no central planner/dispatcher argument exists in the signature (the contrast point with `radio-moe`'s centralized routing).
2. **Shared thought map** (`thoughtmap.rs`) — reasoning-step nodes, peer-interaction edges; success **reinforces** edges, failure **weakens** them. First-party in-memory graph with a documented **RuVector-integration seam** (`ThoughtGraphBackend` trait) a `ruvector-graph`-backed store implements later without changing callers.
3. **LatentMesh transport + causal verification** (`transport.rs`, `causal.rs`) — messages carry **compressed latent state** (KV-cache-style, not token streams — ADR-326's first-mover angle). A received message must pass ADR-310's **controlled-replacement causal audit** (`ControlledReplacementVerifier`) before it is incorporated: it must be *attributable* to its claimed antecedent **and** that antecedent must be *specifically* load-bearing (a control substitution catches spurious attribution). Full wire transport (codec + `ruvnet/LatentMesh` coordination) is a defined seam with an in-process stub — **deferred to WP5**.
4. **Dead-end handling** (`topology.rs`) — on a dead end, change topology and **continue from current state, not restart**.
5. **Security** (`quarantine.rs`, `engine.rs`) — `ReasoningEngine::incorporate` is the single choke point: integrity (ADR-311) → quarantine dampening → causal attribution (ADR-310). A misbehaving peer accrues strikes and is quarantined; quarantined peers are dampened and **cannot trigger topology churn**. Every topology change is attributable (carries `cause` + `triggered_by`).

### ⚠️ Continue-vs-restart is marked UNVERIFIED (ADR-326 §6 binding gate)
The "continue, not restart" dead-end behavior is **this program's own working interpretation** of the abstract's four-word *"topology update for adaptive error correction"* — **NOT a reproduced DeAR claim**. Per ADR-326's binding design-choice-labeling discipline it is loudly labeled as such in `topology.rs` module docs, on `DeadEndPolicy` and its `Default`, in `engine.rs`, and in `lib.rs`, and kept **swappable** (`DeadEndPolicy::Restart` is a one-line flip with nothing else depending on the choice). Must be re-verified against the full paper or released code before being treated as a fixed contract.

### Tests (28 green: 20 unit + 8 acceptance)
- Capability-vector scoring is the ADR formula; competence is query-dependent; local peer selection picks best with no central planner.
- Edge reinforce on success / weaken on failure.
- **A latent message failing causal verification is REJECTED, not incorporated** (`node_count` unchanged); genuine message accepted; tampered message rejected on integrity.
- Dead-end → topology change **continues from current state** (resume_from = current node, not root), and policy is swappable to Restart.
- A misbehaving peer is quarantined, dampened (a later good message still not incorporated), and cannot churn topology; re-routing skips quarantined peers.
- Topology changes are attributable.

`cargo test -p latentmesh-thoughtmap` green, `cargo check` + `clippy` clean, workspace membership added and verified via `scripts/workspace-check.mjs` (0 orphans). `npx @claude-flow/cli security scan` run (touches a topology-mutation surface) — no findings in this crate.

### Deferred (honest scope)
Real multi-agent execution, the 9-benchmark DeAR evaluation, and full LatentMesh **wire** transport (compression codec + wire-format coordination) — the transport is a defined seam (`LatentTransport`) with an in-process stub for WP5 to implement.

Refs #882, #837. Depends on ADR-309/310/311.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4915,6 +4915,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "latentmesh-thoughtmap"
+version = "0.1.0"
+dependencies = [
+ "serde",
+ "serde_json",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "lattice-embed"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -209,6 +209,9 @@ members = [
     "crates/rvAgent/rvagent-mcp",
     "crates/rvAgent/rvagent-wasm",
     "crates/rvAgent/rvagent-a2a",
+    # PIR WP23 / ADR-326 — DeAR-pattern decentralized capability-grounded
+    # reasoning. First latentmesh-* crate; extended by ADR-309 WP5 transport.
+    "crates/rvAgent/latentmesh-thoughtmap",
     # ADR-159 a2a-swarm demo
     "examples/a2a-swarm",
     # ETL pipeline example
--- a/crates/rvAgent/latentmesh-thoughtmap/Cargo.toml
+++ b/crates/rvAgent/latentmesh-thoughtmap/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "latentmesh-thoughtmap"
+version = "0.1.0"
+edition = "2021"
+description = "DeAR-pattern decentralized capability-grounded reasoning over a shared thought map (PIR WP23, ADR-326). First latentmesh-* crate; extended by ADR-309 WP5 transport crates."
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/ruvnet/RuVector"
+
+[dependencies]
+serde = { workspace = true }
+thiserror = { workspace = true }
+
+[dev-dependencies]
+serde_json = { workspace = true }
+
+# Research-tier greenfield crate (PIR WP23). Correctness + suspicious lints
+# stay denied; doc/style churn deferred, matching sibling rvAgent crates.
+[lints.rust]
+missing_docs = "allow"
--- a/crates/rvAgent/latentmesh-thoughtmap/src/capability.rs
+++ b/crates/rvAgent/latentmesh-thoughtmap/src/capability.rs
@@ -1,0 +1,272 @@
+//! Capability vectors replacing named agent roles (ADR-326 Decision §1–§2).
+//!
+//! DeAR's confirmed mechanism (arXiv:2608.17282, part 1) is *query-dependent
+//! agent specialization*. We operationalize it by replacing fixed labels like
+//! `"coder"` / `"reviewer"` with a capability vector recomputed **per query**:
+//!
+//! ```text
+//! C(i, j, q) = competence × trust × locality × freshness ÷ cost
+//! ```
+//!
+//! (ruv's formula in RuVector#882 / ADR-326.) `competence`, `locality`, and
+//! `freshness` are grounded against the specific query `q`, so an agent's
+//! effective specialization is a function of its measured profile against that
+//! query — not a static role.
+//!
+//! **Local peer selection is decentralized (ADR-326 §2).** [`LocalView::select_peer`]
+//! takes only the *selecting* agent's own view plus the candidate profiles it
+//! can observe. There is deliberately **no central planner / dispatcher / global
+//! registry** argument — the contrast point with `radio-moe`'s centralized
+//! reputation-weighted routing called out in ADR-326. Each agent chooses its
+//! next collaboration peer by locally comparing `C(i,j,q)`.
+
+use serde::{Deserialize, Serialize};
+
+use crate::vecmath::cosine;
+
+/// Stable identifier for an agent / peer participating in the reasoning fabric.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct PeerId(pub u64);
+
+/// A query the reasoning fabric is collaborating on. Carries a small feature
+/// embedding so competence/locality/freshness can be grounded query-dependently
+/// (`C(i,j,q)` depends on `q`).
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct Query {
+    pub id: u64,
+    /// Feature/embedding vector describing what the query needs.
+    pub features: Vec<f32>,
+}
+
+impl Query {
+    pub fn new(id: u64, features: Vec<f32>) -> Self {
+        Self { id, features }
+    }
+}
+
+/// An observable profile of a candidate peer. This is what an agent can *see*
+/// about another agent locally — its advertised skill embedding, its
+/// witness-logged trust score (ADR-312 tamper-evident inputs), its cost, and
+/// the last step at which it was observed active (drives freshness).
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct PeerProfile {
+    pub id: PeerId,
+    /// Capability embedding — dotted against the query to get query-dependent
+    /// competence.
+    pub skill: Vec<f32>,
+    /// Witnessed reliability in `[0, 1]`. MUST be a tamper-evident, witness-logged
+    /// input per ADR-326's capability-vector-integrity gate — a peer inflating
+    /// its own trust to attract traffic is a modeled attack, not assumed-honest.
+    pub trust: f32,
+    /// Relative resource cost of collaborating with this peer (`> 0`).
+    pub cost: f32,
+    /// Step index at which this peer was last observed contributing. Older ⇒
+    /// less fresh.
+    pub last_active_step: u64,
+}
+
+/// The resolved five-factor capability vector for one `(i, j, q)` triple.
+///
+/// This is exactly ADR-326's formula operands; [`CapabilityVector::score`] is
+/// the `C(i,j,q)` product.
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+pub struct CapabilityVector {
+    /// Query-dependent skill match in `[0, 1]`.
+    pub competence: f32,
+    /// Witnessed trust in `[0, 1]`.
+    pub trust: f32,
+    /// Selecting agent's local affinity to the peer in `[0, 1]`.
+    pub locality: f32,
+    /// Recency of the peer's last contribution in `[0, 1]`.
+    pub freshness: f32,
+    /// Resource cost (`> 0`); the divisor.
+    pub cost: f32,
+}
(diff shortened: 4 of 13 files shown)


ruvnet:
### Security review (Wave-3 auditor) — 1 HIGH in the causal gate

Strong crate overall: the `incorporate` choke point enforces integrity→quarantine→causal in order and writes to the graph only on the causal-Accept arm (node_count verified unchanged on every rejection path); quarantine is sound (integrity=immediate, 3 causal-rejections quarantine, explicit reinstate, refuses dead-end churn from a quarantined `triggered_by`); `select_peer` takes no central-planner argument and there is no hidden global state (fully decentralized); the cost floor (`self.cost.max(1e-6)`) is robust to zero/negative/NaN so no infinite score; and the continue-vs-restart choice is loudly labeled unverified/swappable. Workspace membership is clean (no orphan), and all 8 shipped tests pass.

**HIGH — a non-finite (NaN/±inf) latent bypasses the causal-attribution gate and is written to the graph as a genuinely-attributed step, with no quarantine consequence; one such node then disables the specificity check map-wide.**

`ControlledReplacementVerifier::verify` (causal.rs) rejects via `attribution < accept_threshold` and `attribution - control_attribution < specificity_margin`. `cosine` (vecmath.rs) returns `NaN` for a non-finite latent (`norm` is `NaN`, `NaN == 0.0` is false, and `NaN.clamp(-1,1)` stays `NaN`). Every `<` comparison against `NaN` is **false**, so *both* rejection conditions are skipped and `verify` falls through to `Accept`. Nothing upstream validates finiteness — `integrity_tag` hashes `v.to_bits()` (a NaN payload gets a valid tag; and the tag is keyless FNV-1a, so a malicious *sender* stamps its own), quarantine only sees the message after it's already Accepted, and `add_step` stores the latent verbatim.

Reproduced against the crate as-is (adversarial test, then removed):
```
NaN outcome = Accepted { step: StepId(2) }      // orthogonal-but-NaN latent claiming antecedent A
node_count 2 -> 3                               // poison node written to the shared map
sender strikes = 0                              // NOT quarantined — repeatable indefinitely
```
Cascade — a single poison node then defeats specificity for everyone:
```
pre-poison  generic (spurious) msg -> Rejected(Causal(Spurious))   // correct
post-poison generic (spurious) msg -> Accepted { step: StepId(4) } // control mean now contains NaN
```
Because `control_signatures` includes the NaN node and `mean` propagates `NaN`, `control_attribution` becomes `NaN` for *every* later verification, so `attribution - NaN < margin` is always false — the gate silently degrades from "attributable AND specific" to "attributable only," admitting generically-attributed messages it used to block. This is the exact property WP23 rests on ("every incorporated message is causally attributable"), and the keyless integrity stand-in means any sender can do it.

**Fix:** reject non-finite latents at the choke point before the causal Accept — e.g. in `incorporate`, `if !msg.latent.iter().all(|x| x.is_finite()) { record(msg.from, Anomaly::IntegrityFailure); return Rejected(...) }` (so repeat offenders quarantine). Defense-in-depth: mirror the existing zero-vector handling by having `cosine` treat a non-finite result as `0.0` (no evidence → Unattributable), guard `mean`, and have `add_step` refuse a non-finite latent so the map can never hold a poison node. Worth an acceptance test alongside the existing "failing causal verification is rejected" case, which currently only covers a finite orthogonal latent.

Scope items confirmed honest (not findings): keyless FNV-1a integrity (WP5 keyed MAC), advertised trust/skill relying on ADR-312 witness-logged inputs, `Anomaly::MapDegradation` defined but not yet wired to a recorder (deferred outcome-feedback path), in-process transport stub, real multi-agent execution + 9-benchmark eval.

ruvnet:
## HIGH fixed — non-finite latent NaN-bypass (SHA `148ee6ae0`)

Thanks for the catch. The root cause was exactly as described: a non-finite latent made `cosine` return NaN, both gate `<`-comparisons vs NaN evaluated false, and `verify` fell through to `Accept` — then `control_signatures → mean` propagated NaN and silently degraded the gate map-wide from "attributable AND specific" to "attributable only." HIGH → fix mapping:

| Audit requirement | Fix | Where |
|---|---|---|
| **PRIMARY** — reject non-finite before the causal gate, never write it | `incorporate` step 0 validates `is_finite(&msg.latent)` → `Rejected(RejectionKind::NonFinite)` before integrity/causal | `src/engine.rs` |
| **QUARANTINE** — strike a repeat offender; wire the unrecorded `MapDegradation` | non-finite injection records `Anomaly::MapDegradation` (weight 2) — the previously defined-but-never-recorded anomaly, now wired here | `src/engine.rs` |
| **DEFENSE — cosine** non-finite operand → 0.0, not NaN | guarded `na/nb.is_finite()` + final `sim.is_finite()` checks | `src/vecmath.rs` |
| **DEFENSE — mean/control** one bad value can't poison map-wide control | `mean` filters non-finite vectors | `src/vecmath.rs` |
| **DEFENSE — add_step** never store a non-finite latent | non-finite elements sanitized to 0.0 on store | `src/thoughtmap.rs` |

### Tests (+4 → 32 green: 21 unit + 11 acceptance)
- `nan_latent_is_rejected_not_accepted_and_strikes_sender` — NaN claiming a real antecedent → `Rejected(NonFinite)`, `node_count` unchanged, `strikes > 0`, and a repeat offender quarantines (no strikes=0 retry-forever).
- `infinite_latent_is_rejected` — +inf / -inf likewise.
- `nonfinite_injection_does_not_poison_mapwide_specificity_control` — **the CASCADE test**: a genuinely-spurious message is `Rejected(Spurious)` both *before* and *after* a rejected NaN attempt, proving the map-wide control (`mean`) wasn't poisoned.
- `nan_latent_never_accepted_at_the_gate` (unit) — even a *direct* `verify` call on a NaN/inf latent is not `Accept` (guarded cosine → 0.0 → `Unattributable`), so the gate can't silently degrade even if the choke point were bypassed.

`cargo test -p latentmesh-thoughtmap` 32/32, `cargo check` + `clippy` clean. Nothing else from the audit disturbed (incorporate order, quarantine trigger, decentralization, continue-vs-restart labeling all unchanged). Ready for re-verification of the NaN paths.

ruvnet:
### Re-verification (Wave-3 auditor) — HIGH RESOLVED ✅

Commit `148ee6ae` fully fixes the non-finite-latent bypass and its map-wide cascade. All four layers verified by logic review + rebuilding the crate and re-running my ORIGINAL audit repro against the fix:

```
# was: Accepted { step 2 }, node_count 2->3, strikes 0
NaN outcome now = Rejected(NonFinite)   node_count 2 -> 2   sender strikes = 2
# was: post-poison spurious message = Accepted
pre-poison  spurious = Rejected(Causal(Spurious))
post-poison spurious = Rejected(Causal(Spurious))   # control NOT poisoned
```

1. **PRIMARY** — `incorporate` rejects at step 0 (`if !vecmath::is_finite(&msg.latent)`), genuinely BEFORE integrity/quarantine/causal and before `add_step`; no path reaches the gate or the graph with a non-finite latent. `is_finite` is element-wise (`.all`), so a single NaN/±inf element in an otherwise-finite vector is caught (repro used `[NaN, 0.1, 0.0]`); NaN and both infinities all fail `f32::is_finite`. The antecedent side is covered too — `add_step` sanitization means the node latents the gate reads are always finite.
2. **QUARANTINE** — the rejection records `Anomaly::MapDegradation` (weight 2, the previously-defined-but-unwired anomaly, now wired). Confirmed: one injection → strikes 2; a second → ≥3 → quarantined. No more strikes=0-forever.
3. **DEFENSE-IN-DEPTH (the cascade regression)** — three independent guards make the map-wide poison impossible even granting a hypothetical layer-1 bypass: `cosine` returns `0.0` for any non-finite operand OR non-finite result (never NaN → the gate's `<`-comparisons behave, non-finite → Unattributable reject, proven by the new gate-level unit test hitting `verify` directly); `mean` filters non-finite vectors so one bad signature can't poison the control population; `add_step` sanitizes non-finite elements to `0.0` so the map never stores a poison node. My re-run of the exact original cascade confirms the spurious message stays `Rejected(Spurious)` after a rejected NaN injection.
4. **TESTS** — the +4 tests genuinely exercise each path (NaN→Rejected+struck+node_count-unchanged+repeat-quarantines; ±inf likewise; cascade→spurious-still-Spurious; direct-`verify` NaN/inf never accepted). Full suite 32/32 green, `cargo check`/`clippy` clean (only the pre-existing workspace-profile warning).

Previously-sound parts left undisturbed — the diff touches only `engine.rs` (step-0 insert; the integrity→quarantine→causal order is byte-identical below it), `vecmath.rs`, `thoughtmap.rs`, and `causal.rs`/`acceptance.rs` (tests). `capability.rs`, `quarantine.rs`, `topology.rs`, `transport.rs`, `lib.rs` are untouched → incorporate order, quarantine trigger, decentralization, and the continue-vs-restart labeling all unchanged.

**CLEAR TO MERGE** on this finding. (Minor, non-blocking: the finiteness check sits before the quarantine check, so an already-quarantined peer's non-finite message returns `Rejected(NonFinite)` + another strike rather than `Dampened` — both reject with no write, so it's a harmless label nuance.)

PR #902, opened 2026-08-22
feat(context): add scope-sharded vector index

ruvnet:
## Summary

Adds a persistent, scope-sharded vector index for capability-authorized `ruv://` context retrieval.

- physically isolates every exact authority/tenant/subject/collection/path scope in its own REDB-backed `VectorDB`
- performs exact namespace and segment-aware descendant selection before ANN access
- enforces top-K, scope-capacity, and fanout limits before search
- persists a hash-bound scope manifest and fails closed on malformed or renamed shards
- makes point insertion immutable/idempotent and supports exact-scope deletion
- changes the REDB connection pool to weak references so erased/recreated shards cannot reuse an unlinked database
- records the architecture and trust boundary in ADR-332

RVM remains the capability/policy authority. This crate accepts only an already-authorized structured scope and preserves physical retrieval isolation.

## Security invariants

1. A tenant search never calls another tenant's ANN index.
2. Prefix selection is by path segments, not text.
3. Fanout refusal happens before the first ANN call.
4. Scope strings never become filesystem paths.
5. Non-finite vectors and noncanonical deserialized scopes fail closed.
6. Recovery verifies the scope-to-filename binding.

## Verification

- `cargo test -p ruvector-context --locked`: 11 passed
- `cargo test -p ruvector-core --lib --locked`: 265 passed, 1 ignored
- `cargo clippy -p ruvector-core -p ruvector-context --all-targets --locked -- -D warnings`: passed
- `cargo fmt -p ruvector-core -p ruvector-context -- --check`: passed
- changed-file secret scan: passed
- `git diff --check`: passed

Repository-wide `cargo audit --no-fetch` reports the existing RUSTSEC-2026-0258 findings for `h2` 0.3.27/0.4.15 and warnings for `lru` 0.12.5/0.16.4. Neither `h2` nor `lru` is present in the normal `ruvector-context` dependency tree; this PR adds no new registry dependencies.

## Acceptance tests

The integration suite proves cross-tenant zero-touch behavior with per-shard ANN counters, sibling exclusion, restart recovery, pre-search fanout failure, immutable replay/conflict handling, non-finite rejection, erasure persistence, and erase/recreate freshness.

## Upstream refresh and repository-wide CI

The branch includes current `main` at `9030e31b`. Scoped context/core gates above pass after that refresh.

The repository-wide workflows currently retain baseline failures outside this PR's changed paths:

- new Rust 1.98 lints in generated Hailo code and existing `ruvector-mincut` sources
- existing repository-wide rustfmt drift beginning in `mcp-gate` and `ruvector-agent-memory`
- the already disclosed `h2` and `lru` advisories in unrelated workspace dependency trees

These failures are not suppressed or represented as context-integration regressions.


ruvnet:
Reviewing this now — three adversarial passes are still running against the six security invariants (scope/path handling, isolation+limits, and the `ruvector-core` pool change). Posting two concrete findings I've already confirmed at head `41d690f`, since one is a merge blocker.

---

### BLOCKER — ADR-332 is a duplicate number

This PR adds `docs/adr/ADR-332-SCOPE-SHARDED-CONTEXT-INDEX.md`, but **ADR-332 is already taken on `main`** by `ADR-332-rf-sensing-modality-router.md` (merged earlier today via #892, PIR Wave 4).

This branch already contains that merge — I verified all five Wave-4 merge commits are ancestors of `41d690f` — so the collision is live in the branch right now, not a stale-base artifact. Merging as-is puts two ADR-332 files on `main`.

The repo's own gate confirms it:

```
$ node scripts/adr-index.mjs --check
ADR duplicate-number check FAILED:
  - NEW duplicate 332: docs/adr/ADR-332-rf-sensing-modality-router.md,
                       docs/adr/ADR-332-SCOPE-SHARDED-CONTEXT-INDEX.md
$ echo $?
1
```

**Fix:** renumber to **ADR-334** (INDEX.md says "next available: 334"; 328–333 went to Wave 4), then regenerate with `node scripts/adr-index.mjs` and update the in-PR reference in the description. Per ADR-316's numbering hygiene, never reuse a number — and note ADR-322 is deliberately skipped, so don't reach for it.

Worth flagging that the filename also breaks convention: every other ADR uses lowercase-kebab (`ADR-331-pandora-pattern-voi-cost-aware-routing.md`), this one is `SCREAMING-KEBAB`.

---

### Systemic — the ADR gate exists, works, and is wired to nothing

The reason this collision shows all-green is that **`scripts/adr-index.mjs` is not invoked by any workflow**:

```
$ grep -rn "scripts/adr" .github/
(no matches)
```

So the duplicate-number check is a script someone has to remember to run. That's worth its own follow-up — a gate that isn't wired in is indistinguishable from no gate, and this PR is the proof. I'm happy to file it (or add the step) separately; it isn't this PR's job to fix.

---

### Verified accurate

Credit where due — I re-ran your claims at `41d690f` rather than taking them:

- `cargo test -p ruvector-context --locked` → **11 passed** (3 unit + 8 integration), matching the description exactly.
- The baseline-CI paragraph is honest. `Clippy + fmt` and the advisory jobs do fail on `main` independently of this PR; I established the same baseline yesterday using a docs-only PR that reproduced the identical failure set. I've opened #904 to clear the rustfmt half (23 files / 95 hunks, all PIR-era, **no overlap with this PR's files**), which should remove one line from your "baseline failures" list.

### Design notes from my own read of `scope.rs`

Two things are done right and I want them on the record so they don't get "simplified" later:

1. **`hash_component` length-prefixes every field** before hashing. That is what makes tenant `a` + subject `bc` hash differently from tenant `ab` + subject `c` — the field-injection class is closed *by construction*, not by escaping. Please keep the length prefix if that function is ever refactored.
2. **Both `Deserialize` impls route through `new()`** rather than deriving straight onto the validated struct. That closes the classic serde-bypasses-the-constructor hole, and the test at `scope.rs` asserting a `"tenant":"ACME"` payload is rejected proves it bites.

One latent hazard, not a bug today: `hash_component` and `shard_id` use `u16::try_from(len).unwrap_or(u16::MAX)`, so a component ≥ 64 KiB would saturate and make two distinct scopes hash identically. Unreachable now (authority ≤ 253, slug ≤ 63, segment ≤ 128, ≤ 32 segments — all validated first), but it is a silent-collision failure mode gated only by limits declared elsewhere in the file. Consider returning an error instead of saturating, so a future limit change can't quietly become a shard collision.

Full adversarial results to follow when the three passes report.

ruvnet:
## Review: request changes — two HIGH findings, both demonstrated by running code

Adversarial review at head `41d690f`. Two of the six security invariants do not hold as stated, and the `ruvector-core` change introduces a reproducible concurrency regression in a shared crate. The design is sound and the scope-handling half is genuinely well built — these are fixable, and worth fixing rather than reverting.

---

## HIGH-1 — the pool change regresses concurrent open in `ruvector-core` (~97% failure rate)

`crates/ruvector-core/src/storage.rs:106-124`. **I reproduced this myself, with a control.**

`VectorStorage` has no `Drop`, and the pool now holds only `Weak`. When the last `Arc` drops, `Weak::upgrade` returns `None` the moment the strong count hits zero — but redb's `Database::drop` (a write txn + fsync) and the `flock(LOCK_UN)` in the file backend's own `Drop` happen *after* that. The `DB_POOL` mutex does not cover the drop path, so a concurrent `VectorStorage::new(p)` takes the else-branch, calls `Database::create`, and hits the still-held lock.

A barrier-synchronized probe (one thread drops the last handle, one opens the same path, 200 iterations):

| revision | result |
|---|---|
| PR head `41d690f` | **194/200 failed** — `Database error: Database already open. Cannot acquire lock.` |
| `main` `9030e31b` | **0/200 failed** |

That is the exact error `storage.rs:49-51` says the pool exists to prevent. The window isn't microseconds — it's the whole fsync-bearing drop.

**The fix is still load-bearing, so please fix forward rather than reverting.** Reverting the diff and re-running your own `erased_scope_recreation_never_reuses_the_unlinked_database` fails (`left ["old","new"], right ["new"]`) — the Arc pool really does hand back a database pointing at an unlinked inode.

**Suggested fix:** give `VectorStorage` a `Drop` that takes the pool lock, and if it holds the last strong ref, removes the entry *and* drops the `Arc` while still under the guard — so the flock is released before the guard is, and a concurrent `new()` blocks on the mutex instead of racing. The `Arc<Database>` never escapes `VectorStorage` (private field, only `begin_read`/`begin_write` used), so its `Drop` is always the last dropper. A per-path guard would avoid serializing unrelated paths behind one global mutex during fsync.

**No test covers the new lifetime semantics**, which is why a ~97% regression ships green. Please add: (1) open → insert → drop → reopen, assert data intact; (2) the concurrent drop-vs-open probe above, asserting zero failures.

---

## HIGH-2 — `create_shard` adopts a pre-existing file, laundering cross-tenant data

`crates/ruvector-context/src/index.rs:298-313` (vs the binding check at `:100-107`).

`open()` verifies scope↔filename and raises `CorruptShard` on mismatch. **`create_shard` does not.** It calls `VectorDB::new` on the deterministic path, and per `vector_db.rs:104-150` that silently adopts an existing file's stored config *and all its stored vectors* — then blind-writes `MANIFEST_KEY` with the scope being inserted, without checking the file was new or that any existing manifest matched.

Demonstrated: an attacker builds a shard for `attacker/docs`, copies the file to `hash(victim/docs)` **while the index is already open** (so `open()`'s check never sees it). The next insert into `victim/docs` adopts it, and search returns `["attacker-planted-row", "victim-own-row"]`. Worse, because `create_shard` rewrites the manifest to the victim scope, the rows are **laundered** — after restart, `open()`'s binding check now *passes* and the victim scope permanently contains the attacker's vectors, with no `CorruptShard` anywhere. The baseline was verified first: the same file planted *before* `open()` is correctly rejected, so this is specifically a hole in the create path.

This sits inside the same "attacker who can write the data dir" model that invariant 6 exists to defend, so it defeats a stated protection rather than falling outside scope. Targeting is easy: `shard_id` is an **unkeyed** hash over non-secret identifiers, so any victim's filename is computable offline.

**Suggested fix:** refuse adoption in `create_shard` — create with `O_EXCL` semantics, or after opening require `MANIFEST_KEY` to be absent-on-a-provably-empty-db or exactly equal to the scope being created (reuse the check at `:105`). Consider HMAC-SHA256 with a per-deployment secret so filenames aren't guessable. Also: the rollback at `:306` unconditionally `remove_file`s on a `save_config_value` failure — with adoption in play, that destroys a pre-existing shard the call never created.

---

## MEDIUM — `erase_scope` does not reliably erase

`index.rs:249-268` — in-memory state is removed at `:257`, the unlink happens at `:266`.

**(a) Failed unlink still reports erased in-process.** With the file made immutable, `erase_scope` returns `Err`, but the scope is already gone from the map (`scope_count` 0, `scope_stats` None) while the file survives. On restart the data returns; in-process, the next insert re-adopts it via HIGH-2, yielding `["fresh-row", "pii-must-be-deleted"]`.

**(b) Two handles over one root — no attacker needed.** `DB_POOL` is process-global, so two `ScopedContextIndex` handles share one pooled `Database`. Handle A's `erase_scope` reports **success** and unlinks, but B still holds an `Arc` and keeps serving the erased rows; A's next insert upgrades the same `Weak` and gets the unlinked database back. Your `erased_scope_recreation_never_reuses_the_unlinked_database` passes only because it uses a single handle.

For any right-to-erasure use, "erase reported success" must mean the data is gone. **Suggested fix:** unlink first, remove in-memory state second, report success only once the file is gone; and either delete rows through the DB handle before unlinking, or have `open()` take an exclusive lock on the root so two handles can't coexist.

---

## Invariants that DO hold (verified adversarially, not taken on trust)

Credit where it's due — these were probed hard and held:

- **Segment-aware prefix selection.** 25 adversarial shapes (`a`/`ab`/`a-b`/`a.b`/`a_b`, `team`/`team-secret`/`teamsecret`, depth 1–4) plus a decoy tenant, compared against ground-truth descendancy — zero mismatches, zero cross-tenant hits.
- **Scope strings never become paths.** Filename is 64 hex chars + `.redb`. 38 hostile segments rejected; the survivors (`...`, `CON`, `NUL`) are harmless because they never reach a filename. Case-insensitive-FS aliasing is impossible (lowercase hex), verified on APFS.
- **No field/separator injection.** `hash_component`'s length-prefixing gave 12/12 distinct digests for the classic `a`+`bc` vs `ab`+`c` shapes. **Please keep the length prefix if that function is ever refactored** — it is what makes this structural rather than escaping-dependent.
- **Noncanonical deserialization fails closed.** Both manual `Deserialize` impls route through the validating `new()`; 11 hostile JSON payloads rejected. This is the classic serde-bypasses-the-constructor hole and you closed it deliberately.
- **Non-finite vectors fail closed** at both entry points, before any ANN access.

## Lower severity

- `load_shard` compares only `dimensions`/`distance_metric`; stored `hnsw_config`/`quantization` silently win, so the caller's configured search parameters may not be what runs.
- `open()` enforces `max_scopes` only *after* loading every shard on disk — a large planted data dir is fully parsed before the bound applies.
- `validate_path_segment` allows uppercase while authority/slugs are forced lowercase, so `/Project` and `/project` are distinct scopes. Not a break, but an ACL-normalization hazard for any caller that case-folds capability strings.
- **Worth stating in the ADR:** the manifest binds scope→*filename*, not scope→*content*. There is no MAC over shard contents, so anyone with data-dir write can author a well-formed shard for any scope. Readers will otherwise over-read invariant 6.

## Blast radius of the core change

`ruvector-collections/src/manager.rs:165-172` hands out cloned `Arc<RwLock<Collection>>` that escape the DashMap, so `delete_collection` can leave handles alive and recreate onto an unlinked DB — the same class this PR fixes for shards, meaning the core fix doesn't achieve its goal repo-wide. Separately, `crates/ruvector-graph/src/storage.rs:43,96-121` is a byte-for-byte copy of the *unfixed* pool; worth a follow-up so the two don't drift.

---

Also still outstanding from my earlier comment: **ADR-332 is a duplicate number** (rename to ADR-334). A third review pass on isolation/limits is still running; I'll add anything it finds.

Happy to re-review once these are addressed — the scope-handling design here is good work and I'd like to see it land.

ruvnet:
## Third pass: isolation & limits — plus a verdict on what the test suite actually proves

The retrieval logic held up under everything I threw at it. The findings are in lifecycle/durability, and in the evidence the suite provides.

---

## The most important thing in this review: the suite never exercises an ANN index

`tests/scoped_context.rs:22-35` sets `hnsw_config: None`, which takes the `else` branch at `ruvector-core/src/vector_db.rs:212` straight to `FlatIndex` — a brute-force scan. **So the headline invariant, "a tenant search never calls another tenant's ANN index," is tested with no ANN index anywhere in the process** — even though this crate's `Cargo.toml` deliberately enables core's `hnsw` feature.

I verified both halves myself. The good news: re-running the cross-tenant, restart, and single-vector cases with `hnsw_config: Some(HnswConfig::default())` **does pass**, so the claim is true. But the shipped suite is not what establishes it. Please parameterize over both index kinds.

**The counter is also weaker evidence than it looks.** `shard.searches.fetch_add` (`index.rs:208`) sits immediately before the only `shard.db.search` call. There are five `shard.db.*` touch sites — `get` (151), `insert` (158), `search` (209), `delete` (245), `len` (283) — and exactly one increments:

```
151:  shard.db.get(&point.id)?
158:  shard.db.insert(VectorEntry {
208:  shard.searches.fetch_add(1, Ordering::Relaxed);   <-- only here
209:  shard.db.search(SearchQuery {
245:  shard.db.delete(id)
283:  shard.db.len()?
```

So the counter proves "the search loop did not iterate the foreign shard" — close to tautological — and says nothing about the other four paths. Nothing structural stops a sixth ANN call site being added with no counter and the test staying green.

**Suggested fix, and it's a good one:** wrap `db` in a private newtype whose only ANN accessor increments the counter, so *"touched without counting" stops compiling*. That converts the invariant from test-enforced to type-enforced. The counter is also reset on every shard load (`index.rs:311,333`), so it can't establish anything across a restart.

---

## HIGH-3 — `erase_scope` resurrects erased data (independently confirms my earlier MEDIUM)

`index.rs:249-268` against the process-global pool at `ruvector-core/src/storage.rs:36-38`. **Two independent reviewers reached this from different directions**, which is why I'm raising it from MEDIUM.

With two `ScopedContextIndex` handles on one root: insert `"gdpr-erase-me"`, call `erase_scope` → returns `Ok(true)` and the file leaves disk. Re-insert `"fresh"`. Search now returns **`["gdpr-erase-me", "fresh"]`**. The second handle keeps the pooled `Database` alive, so the `Weak` still upgrades and the "erased" rows come back. The other handle can also still read the erased scope directly.

A deletion API that reports success while the data survives is the one failure mode a right-to-erasure feature cannot have. And `erased_scope_recreation_never_reuses_the_unlinked_database` (`tests/scoped_context.rs:163`) is *named* for this hazard but passes only because its single-handle setup happens to drive the refcount to zero.

**Cheapest high-value fix:** take an exclusive advisory lock on the root in `open()`, so a second handle fails loudly instead of silently sharing state. That closes M1 below at the same time.

## MEDIUM — two handles silently diverge (stale reads, no error)

Same root cause. Each handle builds its **own** in-memory index over shared storage. Handle B's insert lands in storage; handle A's `search` never sees it — while A's `scope_stats` reports the higher count. Demonstrated: A returns `["from-a"]` while `scope_stats().vectors == 2`. A retrieval system silently serving stale results is worse than one that fails.

## MEDIUM — `max_scopes` is a cross-tenant denial of service

`index.rs:140-144`. The scope budget is process-global with no per-tenant quota, so tenant `noisy` creating 4 scopes under `max_scopes: 4` makes tenant `victim`'s **first ever insert** fail with `ScopeCapacity { maximum: 4 }` — which also discloses the global limit. In a crate whose premise is tenant isolation, availability isn't isolated at all. Needs a per-`ContextNamespace` quota under the global ceiling.

## MEDIUM — one malformed file denies `open()` to every tenant

`index.rs:95-112` returns `Err` on the first shard that fails to load. A single junk file matching `is_shard_path` (64 hex + `.redb`) makes the whole multi-tenant index unopenable, persisting across restarts — one tenant's corrupt shard is total loss of service for all of them. Quarantine and skip unloadable shards; fail that scope's operations, not the whole open.

## MEDIUM — `open()` eagerly materializes every tenant

`index.rs:95-112` → `vector_db.rs:295-317` rebuilds each index from all persisted vectors. A 30-tenant × 50-vector directory took **407 ms** and loaded every tenant into memory before a single scope was queried. Startup and memory are O(total corpus), bounded only by `max_scopes`. Either open shards on demand with an LRU cap, or make this an explicit documented non-goal.

## LOW

- **`max_results: usize::MAX` panics** (`index.rs:29-41`, `:206`). `validate()` only rejects zero, so a plausible operator spelling of "unlimited" reaches `Vec::with_capacity(k)` → `capacity overflow`. This is the mirror image of the "0 means unlimited" trap the fail-closed checks were written to avoid. Contained (no lock poisoning, index still usable, `panic = "unwind"`), but a panic unwinding out of a `#![forbid(unsafe_code)]` boundary crate on caller-supplied input is a contract violation. Cap `max_results` in `validate()`. Even a merely large value (20M) reserves ~3.3 GiB of address space against a one-vector shard.
- **`CorruptShard` leaks the absolute path** (`index.rs:322-328`) — the filename is opaque, but the full path discloses deployment layout. Report the filename only.
- **`scope_stats` is an unauthenticated existence oracle** (`index.rs:271-286`): for *any* scope the caller can name, including another tenant's, it returns existence and exact vector count. Only `search` carries the "already authorized" caveat in its docs; `insert`, `delete_point`, `erase_scope`, and `scope_stats` carry none, and `lib.rs:1-5` reads as though the crate itself is the boundary. Put the caller-authorization requirement in the crate docs and on every method.
- **"Immutable" survives only while the point is present**: `delete_point` then re-insert with the same id and different bytes succeeds, so the id→bytes binding isn't durable. Defensible, but the wording oversells it.

## Held up under attack (verified, not assumed)

- **Fanout refusal genuinely precedes the first ANN call** — on the common path, the collection-root path (empty prefix matching everything), a 10-wide subtree, and a 20-deep chain. All refused with `ScopeFanout` and per-shard counters summed to **0**. No early-return, retry, or recovery path reaches ANN first.
- **Prefix selection is correct**, including the subtle part: lexicographic order on `Vec<String>` makes descendants contiguous immediately after the prefix, so `range(..) + take_while(starts_with)` cannot terminate early on a real descendant. `project-archive` correctly excluded from `project`.
- **Limits**: `k = 0` and `k = usize::MAX` both rejected at the `k > max_results` gate; no limit value means "unlimited"; `push_top_match` is a correct bounded selection.
- **`delete_point` is exact-scope** — deleting `id` from `["p"]` leaves the identically-named point in `["p","c"]` untouched.
- **Concurrency within one handle is sound**: 160 interleaved insert/search/erase/stats operations across 4 threads on one scope — zero errors, zero panics, no deadlock, in-memory state consistent with disk. Every map mutation is under the write guard and no `Arc<Shard>` escapes a guard. **All the lifecycle damage requires the second handle.**

---

## Summary across all three passes

Blocking: **HIGH-1** (core pool concurrency regression, 194/200 reproduced with a clean control on main), **HIGH-2** (`create_shard` adopts and launders a planted shard), **HIGH-3** (`erase_scope` resurrection), and the **duplicate ADR-332**.

The retrieval path — scope selection, fanout ordering, top-K, exact-scope delete, length-prefixed hashing, validating deserialization — is solid and I'd keep it as written. The gap is lifecycle and durability, plus a test suite that under-determines its own claims. An advisory lock in `open()` is the single highest-value fix: it closes HIGH-3 and the divergence MEDIUM together.

Happy to re-review on a new head SHA.

ruvnet:
## All findings addressed — pushed to `707162d8`

I've fixed the three HIGHs plus the MEDIUM/LOW set rather than asking you to. Every regression test was verified to **fail against the unfixed code before the fix was applied**, so none of them are vacuous. Two adversarial re-audits are running against the new SHA; I'll post their verdicts before merging.

### HIGH-1 — core pool concurrency

`VectorStorage` now has a `Drop` that clears the slot and drops the `Database` **while still holding the path guard**, so redb's fsync-bearing drop and the `flock` release complete before a racing `new()` can observe an empty slot. The `Weak` change is kept — it fixes a real data-correctness bug, and reverting it fails your own `erased_scope_recreation_never_reuses_the_unlinked_database`.

`DB_POOL` is now per-path (`HashMap<PathBuf, Arc<Mutex<Option<Weak<Database>>>>>`), so one path's fsync no longer blocks opens of unrelated paths, and dead entries are reaped instead of accumulating forever.

| | before | after |
|---|---|---|
| concurrent drop-vs-open | **194/200 failed** | **0/200** |

That 0/200 is from **my own independent probe**, not the fix author's.

### HIGH-2 — shard adoption / laundering

Creation refuses an existing file, then refuses **again after opening** if the database carries a manifest, is non-empty, or opened with unrequested dimensions/metric — closing the check→open window. Neither refusal unlinks, so a planted file survives to be quarantined rather than deleted, and the rollback `remove_file` is now reachable only when the call actually created the file.

### HIGH-3 — erase resurrection and silent divergence

`open()` takes a non-blocking exclusive advisory lock on the root, so a second handle fails loudly with `RootLocked` instead of silently sharing the pooled database. That closes the stale-read divergence too. Verified that a crashed process does not brick a root — the kernel releases `flock` on exit. (Used `fs2` rather than `std::fs::File::try_lock`, which needs Rust 1.89 against this workspace's 1.77 MSRV.)

### Also fixed

- `erase_scope` unlinks **before** mutating the catalog — no phantom erase over a surviving file
- A single unreadable shard-shaped file is **quarantined** and surfaced via `quarantined_shards()` instead of denying `open()` to every tenant
- `max_results` capped at `MAX_RESULT_LIMIT` (4096) so `usize::MAX` is refused by `validate()` rather than panicking out of a `#![forbid(unsafe_code)]` crate
- Errors report the shard **filename**, not an absolute path

### The two evidence problems

**The counter is now type-enforced.** `Shard` moved into a private module where `ann_search` is the only route to the ANN index and always increments — an uncounted traversal no longer compiles. It previously guarded one of five `db` touch sites.

**The suite now runs against both engines.** It set `hnsw_config: None` throughout, so "a tenant search never calls another tenant's ANN index" was tested with no ANN index in the process. Parameterized via an `engine_tests!` macro — 5 unit + 29 integration, all green under both. The claim does hold under HNSW.

### ADR

Renumbered **332 → 334** (`ADR-334-scope-sharded-context-index.md`, lowercase-kebab to match convention). It now also records that the manifest binds scope→**filename**, not scope→**content**, that there is no MAC over stored vectors — which is exactly why refusing adoption is load-bearing — and marks the global scope ceiling and eager-open as explicit non-goals.

The gate that would have caught the duplicate is now wired into CI (#905, merged); #904 and #906 cleared the workspace rustfmt drift and the last Rust 1.98 clippy denial, so this branch's red CI should now be limited to the inherited `lru`/`h2` advisories.

### Known limits, stated rather than buried

- The existence check before creation is inherently TOCTOU; the **post-open** manifest/emptiness/config check is what actually closes it. True `O_EXCL` would need `VectorDB::new` to expose the flag.
- **No MAC over shard contents.** A crafted shard whose stored manifest names the victim scope would still pass every check. Closing that needs a keyed authenticator — recorded in the ADR as a design boundary, not fixed here.
- `erase_scope` ordering relies on POSIX unlink-while-open; on Windows the still-open shard file would fail to unlink. No Windows CI covers this crate.

### Separate follow-ups (not this PR)

- `ruvector-collections/src/manager.rs` hands out cloned `Arc<RwLock<Collection>>` that escape the DashMap, so `delete_collection` can still leave a live handle across the unlink — the pool correctly refuses to resurrect a *dead* handle, but a genuinely *live* escaped one is still live. Needs a manager-side fix.
- `ruvector-graph/src/storage.rs:43` still holds the **old strong-`Arc` pool** with no eviction, so it carries the original unlinked-database bug.

PR #922, opened 2026-08-23
fix(sota-bench): hard_regression must compare against the baseline, not an absolute threshold (issue #920)

### LINKED ISSUE #920: flywheel: hard_regression gate is permanently stuck closed — 30 strictly-dominant candidates rejected across 30 unattended generations
Found by running the PIR Wave-5 acceptance test: 30 unattended optimization generations (120 candidate experiments) via `ruvector-metaharness flywheel --generations 30`.

**Result: 120/120 candidates REJECTED, zero promotions, final policy identical to the starting policy.** The cause is not "no improvement was available." It is a gate that cannot open.

## Root cause

`crates/ruvector-sota-bench/harness/src/metrics.ts:83-86` computes regressions **absolutely**:

```ts
const regressions: string[] = [];
if (recallAt10 < 0.95) regressions.push("recall_below_0.95");
if (qps < BASELINE_QPS * 0.8) regressions.push("qps_below_baseline_floor");
```

`flywheel.ts:180` collapses that to a boolean, and `flywheel.ts:106` gates on the **candidate's flag alone**, with no comparison to the baseline:

```ts
if (candidate.regressed) reasons.push("hard_regression");
```

So on any workload whose absolute recall sits below 0.95, **the baseline is already `regressed = true`, therefore every candidate is too, therefore `hard_regression` fires unconditionally and nothing can ever be promoted.**

## Measured evidence from the run

| Check | Result |
|---|---|
| Candidates where baseline was **also** `regressed` | **120 / 120** |
| Candidates where **only the candidate** regressed | **0 / 120** |
| Max `recallAt10` reached anywhere | 0.936 (gate needs ≥ 0.95) |
| Candidates strictly better on primary **and** recall **and** qps **and** cost | **30 — all rejected** |

`hard_regression` never once identified a candidate-introduced regression. It fired 120 times and was correct zero times.

### The best rejected candidate

Mutation: adapt `m`. Rejected with `failureReasons: ['hard_regression']` — and it is **strictly dominant on every axis**:

| Metric | Baseline | Candidate | |
|---|---|---|---|
| primary | 0.8598 | 0.9593 | **+11.6%** |
| recallAt10 | 0.888 | 0.936 | better |
| qps | 1,437 | 4,222 | **2.9×** |
| p99 latency | 4,444 µs | 400 µs | **11× lower** |
| costPerWin | 22.65 | 0.69 | **33× lower** |
| memoryMb | 7.324 | 7.324 | unchanged |
| noopRate | 0.112 | 0.064 | better |

A configuration that is 33× cheaper per win, 11× lower latency, and *higher* recall was refused as a regression.

## Why this matters more than one bad run

The flywheel is designed to run **unattended**. In this state it burns compute indefinitely, produces checkpoints, emits plausible-looking reasons, and promotes nothing — it looks healthy while being inert. This is the same failure shape the program has hit twice before: a gate that appears configured and is actually a permanent no.

It also means any past "the flywheel found no improvement" conclusion drawn on a sub-0.95-recall workload is uninformative and should be re-run after the fix.

## Suggested fix

`hard_regression` should mean *"the candidate introduced a regression the baseline did not have"* — a **set difference**, not an absolute flag:

1. Carry the `regressions: string[]` list through onto the score rather than collapsing it to a boolean at `flywheel.ts:180`.
2. Fire `hard_regression` only when `candidate.regressions \ baseline.regressions` is non-empty.

This preserves the gate's real purpose (catch candidate-introduced regressions) while removing the permanent block. Keep the absolute thresholds — they are still useful as *reported* health signals; they just must not be promotion gates on their own.

**Note this loosens a safety gate**, so it needs the usual treatment: negative checks proving the fixed gate still blocks a genuinely-regressing candidate, and an adversarial review before merge. The other four reason codes observed in the run (`recall_regressed` 90, `paired_confirmation_inconclusive` 30, `resource_cost_worsened` 30, `pareto_dominated_by_baseline` 30) all compare against baseline correctly and are **not** affected.

Found during PIR Wave 5 (ADR-335, WP31 #914).

ruvnet:
Fixes #920. Part of epic #837.

## The bug

`aggregateReports` (`crates/ruvector-sota-bench/harness/src/metrics.ts:83-86`) itemises **absolute** health thresholds — `recallAt10 < 0.95` and `qps < BASELINE_QPS * 0.8`. `toScore` collapsed that list to a boolean, and `ruvectorPromotionRule` gated on the candidate's boolean alone with no comparison to the baseline:

```ts
if (candidate.regressed) reasons.push("hard_regression");
```

On any workload whose absolute recall sits below 0.95 the **baseline** already trips the threshold, so every candidate does too, so `hard_regression` fires unconditionally and nothing can ever be promoted. In the 30-generation run recorded in #920 it fired on **120/120** candidates and identified a candidate-introduced regression **zero** times — including on a candidate that was strictly better on primary, recall, qps, p99 and cost, and whose *only* rejection reason was `hard_regression`.

## The fix

1. `RuvectorScore` carries the itemised `regressions: string[]` alongside the existing boolean, which is left unchanged for the `@metaharness/flywheel` `Score` contract.
2. The gate fires only on the set difference `candidate.regressions \ baseline.regressions`, emitting one reason per new regression named as `hard_regression:<name>` so a rejection is diagnosable.
3. The absolute thresholds stay as reported health signals; they are simply no longer standalone promotion gates.

A score that lacks the itemised list falls back to failing closed on the boolean, so an evaluator that predates this change cannot silently open the gate.

## This loosens a safety gate

Everything else is untouched and was verified to already compare against the baseline:

| Reason code | Comparison | Touched |
|---|---|---|
| `paired_confirmation_*` | `pairedBootstrapDecision(baseline.primarySamples, candidate.primarySamples)` — paired per-item deltas | no |
| `recall_regressed` | `candidate.recallAt10 < baseline.recallAt10` | no |
| `resource_cost_worsened` | `candidate.costPerWin > baseline.costPerWin * 1.05` | no |
| `anchor_regressed` | `anchor.candidate < anchor.baseline` | no |
| veto reasons | absolute by design — pushed unconditionally, never diffed | no |

The Pareto gate, veto composition and the holdout/anchor disjointness assertion are unmodified.

## Tests

Eight assertions on `ruvectorPromotionRule` in `test/flywheel.test.ts`. Against `main` **5 of them fail**; all 8 pass after the fix. Full suite: 129 passed, 0 failed, 1 pre-existing opt-in skip (`RVF_ADAPTER_IT=1`).

| Test | Before | After |
|---|---|---|
| shared regression set → NOT fired (the stuck case) | fail | pass |
| candidate adds a regression → fired, names the new one | fail | pass |
| candidate has a *different* regression → fired | fail | pass |
| clean baseline, regressing candidate → fired, names both | fail | pass |
| both clean → NOT fired | pass | pass |
| score without the itemised list → fails closed on the boolean | pass | pass |
| vetoes stay absolute, not diffed against the baseline | pass | pass |
| the real #920 rejected candidate → no longer blocked by `hard_regression` | fail | pass |

The last test uses the genuine baseline/candidate scores (including per-item `primarySamples`) from the `adapt m` candidate in the #920 run, and asserts specifically on the absence of `hard_regression` rather than on promotion.

## Note on the #920 histogram

The reason `pareto_dominated_by_baseline` in the issue's histogram comes from `feat/pir-wp31-optimization-manifest`, which is not yet merged — that branch adds the in-repo Pareto gate and a `p99Us` field on the score. This PR is based on `main`, where that gate does not exist. The `hard_regression` defect is byte-identical on both branches, so this fix applies unchanged there.

## Verification

`npm ci --ignore-scripts`, `npm run check`, `npm test`, `npm run doctor`, and `cargo check -p ruvector-sota-bench --bin sota-all` all pass.

**Do not merge** pending adversarial review.

### DIFF
--- a/crates/ruvector-sota-bench/harness/src/flywheel.ts
+++ b/crates/ruvector-sota-bench/harness/src/flywheel.ts
@@ -55,6 +55,10 @@ interface RuvectorScore extends Score {
   qps: number;
   memoryMb: number;
   peakRssBytes: number;
+  /** The absolute health thresholds this score tripped, itemised. The gate diffs
+   *  candidate against baseline over these; the inherited `regressed` boolean
+   *  cannot express "already true of the baseline". */
+  regressions: string[];
   vetoReasons: string[];
   observations: BenchmarkObservation[];
 }
@@ -76,6 +80,30 @@ function asRuvectorScore(score: Score): RuvectorScore {
   return score as RuvectorScore;
 }
 
+/**
+ * The regressions the candidate introduces that the baseline did not already
+ * carry, as `hard_regression:<name>` reasons.
+ *
+ * `regressions` itemises ABSOLUTE health thresholds (the 0.95 recall floor, the
+ * qps floor), so on any workload sitting under one of them the baseline trips it
+ * too. Gating on the candidate's own flag therefore made `hard_regression` a
+ * permanent no — it fired on 120/120 candidates in issue #920 and identified a
+ * candidate-introduced regression zero times. Only the set difference is
+ * attributable to the candidate; the absolute list stays a reported signal.
+ *
+ * A score that predates the itemised list fails closed on the boolean rather
+ * than opening the gate, and an unknown baseline counts every candidate
+ * regression as new.
+ */
+function candidateIntroducedRegressions(baseline: RuvectorScore, candidate: RuvectorScore): string[] {
+  if (!Array.isArray(candidate.regressions)) return candidate.regressed ? ["hard_regression"] : [];
+  const inherited = new Set(Array.isArray(baseline.regressions) ? baseline.regressions : []);
+  return [...new Set(candidate.regressions)]
+    .filter((name) => !inherited.has(name))
+    .sort()
+    .map((name) => `hard_regression:${name}`);
+}
+
 export function ruvectorPromotionRule(evidence: {
   baseline: Score;
   candidate: Score;
@@ -88,7 +116,7 @@ export function ruvectorPromotionRule(evidence: {
   if (decision.outcome !== "pass") reasons.push(`paired_confirmation_${decision.outcome}`);
   if (candidate.recallAt10 < baseline.recallAt10) reasons.push("recall_regressed");
   if (candidate.costPerWin > baseline.costPerWin * 1.05) reasons.push("resource_cost_worsened");
-  if (candidate.regressed) reasons.push("hard_regression");
+  reasons.push(...candidateIntroducedRegressions(baseline, candidate));
   reasons.push(...candidate.vetoReasons);
   if (evidence.anchor && evidence.anchor.candidate < evidence.anchor.baseline) reasons.push("anchor_regressed");
   return { promote: reasons.length === 0, reasons };
@@ -149,6 +177,7 @@ async function toScore(
     memoryMb: aggregate.memoryMb,
     peakRssBytes: Math.max(...observations.map((observation) => observation.resources.processPeakRssBytes)),
     costPerWin: aggregate.memoryMb * Math.max(aggregate.p99Us, 1) / Math.max(aggregate.qps, 1),
+    regressions: aggregate.regressions,
     regressed: aggregate.regressions.length > 0 || vetoReasons.length > 0,
     vetoReasons,
     observations,
--- a/crates/ruvector-sota-bench/harness/test/flywheel.test.ts
+++ b/crates/ruvector-sota-bench/harness/test/flywheel.test.ts
@@ -3,7 +3,126 @@ import { mkdtemp, readFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import test from "node:test";
-import { runRuvectorFlywheel } from "../src/flywheel.js";
+import type { Score } from "@metaharness/flywheel";
+import { runRuvectorFlywheel, ruvectorPromotionRule } from "../src/flywheel.js";
+
+/** Build a gate-visible score. Defaults keep every gate other than
+ *  `hard_regression` silent, so a reason list is attributable to one check. */
+function gateScore(overrides: {
+  primarySamples?: number[];
+  recallAt10?: number;
+  costPerWin?: number;
+  regressions?: string[];
+  vetoReasons?: string[];
+}): Score {
+  const primarySamples = overrides.primarySamples ?? [0.9, 0.9, 0.9, 0.9, 0.9];
+  const recallAt10 = overrides.recallAt10 ?? 0.9;
+  const regressions = overrides.regressions ?? [];
+  const vetoReasons = overrides.vetoReasons ?? [];
+  return {
+    primary: primarySamples[0]!,
+    primarySamples,
+    noopRate: 1 - recallAt10,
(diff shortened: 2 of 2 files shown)


ruvnet:
## Acceptance re-run: 30 unattended generations, 120 candidates

`cargo build --release -p ruvector-sota-bench --bin sota-all`, then `node crates/ruvector-sota-bench/harness/dist/src/cli.js flywheel --generations 30 --output /tmp/fw-after` on this branch.

| | Before (#920 run) | After (this PR) |
|---|---|---|
| Generations | 30 | 30 |
| Candidates | 120 | 120 |
| **Promotions** | **0** | **1** |
| Verdicts | 120 REJECTED | 119 REJECTED, 1 PROMOTED |
| `verified_improvements` | 0 | 1 |
| `anchor_surviving_improvements` | 0 | 1 |
| Final policy | identical to root | `m` 16 → 24 |
| `replay_verified` | true | true |

### Block-reason histogram

| Reason | Before | After |
|---|---|---|
| `hard_regression` | **120** | **0** |
| `recall_regressed` | 90 | 59 |
| `resource_cost_worsened` | 30 | 87 |
| `paired_confirmation_inconclusive` | 30 | 61 |
| `paired_confirmation_fail` | 0 | 58 |
| `pareto_dominated_by_baseline` | 30 | n/a (gate lives on the unmerged WP31 branch) |

### `hard_regression: 0` is a true negative, not a disabled gate

All 120 candidates carry the itemised `regressions` list, so the diff ran on every one:

| Regression sets | Count | Old gate | New gate |
|---|---|---|---|
| baseline and candidate both regressed (identical sets) | 91 | blocked | evaluated on merits |
| baseline regressed, candidate clean | 29 | blocked | evaluated on merits |
| **candidate introduced a regression the baseline lacked** | **0** | blocked | would block — nothing to fire on |
| neither regressed | 0 | — | — |

The old gate would have blocked 91 of these on the candidate's own flag alone, including the one that was promoted. No candidate in the run introduced a new regression, which is why the reason count is zero; the unit tests cover the four cases where it must still fire.

### The promotion

Generation 1, mutation `adapt m` — the same mutation the issue identified as the best rejected candidate, and previously rejected with `hard_regression` as its **sole** reason.

| Metric | Baseline | Candidate |
|---|---|---|
| primary | 0.93714162125 | 0.96209496125 (+2.66%) |
| recallAt10 | 0.8830 | 0.9410 |
| qps | 4863.59 | 4714.25 |
| costPerWin | 0.5475 | 0.5282 |
| `regressions` | `["recall_below_0.95"]` | `["recall_below_0.95"]` |

Identical regression sets, so `hard_regression` is correctly silent and the candidate is judged by the paired bootstrap, recall, cost and anchor gates — all of which it passes. Best relative improvement seen anywhere in the run was +3.29% over the root primary; that candidate was still **rejected** on `paired_confirmation_inconclusive` and `resource_cost_worsened`, so the remaining gates are demonstrably still biting.

### Caveat on cross-run comparison

The two runs measured different absolute baselines (root primary 0.8598 before vs 0.93714 after, driven mostly by baseline qps 1437 vs 4864 — machine-load variance between the runs). Absolute primary values are therefore not comparable across the two runs; the verdict counts and reason histograms within each run are.

Issue #932, opened 2026-08-26
Storage pools key on non-canonicalized PathBuf — two spellings of one file produce two Database::create calls and a redb lock error


ruvnet:
## Defect

Both redb connection pools — `ruvector-core/src/storage.rs` and `ruvector-graph/src/storage.rs` (the latter now on the Weak+Drop shape after #907) — key their pool map on a `PathBuf` that is absolutized (joined onto cwd when relative) but **never canonicalized**. Two spellings of the same file resolve to two slots:

- `a/b/../b/x.db` vs `a/b/x.db`
- a symlinked directory vs its target
- on macOS, differing case on a case-insensitive filesystem

Each slot independently calls `Database::create` against the same underlying file; the second open fails with redb's "Database already open. Cannot acquire lock" — the exact error the pool exists to prevent.

This was noted in #907's tail ("Related, pre-existing in both pools") and is split out here so it survives #907 closing. It predates #902 in `ruvector-core` as well.

## Suggested fix

Canonicalize before keying, in both pools:

1. The parent directory always exists by the time the key is formed (both pools `create_dir_all` it first), so `path.parent().canonicalize()?.join(file_name)` is total for the create-if-absent case — `Path::canonicalize` on the full path fails when the DB file doesn't exist yet, so canonicalize the parent, not the file.
2. Keep the traversal check on the *user-supplied* path (it exists to reject intent, and canonicalizing first would erase the evidence), and use the canonical form only as the pool key + `Database::create` argument.
3. Regression tests, same in both crates: open via `dir/x.db` and `dir/sub/../x.db` concurrently — must share one `Database` (no lock error, both handles see each other's writes); open via a symlinked dir and the real dir — same assertion.

Since the two pools are intentionally copies of one reference shape (ADR-340 invariant 3), fix them in the same PR so they don't drift again — or better, extract the pool into a small shared module both crates use, which #907 stopped short of doing.

Issue #931, opened 2026-08-26
Sweep: 91 panicky float sorts (`sort_by(partial_cmp().unwrap())`) across 27 crates — NaN-reachable in core query paths (ADR-340 invariant 1)


ruvnet:
## Defect class

The #901 fix (NaN candidate confidence panicking `route()`) closed one instance of a pattern the workspace has **91 occurrences of, across 27 crates**:

```rust
xs.sort_by(|a, b| a.partial_cmp(b).unwrap())
```

`partial_cmp` returns `None` when either operand is NaN, so every one of these sites panics the moment a NaN reaches it with 2+ elements. Per ADR-340 invariant 1, sorts over floats that originate outside the module's control must reject non-finite values at the boundary or use `total_cmp` (preferably both).

## Inventory

`grep -rn "sort_by(.*partial_cmp.*unwrap())" crates/*/src` (excluding `unwrap_or` variants): 91 hits. Affected crates:

`emergent-time`, `mcp-brain-server`, `ruvector-bench`, `ruvector-bounded-rag`, `ruvector-capgated`, `ruvector-consciousness`, `ruvector-core`, `ruvector-diskann`, `ruvector-entropy-ann`, `ruvector-graph`, `ruvector-graph-transformer`, `ruvector-hyperbolic-hnsw`, `ruvector-lsm-ann`, `ruvector-math`, `ruvector-matryoshka`, `ruvector-maxsim`, `ruvector-mincut-gated-transformer`, `ruvector-postgres`, `ruvector-query-cache`, `ruvector-sota-bench`, `ruvector-sparse-inference`, `ruvector-streaming-qng`, `ruvector-temporal-coherence`, `ruvector-tiny-dancer-core` (non-router files), `ruvllm`, `ruvllm-cli`, `ruvllm-wasm`

## Triage order — user-reachable first

Not all 91 are equal. The ones in **query paths fed by user-supplied vectors** are remotely triggerable: inserting or querying with a NaN component makes every distance NaN, and the first result sort panics the server. Highest priority:

- `ruvector-core/src/advanced_features/hybrid_search.rs:248` — result score sort
- `ruvector-core/src/advanced_features/diskann.rs:182,186,202,293,297,352` — distance sorts in search
- `ruvector-core/src/advanced_features/product_quantization.rs:257` — distance sort
- `ruvector-core/src/advanced_features/conformal_prediction.rs:148,316` — score quantiles
[...]

Issue #930, opened 2026-08-26
ruvector-delta-index: delete() of the entry point leaves search anchored to a phantom node — the rest of the graph becomes unreachable


ruvnet:
## Defect

`DeltaHnsw::delete()` (`crates/ruvector-delta-index/src/lib.rs`) blanks the node (`id = String::new()`, clears `vector` and `neighbors`) and removes it from every *other* node's neighbor list — but never touches `entry_point`. If the deleted node **is** the entry point, every subsequent search starts from a node whose id is empty, whose distance is defined as `f32::MAX`, and whose neighbor lists are empty.

## Verified failure (probe run on the post-#825 code)

```rust
let mut index = DeltaHnsw::new(4, DeltaHnswConfig::default());
index.insert("a", vec[1.0, 0.0, 0.0, 0.0]).unwrap();
index.insert("b", vec[0.0, 1.0, 0.0, 0.0]).unwrap();
index.delete("a").unwrap();               // "a" is the entry point
let results = index.search(&[1.0, 0.0, 0.0, 0.0], 2).unwrap();
```

Observed output:

```
result: id="" dist=340282350000000000000000000000000000000
```

Two distinct consequences, both worse than a stale row:

1. **Phantom results.** The deleted node is returned as a `SearchResult` with an empty id and `f32::MAX` distance — `search_layer` unconditionally seeds its result heap with the start node.
2. **Total recall loss.** Because `delete()` cleared the deleted node's own neighbor lists, the entry point now has no outgoing edges. `greedy_search`/`search_layer` cannot leave it, so **every live vector in the index becomes unreachable**. In the probe, "b" is simply gone: the search returns only the phantom.

The same dangling reference corrupts *inserts*: `connect_node` navigates from `entry_point`, so new nodes connect to nothing until one of them happens to be assigned a level higher than the dead entry point's.

## Suggested fix

In `delete()`, after blanking the node, check `entry_point`:

- if the deleted index is the entry point, re-seat it on any live node (e.g. scan for the highest-level node with a non-empty vector), or `None` if the index is now empty;
- have `search_layer` skip nodes with empty vectors when seeding/collecting results (defense-in-depth — tombstones should never be returnable);
- regression tests: delete-the-entry-point then search (must return only live ids, and must reach all live nodes), and delete-then-insert (new node must be reachable).

## Related, same file

[...]

Issue #928, opened 2026-08-25
Profile and shard core-and-rest compile path before the 240-minute cap


ruvnet:
## Finding

The current `core-and-rest` job reaches its 240-minute cap during compilation, before its relevant test workload is attributable. This is distinct from the excluded `ruvector-delta-index` deadlock tracked in #825 / PR #787.

- Severity: **Medium functionality**
- Operational impact: **High** — the required shard cannot provide regression signal
- Confidence: **High** for the observed compile-time bottleneck; root cause still unlocalized
- Estimated effort: **1–2 days**
- Expected impact: restore a bounded, attributable Rust test gate and prevent unrelated PRs from inheriting a four-hour cancellation.

## Confirmed evidence

In [PR #926 Workspace run 32702332738](https://github.com/ruvnet/RuVector/actions/runs/32702332738), 16 of 17 jobs succeeded. [Job 97356253593](https://github.com/ruvnet/RuVector/actions/runs/32702332738/job/97356253593) was cancelled at the cap while Cargo/Rustc were still compiling `ruvector-temporal-tensor-wasm`. No DeltaHnsw test ran.

The current workflow explicitly excludes `ruvector-delta-index` from both `core-platform` and `core-and-rest`, so PR #787 cannot by itself explain or clear this current failure. The causal correction is recorded on #825.

## Reproduction / measurement

1. Run the current `core-and-rest` job with timestamped Cargo build/test phases and per-package timing.
2. Record the last active Cargo/Rustc process at cancellation.
3. Separate dependency compilation, crate compilation, test compilation, and test execution time.
4. Repeat on a warm cache and a cold cache to avoid attributing cache behavior to test hangs.

## Acceptance test

- Identify the package or dependency phase consuming the active compiler budget.
- Split or cache the shard so an attributable `core-and-rest` run completes below 240 minutes.
- Report compile time separately from test time.
- Run the focused delta-index suite independently under the timeout policy.
- One clean PR run completes with the shard's intended tests executed; no gate is weakened or removed.

This issue intentionally proposes no CI-policy patch; workflow changes require maintainer review.

PR #933, opened 2026-08-26
Cleanup/hardening: fix five open correctness/security defects, adopt ADR-341 invariants

### LINKED ISSUE #825: ruvector-delta-index test_insert_and_search hangs indefinitely (DeltaHnsw insert/search)
## Summary

`ruvector-delta-index::tests::test_insert_and_search` (`crates/ruvector-delta-index/src/lib.rs:731`) hangs indefinitely rather than failing. It is not slow — it never terminates.

## Evidence

Observed on CI run [31675583441](https://github.com/ruvnet/ruvector/actions/runs/31675583441) (PR #822, `ci/split-core-platform`). The `core-platform` shard sat in a nextest SLOW loop on this single test for **3h 52m** until the job was cancelled at the 240-minute cap. No other test in the shard reported progress after it started.

## Reproduction / nondeterminism

The test builds its input from `random_vector(128)`, which uses an unseeded `rand::thread_rng()`:

```rust
fn random_vector(dim: usize) -> Vec<f32> {
    use rand::Rng;
    let mut rng = rand::thread_rng();
    (0..dim).map(|_| rng.gen()).collect()
}

#[test]
fn test_insert_and_search() {
    let mut index = DeltaHnsw::new(128, DeltaHnswConfig::default());
    for i in 0..100 {
        let vec = random_vector(128);
        index.insert(&format!("vec_{}", i), vec).unwrap();
    }
    assert_eq!(index.len(), 100);
    let query = random_vector(128);
    let results = index.search(&query, 10).unwrap();
    assert_eq!(results.len(), 10);
}
```

Because the vectors are unseeded, the hang is **nondeterministic** — it reproduces only for graph topologies that trigger the bug. This is consistent with the test having passed on earlier runs before the `core-platform` shard split surfaced it.

The hang is almost certainly an unterminated loop in `DeltaHnsw::insert` or `DeltaHnsw::search` — a candidate-set/visited-set traversal that fails to make progress (e.g. a neighbor list containing a cycle, or a `while` over a candidate heap that re-pushes an already-visited node).

## Impact on CI

Until this is fixed, a hung test burns the full 240-minute job budget. Two mitigations landed in PR #822:

1. `.config/nextest.toml` now sets `slow-timeout = { period = "120s", terminate-after = 5 }`, so any test is killed and reported as a failure after 10 minutes instead of running forever.
2. `ruvector-delta-index` is temporarily held out of CI tests entirely — removed from the `core-platform` shard's package list and kept in the `core-and-rest` catch-all's `--exclude` list — so the crate does not block the pipeline.

## Fix checklist

- [ ] Root-cause the non-terminating loop in `DeltaHnsw::insert` / `DeltaHnsw::search`
- [ ] Add a progress/iteration bound so a malformed graph fails loudly instead of hanging
- [ ] Seed the test RNG (or use a fixed fixture) so the case is deterministic and regression-testable
- [ ] Re-add `-p ruvector-delta-index` to the `core-platform` shard and drop its `--exclude` from `core-and-rest` in `.github/workflows/ci.yml`

### LINKED ISSUE #901: router: NaN candidate confidence panics the decision sort in route() (ungated path)
## Defect

`crates/ruvector-tiny-dancer-core/src/router.rs:134`:

```rust
decisions.sort_by(|a, b| b.confidence.partial_cmp(&a.confidence).unwrap())
```

`partial_cmp` returns `None` when either operand is NaN, so this **panics** whenever the model returns `Ok(NaN)` as a candidate confidence and there are 2 or more candidates to sort.

## Scope

- **Pre-existing** — predates PR #893 (PIR WP28); not introduced by it.
- Reachable only on the **ungated** routing path. The VoI gate added in #893 has its own non-finite choke point that rejects non-finite score/uncertainty before any comparison, so enabling that gate happens to prevent this — but the gate is opt-in and `None` by default, so the default configuration is the exposed one.

Found during the WP28 security audit while sweeping the non-finite dimension; deliberately left out of #893 to keep that PR's diff scoped to its own work package.

## Failure scenario

A model (or a corrupted/quantized model, or a feature-engineering divide-by-zero) emits a NaN confidence for one candidate. With a single candidate the sort is trivially fine; with two or more, `route()` panics rather than returning an error. For a library used as an inference-hot-path router, a panic is a materially worse outcome than a refusal — it takes down the caller instead of letting it fall back.

## Suggested fix

Reject non-finite candidate confidences at the same choke point the rest of the crate uses (return an error — consistent with how the VoI path already handles non-finite model output), or make the sort total with `sort_by(|a, b| b.confidence.total_cmp(&a.confidence))`. Preferring the explicit rejection keeps a degenerate model loud rather than silently ordering NaNs.

Either way, add a regression test with 2+ candidates where one confidence is NaN.

### LINKED ISSUE #903: harness: benchmark.ts canonical() sorts keys with localeCompare, making the benchmark cache key locale-dependent
## Defect

`crates/ruvector-sota-bench/harness/src/benchmark.ts` has its own private `canonical()` that sorts object keys with `a.localeCompare(b)`. Locale collation is **locale- and ICU-build-dependent**, so the same payload canonicalizes differently on two machines with different `LANG` (or on a `small-icu` vs `full-icu` Node build).

Demonstrated on the sibling encoder in `ai4aiBench.ts` before it was fixed: keys `{z_metric, ä_metric, a_metric}` order as `a, z, ä` under `LC_ALL=sv_SE.UTF-8` but `a, ä, z` under `en_US.UTF-8` — different bytes, different digest.

## Why it matters here

That `canonical()` feeds the **benchmark cache key** (`.metaharness/cache/<sha256>.json`, hashed over commit, policy, item, embedding space, env fingerprint, threads, limits, args). A locale-dependent cache key means two machines can disagree about whether a cached benchmark result applies — either a spurious miss (wasted re-run, harmless) or, more concerning, keys colliding/diverging in ways that make cache behaviour non-reproducible across a heterogeneous CI fleet and a developer laptop.

It is currently harmless only by accident of which key sets happen to be fixed and ASCII. That is safety by coincidence, not by construction.

## Fix

Replace with code-unit ordering, which is what RFC 8785 (JSON Canonicalization Scheme) specifies:

```js
.sort(([a], [b]) => (a < b ? -1 : a > b ? 1 : 0))
```

This is exactly the change already made to the `ai4aiBench.ts` encoder in #890 (PIR WP25), where it was verified to digest identically under `en_US`, `sv_SE`, `de_DE`, and `C`, with a regression test that stubs `Intl.Collator` with a reversed comparator to prove the digest ignores collation. Copy that test shape.

## Blast radius — the reason this was NOT folded into #890

Changing the ordering **changes the cache key**, invalidating existing `.metaharness/cache` entries repo-wide. That is a one-time recompute rather than a correctness problem, but it deserves its own PR with the invalidation called out, rather than riding along in a feature PR. Deliberately left out of #890 for that reason.

Found during the WP25 security audit.

### LINKED ISSUE #907: ruvector-graph: connection pool still holds strong Arcs with no eviction (unlinked-database reuse + unbounded growth)
`crates/ruvector-graph/src/storage.rs:43` still uses the pre-fix pool shape:

```rust
static DB_POOL: Lazy<Mutex<HashMap<PathBuf, Arc<Database>>>> = ...
```

with the `pool.get` → `Arc::clone` → `Database::create` → `pool.insert(path_buf, Arc::clone(&new_db))` block at lines ~93-118, **no `Drop` impl, and no removal from `DB_POOL` anywhere** (the `.remove(` hits at 179/257/270/351 are all redb table operations, not pool eviction).

## Two consequences, both inherited from the shape

**1. Unlinked-database reuse (data correctness).** Because the pool holds a *strong* `Arc`, a database stays alive after its file is deleted. A path that is erased and then recreated is handed back the old `Database`, which points at the unlinked inode — so "erased" rows reappear in the recreated store. This is exactly the bug that motivated moving `ruvector-core`'s pool to `Weak` in #902, demonstrated there by `erased_scope_recreation_never_reuses_the_unlinked_database` failing against the strong-`Arc` version.

**2. Unbounded growth.** Nothing ever removes an entry, so every distinct path leaks its `PathBuf` + `Arc<Database>` for the life of the process — and the strong ref means the whole `Database` is retained, not just an allocation header.

## Scope

**Not a regression, and not in #902's diff** — that PR touches only `ruvector-core/src/storage.rs` plus the new `ruvector-context` crate and ADR-334. Surfaced twice by the #902 security review as the same bug class living in a second copy of the pool.

## Suggested fix

Port the #902 approach: hold `Weak<Database>` and give the storage type a `Drop` that clears the slot and drops the `Database` **while still holding the pool guard**, so redb's fsync-bearing drop and `flock` release complete before a concurrent open can observe an empty slot. Note that the naive `Arc`→`Weak` swap *alone* reintroduces a ~97% concurrent-open failure rate ("Database already open") — the explicit `Drop` eviction is the part that makes it safe. `ruvector-core`'s per-path `HashMap<PathBuf, Arc<Mutex<Option<Weak<Database>>>>>` is the reference implementation.

Add the same two regression tests: open → insert → drop → reopen (data intact), and a barrier-synchronized concurrent drop-vs-open probe asserting zero failures.

## Related, pre-existing in both pools

Both key the pool on a **non-canonicalized** `PathBuf`, so two spellings of one file (`a/b/../b/x.db` vs `a/b/x.db`, or a symlink) resolve to two slots and two `Database::create` calls against the same file, producing a redb lock error. Verified present in `ruvector-core` at the pre-#902 base as well, so it predates that work — worth fixing wherever the pools are consolidated.

### LINKED ISSUE #908: ruvector-context: require_private_root checks the root's mode but not its owner
Follow-up from the PR #902 security review (LOW, non-blocking — #902 merged at `682e1c75`).

## Gap

`crates/ruvector-context/src/layout.rs`'s `require_private_root` verifies `mode & 0o077 == 0` but never compares the owner. **A 0700 directory owned by someone else passes the check.**

Today that fails closed by accident: the subsequent lock open returns `EACCES`. But the check runs only at `open()` and is never revalidated — a running handle keeps writing into a root that was `chmod`'d to 0777 behind its back (confirmed during review). So an attacker with write access on the **parent** of the index root can:

1. rename the real root away,
2. create a fresh root as 0700 (passing the check),
3.  before the writes land.

At that point the index is writing into a directory the attacker controls, and every substitution attack from PR #902's review rounds 1–4 is reachable again. Their reach is still bounded — shards are 0600 owned by the index uid, so vector *contents* stay unreadable — but they can unlink and replace entries.

## Fix — check the owner, not the parent

In `require_private_root`, after the mode test, on unix also require `metadata.uid() == geteuid()`.

**This has zero effect on which deployments are accepted.** Every root an owner check newly rejects already fails today with a bare `EACCES` — a foreign-owned root our uid could actually *use* would need permissive mode bits, which the existing check already refuses. So it rejects only already-broken cases, earlier and with a clear error.

The crate forbids `unsafe`, so `libc::geteuid()` is out — but `fs4` (already a direct dependency) pulls in `rustix 1.1.4`, so `rustix::process::geteuid()` adds **no new transitive dependency** and is already in `Cargo.lock`.

## Why not check the parent directory

Considered and rejected. A parent check needs sticky-bit special-casing (`/tmp` is 1777, and the sticky bit is what actually prevents non-owners renaming others' entries) and still rejects ordinary temp-dir and shared-service layouts. Once the owner is checked, a substituted root would have to be both owned by us *and* 0700 — which an attacker cannot produce — so the parent question largely dissolves.

## Severity: LOW

Requires attacker write access on the parent of the index root, which is a deployment misconfiguration, and the realistic harm is integrity/availability rather than confidentiality of vector data. Worth noting the renamed-away directory keeps mode 0700 and the victim's ownership, so the data is *hidden*, not lost — an operator can recover it.

ruvnet:
## Summary

One hardening pass over the open defect backlog: five issues fixed with regression tests, the underlying invariants recorded as **ADR-341**, and three new issues filed for defects found during the work (#930, #931, #932).

> **Numbering note:** this ADR was authored as ADR-340, but main allocated ADR-340 to the signed retrieval-receipt anchoring ADR (#949) first, so the merge commit renumbers it to **ADR-341** and fixes the two `cypher_exec.rs` comments that cited "ADR-340 invariant 1" for this ADR's NaN rule.

Closes #825, closes #901, closes #903, closes #907, closes #908.

### #825 — `ruvector-delta-index`: DeltaHnsw insert hang (root-caused: self-deadlock)
The hang was not an unterminated traversal: `connect_node`'s prune branch held a **write lock** on a neighbor node while `prune_neighbors` re-locked the **same node** through `self.distance()`. parking_lot locks are not reentrant, so the inserting thread deadlocked — but only when a neighbor list overflowed `m0`, which the unseeded test RNG made nondeterministic. Reproduced deterministically (seeded, 400 inserts, dim 8 — hung pre-fix, <1s post-fix).
- Neighbor's vector now comes from the held guard via a split borrow; `prune_neighbors` drops self-loops before measuring distances.
- `greedy_search` carries an explicit progress bound (strictly-decreasing distance ⇒ ≤ `nodes.len()` moves) so a malformed graph terminates loudly.
- Tests seeded; prune-heavy regression test added; **`ruvector-delta-index` rejoins the `core-platform` CI shard** (checklist item from #825) — and passed there in 7 minutes on this PR's runs.

### #901 — `ruvector-tiny-dancer-core`: NaN confidence panicking `route()`
Non-finite score/uncertainty now rejected at a single choke point covering **both** the gated and ungated paths, tripping the circuit breaker (matching the VoI path's contract); the decision sort uses `total_cmp` as defense-in-depth. Regression test poisons the model weights with NaN (the issue's corrupted-model scenario) and asserts error-not-panic.

### #907 — `ruvector-graph`: strong-Arc pool (unlinked-database reuse + unbounded growth)
Ported `ruvector-core`'s post-#902 reference shape wholesale: `Weak` per-path slots, `Drop`-under-the-slot-guard eviction, slot reaping. Regression tests: erased-path recreation observes an empty database; reopen-after-last-drop preserves data and reaps the pool entry; 40-iteration barrier-synchronized concurrent drop-vs-open probe, zero tolerated failures. The non-canonicalized-key tail note from #907 is split out as #932 so it survives this close.

### #908 — `ruvector-context`: `require_private_root` owner check
The root must now be owned by the process euid in addition to being 0700, via `rustix::process::geteuid()` (no `unsafe`, already in `Cargo.lock` through `fs4` — one line of lockfile churn, no new transitive dependency). As the issue argued, this rejects only roots that already failed later with a bare `EACCES`.

### #903 — harness: locale-dependent benchmark cache key
`canonical()` in `benchmark.ts` now sorts keys by code unit (RFC 8785), with the `Intl.Collator`-stub regression test copied from the #890 fix shape. The same defect in `research.ts` (feeding `sha256`/`embeddingSpaceId` digests) is fixed in the same way.

> **⚠ One-time cache invalidation (called out per the issue):** changing the ordering changes the benchmark cache key, so existing `.metaharness/cache` entries no longer match and benchmarks recompute once. This is the accepted cost of a locale-independent key and the reason this change ships in its own commit path rather than riding a feature PR.

## ADR-341 — Correctness-Hardening Invariants for Hot-Path Primitives

Records the six invariants these fixes instantiate so the defect classes stop recurring: total float orderings on untrusted values, no lock re-entry (read through the held guard), the Weak+Drop pool shape as the only sanctioned pool pattern, identity-not-just-mode filesystem trust checks, code-unit canonical encodings for anything hashed, and seeded RNGs in tests of nondeterministically-triggered behavior. `node scripts/adr-index.mjs --check` passes post-merge (372 ADR files, next available 342).

## Follow-up issues filed from this pass

- **#930** — `DeltaHnsw::delete()` of the entry point leaves search anchored to a phantom node; verified empirically: the whole graph becomes unreachable and search returns only `id=""`.
- **#931** — inventory of 91 remaining `sort_by(partial_cmp().unwrap())` sites across 27 crates, triaged user-reachable-first (ruvector-core query paths, ruvector-postgres backend-abort risk).
- **#932** — both storage pools key on non-canonicalized `PathBuf` (two spellings of one file ⇒ redb lock error), split out of #907.

## CI status (see PR comments for detail)

Green everywhere the diff reaches, including `core-platform` with delta-index restored. Two pre-existing, tracked reds inherited from main, both commented on this PR: `Tests (core-and-rest)` caps out at 240 min compiling (#928, maintainer CI-policy decision), and `Build Tiny Dancer linux-arm64-musl` (#900, toolchain drift — fix ready in #934).

## Validation

- `cargo test -p ruvector-tiny-dancer-core` — 49 passed
- `cargo test -p ruvector-context` — 44 passed
- `cargo test -p ruvector-delta-index` — 15 passed (pre-fix: hung; suite now <1s)
- `cargo test -p ruvector-graph` — 264 passed post-merge (both sides' changes together)
- `cargo clippy` over the four touched crates — clean; `cargo fmt --all --check` — clean
- harness: `npm test` — 167 pass / 0 fail
- `node scripts/adr-index.mjs --check` — OK

https://claude.ai/code/session_01L5ffi8NiKAQNabK3D5t2gK


ruvnet:
**CI note: `Build Tiny Dancer linux-arm64-musl` is red here, and the failure is not this PR's.**

It is the known toolchain drift from #900: the link of third-party `redb` fails with `unsupported linker arg: --fix-cortex-a53-843419` before any first-party code is reached. This PR surfaces it only because it touches `crates/ruvector-tiny-dancer-core/`, which re-triggers the path-filtered workflow that hadn't run since its last green pass on 2026-08-03.

Root cause is now pinned down empirically (details in the fix PR): stable **1.98.0** (2026-08-18) began emitting the Cortex-A53 erratum flag for `aarch64-unknown-linux-musl`, and **no zig release accepts it** — verified against zig 0.13.0, 0.14.1, 0.15.2, and 0.16.0. The failure reproduces with a minimal cdylib link on 1.98.0 + zig 0.13.0 and disappears on 1.97.0 with the same zig.

The fix is **#934** (pin `toolchain: "1.97.0"` beside the existing zig 0.13.0 pin), kept as the separate CI PR that #900 explicitly asked for rather than folded into this diff. Once #934 merges, this check goes green here on re-run; every other failure signal on this PR (the initial Rustfmt red) is already fixed on the current head.

---
_Generated by [Claude Code](https://claude.ai/code)_

ruvnet:
**CI note: `Tests (core-and-rest)` was cancelled at the 240-minute cap, and this failure is not this PR's.**

It is exactly the pre-existing #928 compile-size problem: the job ran 02:53→06:53 UTC and was killed at the cap during compilation, the same outcome #928 documents on PR #926's run 32702332738 (cancelled mid-compile of `ruvector-temporal-tensor-wasm`, 16/17 other jobs green) — filed the day before this PR existed. This PR's diff adds only test code and small fixes to crates in that shard, nothing that moves a four-hour compile wall.

Not re-running: the cap-out already reproduced on an independent PR (#926, per #928), so a re-run would spend another four runner-hours to confirm what is already confirmed. Not fixing here: #928 explicitly reserves the shard split/caching change for maintainer review as a CI-policy decision, so porting a workflow change into this PR would preempt that.

Signal coverage note: everything this PR actually changes **is** tested by green shards — `core-platform` (which now includes `ruvector-delta-index` again, passing in 7 minutes), `core-and-rest-heavy`, `core-and-rest-wasm`, both Clippy jobs, Rustfmt, and the harness job all pass on this head. The two remaining reds are both pre-existing and tracked: this one (#928) and Tiny Dancer arm64-musl (#900, fix in #934).

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #38, opened 2025-12-01
RFC: Deep Integration of ruvector-attention into ruvector-gnn


ruvnet:
# RFC: Deep Integration of ruvector-attention into ruvector-gnn

## Executive Summary

This RFC proposes a comprehensive integration of `ruvector-attention` capabilities into `ruvector-gnn`, combining advanced attention mechanisms with graph neural network operations to create a state-of-the-art hierarchical vector search system.

### ruvector-attention (Standalone Crate)

The attention crate provides a rich set of attention mechanisms:

| Component | Description | Status |
|-----------|-------------|--------|
| **Scaled Dot-Product** | Standard transformer attention | ✅ Complete |
| **Multi-Head Attention** | Parallel attention heads | ✅ Complete |
| **Edge-Featured Attention (GATv2)** | Graph attention with edge features | ✅ Complete |
| **Hyperbolic Attention** | Poincaré ball model attention | ✅ Complete |
| **Dual-Space Attention** | Euclidean + Hyperbolic combined | ✅ Complete |
| **Flash Attention** | Memory-efficient O(block_size) | ✅ Complete |
| **Linear Attention** | O(n) complexity attention | ✅ Complete |
| **MoE Attention** | Mixture of Experts routing | ✅ Complete |
| **GraphRoPE** | Rotary Position Embeddings for graphs | ✅ Complete |

### ruvector-gnn (Current Implementation)

The GNN crate has its own attention implementation:

| Component | Description | Limitation |
|-----------|-------------|------------|
| **MultiHeadAttention** | Basic MHA in `layer.rs` | Duplicates attention crate functionality |
| **RuvectorLayer** | GNN layer with attention | Uses internal MHA only |
| **differentiable_search** | Soft attention search | Basic cosine similarity only |
| **hierarchical_forward** | Layer-wise processing | No geometric attention support |

### 1. **Hyperbolic Space for Hierarchical Search** 🌐

HNSW graphs are inherently hierarchical. Hyperbolic geometry naturally captures tree-like structures with exponentially more space at boundaries.

```
Current: Euclidean cosine similarity only
Proposed: Poincaré distance for hierarchical relationships
         - Better parent-child similarity
         - Improved layer-wise search
         - ~15-30% better recall on hierarchical data
```

### 2. **Edge-Featured Graph Attention (GATv2)** 🔗

Current GNN layers ignore edge features. GATv2 integration enables:

```rust
// Current: Simple weighted aggregation
[...]

Issue #37, opened 2025-12-01
Security: 10 npm vulnerabilities detected in dependencies


ruvnet:
## Summary

Running `npm install @ruvector/gnn@0.1.19` reports **10 vulnerabilities (2 moderate, 8 high)** in the dependency tree.

#### axios <=0.30.1
- **Severity**: High
- **Location**: `node_modules/axios`
- **Issues**:
  - [GHSA-wf5p-g6vw-rhxx](https://github.com/advisories/GHSA-wf5p-g6vw-rhxx) - Cross-Site Request Forgery Vulnerability
  - [GHSA-4hjh-wcwx-xvwj](https://github.com/advisories/GHSA-4hjh-wcwx-xvwj) - DoS attack through lack of data size check
  - [GHSA-jr5f-v2jv-69x6](https://github.com/advisories/GHSA-jr5f-v2jv-69x6) - SSRF and Credential Leakage via Absolute URL
- **Dependency Chain**: `wasm-pack` → `binary-install` → `axios`

#### body-parser 2.2.0
- **Severity**: Moderate
- **Location**: `node_modules/body-parser`
- **Issue**: [GHSA-wqch-xfxh-vrr4](https://github.com/advisories/GHSA-wqch-xfxh-vrr4) - DoS when url encoding is used
- **Fix**: `npm audit fix`

#### esbuild <=0.24.2
- **Severity**: Moderate
- **Location**: `node_modules/vite/node_modules/esbuild`
- **Issue**: [GHSA-67mh-4wv8-2f99](https://github.com/advisories/GHSA-67mh-4wv8-2f99) - Development server request vulnerability
- **Dependency Chain**: `vitest` → `vite-node` → `vite` → `esbuild`

## Recommended Actions

1. **Quick fix (non-breaking)**:
   ```bash
   npm audit fix
   ```

2. **Full fix (may include breaking changes)**:
   ```bash
   npm audit fix --force
   ```

3. **Manual updates needed**:
   - Update `axios` to version >0.30.1 (requires updating `wasm-pack`)
   - Update `vite` and `vitest` to latest versions

## Environment

- Package: `@ruvector/gnn@0.1.19`
- Node.js: v18+
- Date: 2025-12-01

## Notes

These vulnerabilities are in **dev dependencies** and **build tools**, not in the core `@ruvector/gnn` runtime package. The native NAPI-RS bindings themselves are not affected.

PR #944, opened 2026-08-29
fix(hnsw): stop search() silently dropping stored points (#773)

### LINKED ISSUE #773: VectorDB.search() silently omits stored rows (~6-12%); db.get() still returns them, higher efSearch does not recover
## Summary

`VectorDB.search()` intermittently omits a stored row from its results. The row is still in the store — `db.get(id)` returns it, vector and metadata intact — so the index has lost the node rather than the write having failed.

It is **not** a breadth-of-search limit: it reproduces with `k` and `efSearch` far above the row count, and re-running the same query or raising `efSearch` from 256 to 1024 returns the same short result. Once a row is lost for a given store, it stays lost.

## Environment

| | |
|---|---|
| ruvector | 0.2.41 |
| node | v22.23.0 |
| npm | 10.9.8 |
| OS | Darwin 25.6.0 arm64 (macOS, Apple silicon) |
| native pkgs | `@ruvector/core`, `@ruvector/rvf-node-darwin-arm64`, etc. (default install) |

## Reproduction

Needs only `npm i ruvector` — no embedder, deterministic vectors.

```js
// repro.cjs — node repro.cjs
const { VectorDB } = require('ruvector');
const fs = require('node:fs');
const os = require('node:os');
const path = require('node:path');

const DIM = 384;
const ROWS = 3;
const TRIALS = 200;

const unitVector = (seed) => {
  let x = (seed * 2654435761) % 2 ** 31;
  const v = new Float32Array(DIM);
  let norm = 0;
  for (let i = 0; i < DIM; i += 1) {
    x = (x * 1103515245 + 12345) % 2 ** 31;
    v[i] = x / 2 ** 30 - 1;
    norm += v[i] * v[i];
  }
  norm = Math.sqrt(norm);
  for (let i = 0; i < DIM; i += 1) v[i] /= norm;
  return v;
};

(async () => {
  let short = 0;
  const examples = [];
  for (let t = 0; t < TRIALS; t += 1) {
    const storagePath = path.join(fs.mkdtempSync(path.join(os.tmpdir(), 'rv-')), 's.db');
    const db = new VectorDB({ dimensions: DIM, distanceMetric: 'cosine', storagePath });

    for (let i = 0; i < ROWS; i += 1) {
      await db.insert({ id: `m${i}`, vector: unitVector(t * 100 + i), metadata: { tag: 'x' } });
    }

    const len = await db.len();
    const hits = await db.search({ vector: unitVector(t * 100), k: 64, efSearch: 256 });

    if (hits.length !== len) {
      short += 1;
      const seen = new Set(hits.map((h) => h.id));
      const missing = [...Array(ROWS).keys()].map((i) => `m${i}`).filter((id) => !seen.has(id));
      const byId = await Promise.all(missing.map((id) => db.get(id)));
      const retry = await db.search({ vector: unitVector(t * 100), k: 64, efSearch: 1024 });
      examples.push({
        trial: t, len, returned: hits.length, missing,
        getWorks: byId.every(Boolean), retryReturned: retry.length,
      });
    }
  }
  console.log(`short results: ${short}/${TRIALS}`);
  for (const e of examples.slice(0, 5)) console.log(' ', JSON.stringify(e));
})().catch((e) => { console.error(e); process.exit(1); });
```

## Observed

```
short results: 12/200
  {"trial":7,"len":3,"returned":2,"missing":["m2"],"getWorks":true,"retryReturned":2}
  {"trial":51,"len":3,"returned":2,"missing":["m1"],"getWorks":true,"retryReturned":2}
  {"trial":83,"len":3,"returned":1,"missing":["m1","m2"],"getWorks":true,"retryReturned":1}
  {"trial":93,"len":3,"returned":2,"missing":["m2"],"getWorks":true,"retryReturned":2}
  {"trial":105,"len":3,"returned":2,"missing":["m2"],"getWorks":true,"retryReturned":2}
```

`len` is 3 in every case, `getWorks` is always `true`, and `retryReturned` (efSearch 1024) never recovers the row. Trial 83 lost **two of three** rows.

Expected: with `k=64` and only 3 rows stored, `search()` returns all 3.

## Rates measured

| vectors | rate |
|---|---|
| deterministic pseudo-random (above) | 12/200 (6%) |
| real sentence embeddings, `OnnxEmbedder` all-MiniLM-L6-v2, 384d | 3/25 and 2/25 (~8–12%) |

The rate appears higher with real embeddings, which are clustered, than with near-orthogonal random vectors — consistent with a neighbour-linking problem during index construction, though I have not read the index code.

## Repair

Deleting and re-inserting a lost row restores it: 4 of 4 cases recovered, all within a single pass. So the stored vector is fine and only the graph entry is wrong.

```js
const row = await db.get(id);
await db.delete(id);
await db.insert({ id, vector: row.vector, metadata: row.metadata });
```

Detection, if useful to anyone hitting this: probe each known id with its **own** vector. If a nearest-neighbour search for a vector does not return the row holding that exact vector, the index has lost it.

## Impact

Silent, and it survives every obvious check. `insert()` resolves successfully, `len()` counts the row, `get()` returns it — only search omits it. Any code that enumerates by searching therefore operates on a random subset while reporting success. In our case an audit-trail aggregation folded ~90% of its rows and a compaction step skipped the same ~10%, both reporting completion.

Our workaround is to keep a roster row listing every id and enumerate with `list()` + `get()` instead of search, plus a `verify()`/`reindex()` pair using the probe and repair above. That covers enumeration; `search()` itself is still lossy, which for a vector database is the part that matters.

Happy to test a patch or provide more traces.

vidaunited:
Fixes #773.

`VectorDB.search()` intermittently omitted a stored row while `db.get()` still returned it and `db.len()` still counted it, and raising `efSearch` never recovered it. That last detail is the tell: the point had no in-edges to traverse to, so no amount of extra breadth could reach it.

Root cause is **two** defects in graph construction in the vendored `patches/hnsw_rs`. Both are required — fixing either alone still leaves indexes broken (ablation below).

## 1. Symmetric edges written on the wrong layer

`reverse_update_neighborhood_simple` (hnsw.rs:1245):

```rust
let l_n = n_to_add.point_ref.p_id.0 as usize;   // the NEW POINT's top level
q_point_neighbours[l_n].push(Arc::new(n_to_add));
```

The backlink was stored at the new point's *top level* rather than at `l`, the layer the forward edge was just built on. Consequences:

- a point at level >= 1 receives **no layer-0 in-edges at all**, so a layer-0 traversal can only reach it when it happens to be the entry point;
- edges are written above the neighbour's own level, so level-0 points end up carrying layer-1 neighbour lists.

Graph dump of a failing three-row index (`search` returned `[0, 2]`, dropping 1):

```
origin_id=0 p_id=PointId(0,0) neighbours=[(0,[2]), (1,[1])]   <- level-0 point holding a layer-1 list
origin_id=2 p_id=PointId(0,1) neighbours=[(0,[0]), (1,[0])]
origin_id=1 p_id=PointId(1,0) neighbours=[(0,[0]), (1,[0])]   <- nothing in layer 0 points AT id 1
```

Node 1's edge to node 0 exists; the return edge landed on layer 1, so layer 0 can never arrive at node 1.

Fixed to `min(l, q.level)` — the layer the edge was built on, clamped so it is never written above either endpoint's own level. The clamp is needed because `search_layer` seeds its heap with the entry point unconditionally and can therefore hand back a `q` whose own level is below `l`.

## 2. `search_layer` short-circuited on an empty layer bucket

hnsw.rs:933:

```rust
if self.layer_indexed_points.points_by_layer.read()[layer as usize].is_empty() {
    return return_points;   // empty
}
```

`points_by_layer` buckets each point under its **top** level only (`generate_new_point` pushes to `points_by_layer[p_id.0]`), while a point of level L participates in the graph at every layer `0..=L`. An empty bucket therefore does not mean an empty layer. Once every point in a small index had level >= 1, layer 0's bucket was empty while layer 0 still carried edges — and the bail-out left the point being inserted with **no layer-0 neighbours whatsoever**.

The traversal below that check is seeded from `entry_point` and bounded by neighbour lists, so it is well defined regardless of the bucket: an entry point with no neighbours at that layer simply yields itself. Check removed.

## Measurements

20,000 independent three-row indexes, using the defaults the `ruvector` npm wrapper passes (m=32, efConstruction=200, k=64, efSearch=256):

| build | short results |
|---|---|
| baseline | 780/20000 |
| fix 1 only | 22/20000 |
| fix 2 only | 821/20000 (no improvement) |
| **both** | **0/20000** |

Recall improves rather than regressing, and self-probe loss (search each stored vector with itself) goes to zero:

| | points lost | recall@10 |
|---|---|---|
| n=1000 uniform | 1 -> 0 | 0.9930 -> 0.9940 |
| n=1000 clustered | 0 -> 0 | 0.9970 -> 0.9980 |
| n=5000 uniform | 12 -> 0 | 0.9445 -> 0.9565 |
| n=5000 clustered | 24 -> 0 | 0.8820 -> 0.9200 |

This also confirms the reporter's observation that clustered real embeddings fail at roughly twice the rate of near-orthogonal random vectors (24 vs 12).

## End-to-end verification

Ran the reproduction script from #773 unchanged, against a locally built `@ruvector/core` swapped into `node_modules`. The control was built through the identical pipeline with only the `hnsw.rs` change reverted, so the clean result cannot be an artifact of a local build differing from the published one:

```
fix reverted:   short results: 7/200
fix applied:    short results: 0/1000
```

## Tests

Adds `crates/ruvector-core/tests/hnsw_issue_773_regression.rs`:

- `issue_773_three_row_index_returns_every_row` — the exact shape from the issue, aggregated over 400 indexes
- `issue_773_every_point_retrieves_itself` — the detection probe from the issue over a 2,000-point index

Both were confirmed to **fail** against the unpatched vendored crate (9/400 short; 6/2000 unreachable), so they are not vacuous. Levels come from `StdRng::from_entropy()`, so each test aggregates enough independent indexes that passing with the defect present is ~1e-5.

`cargo test -p ruvector-core --release`: **494 passed, 0 failed**, including the pre-existing `hnsw_integration_test`.

## Note for reviewers

The vendored crate's own `#[cfg(test)]` modules do not compile — they use the rand 0.9 API (`rand::rng()`, `Uniform::new(..).unwrap()`) while the vendoring pins rand 0.8 for WASM compatibility. This is pre-existing and unrelated to this change (an unmodified `patches/hnsw_rs/src/hnsw.rs` produces the identical 25 errors under `cargo test --lib`), and it is why the regression test lives in `ruvector-core` rather than beside the code it guards.

Both defects most likely also exist upstream in `jean-pierreBoth/hnswlib-rs`, since `patches/hnsw_rs` was vendored for a rand downgrade rather than forked for behaviour. I have not checked their current `main`.

https://claude.ai/code/session_016Z5LMV7ZW9PFaG9fzvpuwu

### DIFF
--- a/crates/ruvector-core/src/index/hnsw.rs
+++ b/crates/ruvector-core/src/index/hnsw.rs
@@ -139,6 +139,24 @@ impl HnswIndex {
     }
 
     /// Get configuration
+    /// Structural violations of the layer invariants restored by the
+    /// ruvnet/RuVector#773 fix: `(edges above an endpoint's level, points
+    /// unreachable over layer-0 edges)`.
+    ///
+    /// Both are empty on a healthy index. Exposed so the regression suite can
+    /// assert the invariant EXHAUSTIVELY on any index, instead of sampling
+    /// `search()` results and arguing from probability — levels come from
+    /// `StdRng::from_entropy()`, so a sampled trial cannot be made
+    /// deterministic, while these hold for every draw.
+    #[doc(hidden)]
+    pub fn structural_violations(&self) -> (Vec<String>, Vec<usize>) {
+        let guard = self.inner.read();
+        (
+            guard.hnsw.edges_above_endpoint_level(),
+            guard.hnsw.points_unreachable_at_layer0(),
+        )
+    }
+
     pub fn config(&self) -> &HnswConfig {
         &self.config
     }
--- a/crates/ruvector-core/tests/hnsw_773_recall_distributions.rs
+++ b/crates/ruvector-core/tests/hnsw_773_recall_distributions.rs
@@ -1,0 +1,178 @@
+//! Recall + latency for ruvnet/RuVector#773, on BOTH data distributions.
+//!
+//! The #773 fix changes which layer a symmetric edge is written on and removes
+//! an early-return in `search_layer`, so the fair question is not only "are
+//! points still dropped" but "did graph quality or speed regress". A single
+//! distribution cannot answer it: clustered data stresses neighbour-list
+//! contention, uniform data stresses long-range connectivity, and a change can
+//! easily help one and hurt the other.
+//!
+//! Ground truth is exact brute force over the same vectors, so recall here is
+//! recall@k against the true nearest neighbours -- not agreement with another
+//! approximate index.
+//!
+//! ⚠️ Latency below is meaningful only within one build profile -- a debug run\n//! is several times slower than release. Compare like with like.\n//!\n//! Everything is deterministic: vectors come from a counter-based generator, so
+//! a number in this report can be reproduced exactly. Only the HNSW level draw
+//! stays random, which is the property under test.
+
+use ruvector_core::index::hnsw::HnswIndex;
+use ruvector_core::index::VectorIndex;
+use ruvector_core::types::{DistanceMetric, HnswConfig};
+use std::time::Instant;
+
+const DIMS: usize = 64;
+const ROWS: usize = 2_000;
+const QUERIES: usize = 100;
+const K: usize = 10;
+
+fn cfg() -> HnswConfig {
+    HnswConfig {
+        m: 32,
+        ef_construction: 200,
+        ef_search: 100,
+        max_elements: 10_000,
+    }
+}
+
+/// Counter-based deterministic float in [-1, 1).
+fn rnd(state: &mut u64) -> f32 {
+    *state = state
+        .wrapping_mul(6364136223846793005)
+        .wrapping_add(1442695040888963407);
+    let x = (*state >> 33) as u32;
+    (x as f32) / (1u32 << 30) as f32 - 1.0
+}
+
+fn normalise(mut v: Vec<f32>) -> Vec<f32> {
+    let n: f32 = v.iter().map(|x| x * x).sum::<f32>().sqrt();
+    if n > 0.0 {
+        for x in v.iter_mut() {
+            *x /= n;
+        }
+    }
+    v
+}
+
+/// Uniform on the sphere: stresses long-range connectivity.
+fn uniform_set(n: usize, seed: u64) -> Vec<Vec<f32>> {
+    let mut s = seed;
+    (0..n)
+        .map(|_| normalise((0..DIMS).map(|_| rnd(&mut s)).collect()))
+        .collect()
+}
+
+/// 20 tight clusters: stresses neighbour-list contention, where many points
+/// compete for the same slots and edge bookkeeping matters most.
+fn clustered_set(n: usize, seed: u64) -> Vec<Vec<f32>> {
+    let mut s = seed;
+    let centres: Vec<Vec<f32>> = (0..20)
+        .map(|_| normalise((0..DIMS).map(|_| rnd(&mut s)).collect()))
+        .collect();
+    (0..n)
+        .map(|i| {
+            let c = &centres[i % centres.len()];
+            normalise(c.iter().map(|x| x + 0.05 * rnd(&mut s)).collect())
+        })
+        .collect()
+}
+
+fn cosine(a: &[f32], b: &[f32]) -> f32 {
+    a.iter().zip(b).map(|(x, y)| x * y).sum()
+}
+
+/// Exact top-k by cosine similarity -- the ground truth recall is measured against.
+fn brute_force(data: &[Vec<f32>], q: &[f32], k: usize) -> Vec<usize> {
+    let mut scored: Vec<(usize, f32)> = data
+        .iter()
+        .enumerate()
+        .map(|(i, v)| (i, cosine(q, v)))
+        .collect();
+    scored.sort_by(|a, b| b.1.partial_cmp(&a.1).unwrap());
+    scored.into_iter().take(k).map(|(i, _)| i).collect()
+}
+
+struct Report {
+    recall: f64,
+    mean_us: f64,
+}
+
+fn measure(label: &str, data: &[Vec<f32>]) -> Report {
+    let mut index = HnswIndex::new(DIMS, DistanceMetric::Cosine, cfg()).expect("index");
+    for (i, v) in data.iter().enumerate() {
+        index.add(i.to_string(), v.clone()).expect("insert");
+    }
+    assert_eq!(
+        index.len(),
+        data.len(),
+        "{label}: the index lost rows before measuring"
+    );
+
+    let mut hit = 0usize;
+    let mut total = 0usize;
+    let mut elapsed = 0u128;
+    for qi in 0..QUERIES {
+        // Queries are drawn from the data's own distribution, not uniform noise
+        // against clustered data -- otherwise the clustered number measures the
+        // wrong question.
+        let q = &data[(qi * 7 + 3) % data.len()];
+        let truth = brute_force(data, q, K);
+        let t0 = Instant::now();
+        let hits = index.search(q, K).expect("search");
+        elapsed += t0.elapsed().as_micros();
+        let got: Vec<usize> = hits
+            .iter()
+            .filter_map(|h| h.id.parse::<usize>().ok())
+            .collect();
+        hit += got.iter().filter(|i| truth.contains(i)).count();
+        total += truth.len();
+    }
+    let r = Report {
+        recall: hit as f64 / total as f64,
+        mean_us: elapsed as f64 / QUERIES as f64,
+    };
+    println!(
+        "  {label:<10} recall@{K} = {:.4}   mean search = {:.1} us",
+        r.recall, r.mean_us
+    );
+    r
+}
+
+/// The floor is placed from MEASUREMENT, between the two populations, not
+/// guessed:
+///
+///   fixed build     5/5 independent level draws -> recall@10 exactly 1.0000
+///                   on BOTH distributions
+///   #773 defect     uniform 0.9910, clustered 0.9960
+///
+/// So 0.999 separates them. A looser floor (0.90 was the first draft) measures
+/// the difference and gates nothing -- the defect sailed through it. A floor
+/// that cannot distinguish the two populations it sits between is decoration.
+///
+/// ⚠️ The printed numbers are still the comparison ruvnet asked to retain; the
+/// floor only stops a silent collapse reaching main.
+const RECALL_FLOOR: f64 = 0.999;
+
+#[test]
+fn recall_and_latency_on_both_distributions() {
+    println!("\nHNSW #773 -- {ROWS} rows, {DIMS}d, k={K}, {QUERIES} queries, exact ground truth");
+    let uniform = measure("uniform", &uniform_set(ROWS, 0xA5A5_1234));
+    let clustered = measure("clustered", &clustered_set(ROWS, 0x5A5A_9876));
+
+    assert!(
+        uniform.recall >= RECALL_FLOOR,
+        "uniform recall@{K} collapsed to {:.4} (floor {RECALL_FLOOR})",
+        uniform.recall
+    );
+    assert!(
+        clustered.recall >= RECALL_FLOOR,
+        "clustered recall@{K} collapsed to {:.4} (floor {RECALL_FLOOR})",
+        clustered.recall
+    );
+
+    // Both distributions must be retained: a run that silently measured one of
+    // them would satisfy the floors above while answering half the question.
+    assert!(
+        uniform.mean_us > 0.0 && clustered.mean_us > 0.0,
+        "both distributions must actually have been measured"
+    );
+}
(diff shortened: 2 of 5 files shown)


vidaunited:
Follow-up with measurements I owe this PR, from preparing the same fix against upstream `hnswlib-rs` ([jean-pierreBoth/hnswlib-rs#38](https://github.com/jean-pierreBoth/hnswlib-rs/pull/38), issue [#37](https://github.com/jean-pierreBoth/hnswlib-rs/issues/37)). Two corrections and one caveat — none of them change the fix, but one of them qualifies the claim.

### 1. My original self-probe metric was too tight on dense data

The measurements in the PR description used `k=10, ef=64` for the "is this point retrievable by its own vector" probe. On tightly clustered data at scale that asks *"is this point in its own top 10"*, not *"is this point reachable"* — in one configuration it reported 479 orphans where widening to `k=200, ef=1024` recovered **478 of them at rank 0**. The tiny-index numbers (k=64 against 3 stored rows) and the end-to-end reproduction are unaffected, but the mid-size self-probe figures should be read as reachability-grade only at `k=200, ef=1024`.

### 2. Re-measured with a reachability-grade probe, the fix is broader than reported

Sequential insert, orphans = not returned by a search for its own vector at `k=200, ef=1024`:

| configuration | before | after |
|---|---|---|
| n=3000, M=16, dim=128, uniform | 15 | **0** |
| n=3000, M=16, dim=768, clustered | 37 | **0** |
| n=3000, M=8, dim=64, uniform | 102 | **0** |
| n=10000, M=16, dim=128, uniform | 116 | **0** |
| n=10000, M=32, dim=128, clustered | 43 | **0** |
| n=20000, M=16, dim=128, uniform | 306 | **0** |

### 3. Caveat: two dense-cluster configurations are not fixed, and one is worse

| configuration | before | after |
|---|---|---|
| n=3000, M=6, dim=64, clustered | 99 | **144** |
| n=20000, M=16, dim=128, clustered | 3241 | 3258 |

The second row matters for this repo specifically, because **`HnswConfig::default()` uses `m: 16`** (the npm wrapper passes `m: 32`, which is clean in every configuration I measured). At 20k tightly clustered vectors that regime is ~16% orphaned **before and after** this change — so this PR should be read as "fixes the reported loss", not "search loss is now impossible".

I believe this is a distinct upstream defect that this change *exposes* rather than causes: `reverse_update_neighborhood_simple` shrinks an over-full neighbour list by dropping the farthest entry rather than applying the diversity heuristic from `select_neighbours`. Before the fix, backlinks were scattered to each point's top level, so layer-0 lists stayed short and rarely reached the `2 * max_nb_connection` threshold; now they fill up and the naive shrink evicts the long-range edges, which fragments tightly clustered data. Reported upstream rather than patched here, since it is a real algorithmic change to the vendored crate.

Nothing above affects the two regression tests in this PR, the 494-test `ruvector-core` run, or the end-to-end result on the issue's own reproduction (7/200 short before, 0/1000 after).

jean-pierreBoth:
**Thanks a lot!**
Reverse edge was not in correct layer and the short_circuit in search layer must be removed as a consequence.
The by product is that it will improve graph projection in crate annembed.

vidaunited:
Thanks — and that is exactly it. Both halves are already in [jean-pierreBoth/hnswlib-rs#38](https://github.com/jean-pierreBoth/hnswlib-rs/pull/38): the reverse-edge layer fix in `reverse_update_neighborhood_simple`, and the `search_layer` short-circuit removal that follows from it (`src/hnsw.rs`, +24/-8).

The short-circuit had to go for the reason you would expect: `points_by_layer` buckets each point under its *top* level only, while a point of level L participates in the graph at every layer 0..=L. So an empty bucket does not mean an empty layer — once every point in a small index is drawn at level >= 1, layer 0's bucket is empty while layer 0 still carries edges. Returning early there handed `insert_slice` an empty heap, its `if !sorted_points.is_empty()` guard skipped the layer, and the point being inserted ended up with no layer-0 neighbours at all.

#38 also carries the regression test (`tests/orphan_nodes.rs`) and the probe harness (`examples/orphan_probe.rs`) behind the measurements in that thread.

Good to hear about `annembed` — I had not considered graph projection as a beneficiary. Happy to run the probe against an annembed-shaped workload if that would be useful.

ruvnet (commented):
Dream exact-head review at `52044ae2`.

Frozen hypothesis: the vendored HNSW repair eliminates reachable dropped-point behavior without recall, latency, native/WASM, or workspace regressions.

The root-cause analysis is corroborated by upstream issue #37 and open upstream PR #38, and the reported clustered/uniform results are promising. The upstream change is not yet merged or released, however, and all six exact-head workflow groups here require authorization, so none of the required native, WASM, regression, supply-chain, formatting, or Workspace gates executed.

Before promotion, make the structural invariant or its seed deterministic, retain both clustered and uniform recall/latency comparisons, prove native/WASM parity, and obtain a green Workspace run on this exact head. A probabilistic small-index trial alone is not a durable completeness receipt.

Verdict: **INCONCLUSIVE**.

vidaunited:
Thanks — the INCONCLUSIVE verdict was right, and the "probabilistic small-index trial alone is not a durable completeness receipt" point in particular. Pushed `82766a81` addressing three of the four; the fourth is not mine to clear.

### 1. Structural invariant, made deterministic

Replaced the statistical argument with an exhaustive one: **every stored point must be reachable from the entry point over layer-0 edges alone**. That is the invariant `search()` actually depends on — a point with no layer-0 in-edge is not "found late", it is unreachable, which is why #773 looked like silent data loss while `get()` and `len()` still saw the row.

It needs no seed control, which sidesteps the `StdRng::from_entropy()` problem rather than working around it: the property holds for *every* level draw, and it is exhaustive over the graph, so one index is a receipt rather than a probability. The old test's ~1e-5 false-pass argument is retained alongside it, not replaced.

Proven able to fail, by mutation rather than assertion-counting: restoring `l_n = new_point.p_id.0` gives

```
rows=128: 1 point(s) unreachable from the entry point over layer-0 edges
          -- search() would silently omit them: [113]
```

**Two things I got wrong, both caught by measurement:**

- A second candidate invariant — *"no edge stored on a layer an endpoint does not occupy"* — **fires on the patched, working graph**. A level-0 point legitimately carries layer-1 out-edges (7 in an 8-row index) while reachability stays clean. I have not established whether that is benign upstream behaviour, so it is surfaced via `structural_violations().0` for inspection and deliberately **not** asserted. A check that reddens correct code is as damaging as one that never fires.
- My first reachability implementation walked the **symmetric** closure, and the pre-fix mutant **survived** it. Every point writes its own forward layer-0 out-edges, so undirected reachability is satisfied even with the defect present. `search_layer` advances by reading the *current* node's neighbour list, so the traversal must be **directed**. Without mutating I would have shipped a green test asserting nothing.

The tests live in `crates/ruvector-core/tests/`, not the vendored crate: `patches/hnsw_rs` is a `[patch.crates-io]` target and not a workspace member, so `cargo test` never descends into it and no workflow references it — a test there would never run.

### 2. Recall and latency, both distributions, retained

The existing `benches/hnsw_search.rs` used `vector[j] = (i + j) * 0.1` — a straight line in 128-D, neither clustered nor uniform — and measured latency only, with no recall.

New harness measures **recall@10 against exact brute-force ground truth** on both: uniform on the sphere (long-range connectivity) and 20 tight clusters (neighbour-list contention, where edge bookkeeping matters most). Queries are drawn from each dataset's own distribution, or the clustered figure answers the wrong question. Release, 2000 rows, 64d, k=10, 100 queries:

| distribution | recall@10 | mean search |
|---|---|---|
| uniform | 1.0000 | 73.5 µs |
| clustered | 1.0000 | 28.3 µs |

The floor is placed **between two measured populations**, not guessed:

- fixed build: 5/5 independent level draws → exactly `1.0000` on both
- #773 defect: uniform `0.9910`, clustered `0.9960`

so `RECALL_FLOOR = 0.999` separates them. My first draft used `0.90`, which measured the difference and gated nothing — the defect passed it cleanly. It now fails with `uniform recall@10 collapsed to 0.9910 (floor 0.999)`.

### 3. Native/WASM parity — not applicable here

`ruvector-core`'s own feature table reads `hnsw = ["hnsw_rs"]  # HNSW indexing (not available in WASM due to mmap dependency)`, with WASM on `memory-only`. Only `ruvector-core` and `ruvector-graph` depend on `hnsw_rs`, and neither is a WASM target. The patched code path does not exist in a WASM build, so there is no parity to demonstrate — I would rather say that than produce a comparison that only looks like evidence. Happy to be corrected if there is a WASM path I have missed.

### 4. Green Workspace run on this exact head — blocked on authorization

`gh pr checks 944` still reports *no checks reported on the branch*, before and after this push. As your review notes, all six exact-head groups require authorization on a fork PR, so no native, WASM, regression, supply-chain, formatting or Workspace gate can run until a maintainer approves them. **Nothing I can do from this side changes that**, and I am not claiming any CI evidence I do not have — the numbers above are local (macOS, aarch64, release).

One formatting note: `rustfmt` reorders imports and touches eight unrelated hunks in `patches/hnsw_rs/src/hnsw.rs`, so I left that file unformatted to keep the diff reviewable — it is not a workspace member, so no fmt gate reaches it. `ruvector-core` is `cargo fmt --check` clean.

Upstream `hnswlib-rs#38` is still unmerged, so this remains a vendored repair; happy to rebase onto it whenever it lands.

PR #948, opened 2026-08-30
fix(graph): correct Cypher NOT precedence

### LINKED ISSUE #939: cypher parser: NOT binds tighter than comparison — `NOT n.age = 30` parses as `(NOT n.age) = 30`
## Defect

`ruvector-graph`'s Cypher parser gives `NOT` higher precedence than the comparison operators. In Cypher, `NOT` binds *looser* than comparison, so `NOT n.age = 30` must parse as `NOT (n.age = 30)`.

## Verified

Dumping `MatchClause.where_clause` for `MATCH (n) WHERE NOT n.age = 30 RETURN n` on `c6bb23c84`:

```
BinaryOp {
    left:  UnaryOp { op: Not, operand: Property { object: Variable("n"), property: "age" } },
    op:    Equal,
    right: Integer(30),
}
```

The `NOT` has captured only `n.age`, and the whole thing is an `Equal` comparing a boolean against `30`. Correct would be:

```
UnaryOp { op: Not, operand: BinaryOp { left: Property{..}, op: Equal, right: Integer(30) } }
```

## Consequence

Any executor evaluating this tree faithfully returns the wrong rows. `(NOT n.age)` evaluates to a boolean, comparing it to `30` is false for every node, so `WHERE NOT <comparison>` silently matches nothing. It does not error — it returns an empty result, which reads as "no matching rows" rather than "your query was misparsed".

`<>` is unaffected (`n.age <> 30` parses correctly), so the workaround is to use `<>` and De Morgan the rest by hand.

## Scope

This is in the shared parser (`crates/ruvector-graph/src/cypher/parser.rs`), so it affects every consumer of `parse_cypher`, not just the Node binding. Fixing it means moving `NOT` below the comparison level in the precedence climb, and checking the existing parser tests don't encode the current behaviour.

Found while implementing #879's `WHERE` evaluation (PR #938). Deliberately left out of that PR — a precedence change in a shared parser has a wider blast radius than the binding fix and deserves its own review. PR #938 documents the gap next to the test it had to drop.

## Related

- #879 — `query()` ignored `WHERE` entirely, so this was unreachable before now
- PR #938 — implements `WHERE` evaluation; `cypher_exec.rs` notes this precedence gap

ruvnet:
## Outcome

Cypher NOT now binds looser than comparison and tighter than AND. Queries such as `NOT n.age = 30` produce `NOT (n.age = 30)` instead of `(NOT n.age) = 30`.

## Invariants

1. Comparison expressions remain inside NOT.

2. NOT expressions remain inside AND.

3. Arithmetic unary minus retains its existing precedence.

## Validation

`cargo test -p ruvector-graph` passed 352 tests with 13 existing ignored tests.

`cargo clippy -p ruvector-graph --all-targets -- -D warnings` passed.

RuFlo diff risk reported low risk with score 0 of 100.

## Risk

The intentional behavior change affects AST shape for unparenthesized NOT expressions. Exact AST regression tests lock the Cypher precedence contract.

Fixes #939

### DIFF
--- a/crates/ruvector-graph/src/cypher/parser.rs
+++ b/crates/ruvector-graph/src/cypher/parser.rs
@@ -820,10 +820,10 @@ impl Parser {
     }
 
     fn parse_and(&mut self) -> ParseResult<Expression> {
-        let mut expr = self.parse_comparison()?;
+        let mut expr = self.parse_not()?;
 
         while self.match_token(&[TokenKind::And]) {
-            let right = self.parse_comparison()?;
+            let right = self.parse_not()?;
             expr = Expression::BinaryOp {
                 left: Box::new(expr),
                 op: BinaryOperator::And,
@@ -832,6 +832,18 @@ impl Parser {
         }
 
         Ok(expr)
+    }
+
+    fn parse_not(&mut self) -> ParseResult<Expression> {
+        if self.match_token(&[TokenKind::Not]) {
+            let operand = self.parse_not()?;
+            return Ok(Expression::UnaryOp {
+                op: UnaryOperator::Not,
+                operand: Box::new(operand),
+            });
+        }
+
+        self.parse_comparison()
     }
 
     fn parse_comparison(&mut self) -> ParseResult<Expression> {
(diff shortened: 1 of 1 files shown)


ruvnet:
Post-merge Dream review for exact head `df644e2a4050901fe8b75db9d2c512255b854f78`.

The parser change matches the current [Neo4j Cypher boolean precedence](https://neo4j.com/docs/cypher-manual/current/expressions/predicates/boolean-operators/): comparisons bind before `NOT`, which binds before `AND`. Focused validation reports 352 graph tests passing with 13 existing ignores, clippy passes, and exact-head graph-native, regression, supply-chain, fmt/clippy, and WASM-dedup workflows are green.

Portfolio verdict: **INCONCLUSIVE** rather than ACCEPT because required Workspace CI run 33321172176 cancelled its `core-and-rest` shard. The narrow AST semantics are supported; repository-wide promotion remains blocked until that shard completes on the exact candidate or merge head.

PR #958, opened 2026-09-02
fix(mcp-brain-server): OAuth support so claude.ai connectors can register

PogeystickJoe:
## What this is

**The problem.** The "Permit Brain" claude.ai connector (`https://ubuntu1.tail6b157c.ts.net/sse`) has never successfully connected. Every attempt fails with *"Couldn't register with Permit Brain's sign-in service... add an OAuth Client ID in the connector settings"* (claude.ai error refs `ofid_b644a31dfbf5d339`, `ofid_5d625b13463e61cc`).

**The fix.** `mcp-brain-server` has zero OAuth support — confirmed by grepping its full route table, nothing under `/.well-known/oauth-*` or `/register` exists. claude.ai's remote-connector setup performs OAuth discovery + dynamic client registration (RFC 8414 / RFC 7591) before it will offer to connect at all, so against a server with none of that, registration fails outright, before any query is ever attempted. This PR adds a minimal, additive OAuth layer.

**Why it matters.** Nobody has been able to use this brain from claude.ai at all, ever, via this connector — not a degraded experience, a complete non-start.

**Risk.** Low for existing functionality — the new routes are merged as a separate sub-router with their own state; nothing in the existing bearer-key auth path, `AppState`, or any existing route is touched. Compiles clean. Real risk is in the new surface itself (see Known limitations below) — this is a draft for review, not asserted production-ready.

---

## Evidence (verified directly, live, on the actual host — not inferred)

- **No OAuth routes exist**: `grep -n "oauth\|/register" crates/mcp-brain-server/src/routes.rs` → zero hits besides the word "authentication" in an unrelated string.
- **No HTTPS listener currently serves this hostname either**: `curl https://ubuntu1.tail6b157c.ts.net/...` fails to connect (`curl: (7)`), and `tailscale serve status` on the host returns "No serve config" (as of 2026-09-02, before this session enabled tailnet-only Serve as a diagnostic step).
- **Anthropic's own docs confirm** ([MCP connector docs](https://platform.claude.com/docs/en/agents-and-tools/mcp-connector), [custom connectors help](https://support.claude.com/en/articles/11175166-get-started-with-custom-connectors-using-remote-mcp)) that claude.ai's remote MCP connections originate from **Anthropic's cloud infrastructure**, not the user's browser/device, and that the target server "must be reachable over the public internet from Anthropic's IP ranges." A private/VPN/tailnet-only host — even one the user can reach fine themselves — will not connect. **This means the OAuth fix alone is not sufficient**; the host also needs genuine public reachability (Tailscale Funnel, not just `tailscale serve`). That's an infra decision, not a code change, and is *not* made in this PR — flagged in the companion issue for an explicit call, since it makes a private data store publicly reachable and shouldn't happen as a side effect of a code PR.
- **Compiles clean**: `cargo check -p mcp-brain-server` against this branch, fresh checkout of upstream `main` — 0 errors, 0 warnings, 1m21s.

## Known limitations (flagged for review, not resolved here)

1. Client/auth-code/token store is in-memory (`DashMap`), process-lifetime only. A restart clears every registered client and forces re-registration in claude.ai. Fine for a first draft; should move to the existing persistent store before relying on this long-term.
2. `verify_system_key()` (and the memories-API key check) is **not yet updated** to also accept an OAuth-issued access token — tokens are issued but nothing on the protected `/v1/*` endpoints consults them yet. Left out deliberately to avoid touching the existing, working auth path without your review of the approach first.
3. Requires the public-reachability infra change noted above to actually work end-to-end from claude.ai.

## Companion artifacts

- Issue with full repro/evidence: permitplace/permit-platform#705
- ADR with design + pseudocode: permitplace/permit-platform `docs/adr/ADR-179-permit-brain-oauth-connector.md`

Assigning to @ruvnet as a fix request per Mike — happy to adjust the approach (e.g. persistent store, wiring the token check into the existing auth path, PKCE requirements) to match your conventions elsewhere in this crate.

https://claude.ai/code/session_014J3L618BmhC1udYcqkXJur

### DIFF
--- a/crates/mcp-brain-server/Cargo.toml
+++ b/crates/mcp-brain-server/Cargo.toml
@@ -31,7 +31,7 @@ path = "src/bin/embedder.rs"
 
 [dependencies]
 # Web framework
-axum = { version = "0.7", features = ["json", "query"] }
+axum = { version = "0.7", features = ["json", "query", "form"] }
 tokio = { version = "1.41", features = ["full"] }
 tower = "0.5"
 tower-http = { version = "0.6", features = ["cors", "trace", "limit", "set-header", "compression-gzip"] }
--- a/crates/mcp-brain-server/src/lib.rs
+++ b/crates/mcp-brain-server/src/lib.rs
@@ -14,6 +14,7 @@ pub mod gist;
 pub mod graph;
 pub mod midstream;
 pub mod notify;
+pub mod oauth;
 pub mod optimizer;
 pub mod pipeline;
 pub mod pubmed;
--- a/crates/mcp-brain-server/src/oauth.rs
+++ b/crates/mcp-brain-server/src/oauth.rs
@@ -1,0 +1,378 @@
+//! Minimal OAuth 2.1 layer for the MCP connector flow (draft fix for the
+//! "Permit Brain" connector failing to register from claude.ai).
+//!
+//! Scope, deliberately kept small: claude.ai's remote MCP connector setup
+//! performs OAuth discovery (RFC 8414 / RFC 9728) and, when no manually
+//! configured Client ID is present, Dynamic Client Registration (RFC 7591)
+//! before it will offer to connect at all. This server had none of that,
+//! which is the entire failure — "Couldn't register with the sign-in
+//! service" is exactly what a client sees when discovery/DCR comes back
+//! empty or 404.
+//!
+//! This module does NOT introduce a new authority model. An OAuth access
+//! token issued here is only ever handed out after the caller has proven
+//! they already hold a valid `BRAIN_API_KEY` (or `BRAIN_SYSTEM_KEY`) via the
+//! existing constant-time check in `routes.rs::verify_system_key` /
+//! the equivalent memories-API key check. OAuth is a compliant *wrapper*
+//! around the existing pre-shared-key authority, not a parallel one — on
+//! purpose, so this doesn't expand what a connected client can do or touch
+//! the existing REST/API-key auth path at all.
+//!
+//! DRAFT: this compiles against the route/state shapes as read from
+//! `routes.rs` and `types.rs` at the time of writing, but has not been run
+//! against a live build of the full workspace (100+ crates) in the time
+//! available. Please review the storage choice (in-memory, process-lifetime
+//! only — see "Known limitations" below) and the token/session model before
+//! merging.
+
+use std::sync::Arc;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use axum::{
+    extract::{Query, State},
+    http::{HeaderMap, StatusCode},
+    response::{Html, IntoResponse, Redirect},
+    Json,
+};
+use dashmap::DashMap;
+use serde::{Deserialize, Serialize};
+
+const AUTH_CODE_TTL_SECS: u64 = 300; // 5 minutes, single use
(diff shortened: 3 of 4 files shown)


PogeystickJoe:
## Correction, filed same day — please read before reviewing "Fix 2" context

This PR's companion ADR originally presented public Tailscale Funnel exposure as a reasonable option alongside tailnet-only. That was wrong, filed without checking this repo's/permit-platform's prior brain-security history first.

Checked now: permitplace/permit-platform#609 and #626 confirmed a **real cross-client data leak** in this exact brain corpus (a query for an Austin, TX permit returned a hit carrying another client's real name/project/researcher metadata). The fix for that — `guardedBrainSearch()` — was built and merged, but it lives in the **Cloud Run agent's own code** (`cloud-run/permit-research-agent/lib/brain-guard.js` in permit-platform), not in `mcp-brain-server` at all. I checked this whole checkout for anything guard/scoping-related and found nothing.

**This PR's OAuth fix is still correct and needed** — the connector genuinely can't register without it, that part doesn't change. But once OAuth works, if the server is ever made publicly reachable (which it also needs to be, for claude.ai to reach it at all — separate infra decision, not part of this PR), a direct MCP connection would bypass the guard and hit the raw, unscoped corpus. Worth deciding, before merge or at least before any Funnel decision, whether an equivalent scoping/PII check should live inside `mcp-brain-server` itself so it applies regardless of which path reaches it (Cloud Run or direct MCP). Happy to take a pass at that too if useful, or defer entirely to your judgment on how it should fit alongside the existing guard design.

Full writeup: permitplace/permit-platform#706 (ADR-181) and permitplace/permit-platform#705 (issue, updated with the same correction).

ruvnet (commented):
Dream Machine security review — **REJECT**

Frozen hypothesis: this branch provides an OAuth 2.1-compatible Claude.ai connection path while preserving authorization integrity.

The reachable source contradicts that claim. Registered callback binding and PKCE are not enforced end-to-end, issued bearer tokens are not yet validated by the protected route, and all six exact-head workflows require authorization. That leaves both interoperability and the authorization boundary unverified. I am intentionally withholding reproduction details from this public review.

Before this can be made ready: handle the security-sensitive gaps through a private advisory/review path; add negative tests for callback binding, PKCE, client/code/token binding, replay, expiry, and audience; add a full register → authorize → token → protected-resource test; require native, WASM, supply-chain, regression, formatting, and Workspace CI on this exact head.

ruvnet:
Reviewed — good-faith, well-scoped work (this fixes the RFC 8414/7591 discovery+registration handshake, and is upfront that `is_valid_oauth_token()` isn't wired into the real auth path yet). Found one real security gap that should be closed before this goes further: `authorize_get`/`authorize_post` never validate the caller-supplied `redirect_uri` against the `redirect_uris` list the client registered via DCR (`register_client`) — a classic authorization-code-interception/open-redirect setup. It's not exploitable today only because nothing consumes the resulting token yet; it becomes exploitable the moment `is_valid_oauth_token` gets wired into `verify_system_key` as this PR's own follow-up proposes. Could you add a check comparing the incoming `redirect_uri` against `client.redirect_uris` in both handlers? Also noting: server metadata advertises `refresh_token` in `grant_types_supported` but `/token` only implements `authorization_code` — will break token refresh after the 30-day access-token TTL. Everything else (constant-time API-key comparison, correct S256 PKCE, HTML-escaped consent form) looks solid. Will merge once the redirect_uri allowlist check is in.

Issue #704, opened 2026-07-17
RVF SDK silently drops metadata; native metadata is not durable or returned


thesoulpole:
## Versions

- `@ruvector/rvf@0.2.3`
- `@ruvector/rvf-node@0.2.0`
- darwin-arm64 native binary `0.1.7`
- Node 26.4.0, macOS arm64

## Public SDK reproduction

```js
const db = await RvfDatabase.create(path, {
  dimensions: 2,
  metric: 'cosine',
}, 'node');

await db.ingestBatch([
  { id: 'red-item', vector: new Float32Array([1, 0]), metadata: { color: 'red' } },
  { id: 'blue-item', vector: new Float32Array([0, 1]), metadata: { color: 'blue' } },
]);
```

Observed:

- Ingest reports both vectors accepted.
- Unfiltered results contain no metadata.
- A field-0 equality filter for `red` returns no results.

## Native N-API reproduction

Using the low-level binding, pass one positional metadata record per vector:

```js
db.ingestBatch(
  new Float32Array([1, 0, 0, 1]),
  [101, 102],
  [
    { fieldId: 0, valueType: 'string', value: 'red' },
    { fieldId: 0, valueType: 'string', value: 'blue' },
  ],
);
```

A field-0 red filter works while that database instance remains open. However:

- Results still contain only ID and distance.
- After close/reopen, the same filter returns no results.

## Root cause

- `npm/packages/rvf/src/backend.ts` accepts `RvfIngestEntry.metadata` but calls native `ingestBatch(flat, ids)` without metadata.
- Native `SearchResult` contains only ID/distance.
- Runtime inserts metadata into an in-memory `MetadataStore`, but no metadata segment is written and reloaded.
- The SDK filter declaration omits the native parser's required `valueType`.

## Impact

Applications can successfully ingest records while silently losing metadata. Filtering may appear to work through the native API and then stop working after restart, which can produce incomplete or incorrect search results without an error.

## Requested fix

As an immediate safety measure, reject non-empty SDK metadata until it is supported rather than silently accepting it.

A complete fix needs:

- A per-vector metadata shape across the SDK and N-API.
- Metadata segment persistence and reload.
- Aligned filter types, including `valueType`.
- Either metadata-bearing search results or `getMetadata(id)`.
- Close/reopen tests, including vectors with unequal metadata-field counts.

ruvnet:
[...]

PR #980, opened 2026-09-11
fix(rvf): bump @ruvector/rvf pin ^0.1.0 → ^0.3.4 (string-id data loss in npx ruvector)

ruvnet:
## Problem

`npx ruvector rvf …` silently loses data with non-numeric ids. Building a store with string ids (e.g. `rel:chris-keil`) and then `status`/`query` shows only **1 vector** no matter how many were ingested, and `query` never matches.

**Root cause:** the CLI resolves `@ruvector/rvf` at **`^0.1.0` → 0.1.9**, whose NAPI backend does `ids = entries.map(e => Number(e.id))`. Non-numeric strings become `NaN`, so the native i64 HNSW collapses every vector onto one label. `status: totalVectors:1`, a single `query` hit at dist≈1, and `compact` shrinking the file are all *correct* readings of a 1-vector store — not separate bugs. 0.1.9 also silently dropped per-vector `metadata`.

## This is already fixed in `@ruvector/rvf@0.3.4` — the CLI just can't reach it

`0.3.4` (published) adds `resolveLabel` (a persisted string↔numeric-label map) and **rejects metadata loudly** instead of dropping it. But `^0.1.0` (and `rvf-mcp-server`'s `^0.1.2`) cannot resolve to `0.3.x`, so users stay on the broken `0.1.9`.

## Change

Bump the `@ruvector/rvf` pin `^0.1.0`/`^0.1.2` → `^0.3.4` in:
- `npm/packages/ruvector` (CLI)
- `npm/packages/rvf-mcp-server`

## Verification (against real `@ruvector/rvf@0.3.4`)

```
create(dim=4, cosine); ingestBatch([{id:'rel:chris',vector:[1,0,0,0]},{id:'rel:bo',[0,1,0,0]},{id:'org:cog',[0,0,1,0]}])
status()  -> { totalVectors: 3, ... }                         # was 1 on 0.1.9
query([1,0,0,0],3) -> [{id:"rel:chris",distance:0}, {id:"rel:bo",1}, {id:"org:cog",1}]   # string ids resolved, exact match dist 0
ingestBatch([{id:'x',vector:[0,0,0,1],metadata:{…}}]) -> throws "Per-vector metadata is not yet supported by this SDK (see #704)"  # loud, not silent
```

The CLI wrapper (`RvfDatabase.create/.open`, `ingestBatch`, `query`, `status`, `compact`) is API-compatible with 0.3.4 — a drop-in bump.

## Scope / not included

- **Republish is required** for `npx ruvector` and `@ruvector/rvf-mcp-server` users to receive this — a maintainer release step, not part of this PR.
- The monorepo `npm/package-lock.json` should be regenerated (`npm install`) so the resolved tree picks up 0.3.4; not regenerated here to keep the diff reviewable.
- Metadata is *rejected loudly* in 0.3.4, not supported (#704) — callers needing per-vector metadata should keep a sidecar for now.

🤖 Generated with [RuFlo](https://github.com/ruvnet/ruflo)

### DIFF
--- a/npm/packages/ruvector/package.json
+++ b/npm/packages/ruvector/package.json
@@ -89,7 +89,7 @@
     "@metaharness/weight-eft": "^0.1.1",
     "@metaharness/workspace-lens": "^0.1.2",
     "@metaharness/workspace-probe": "^0.1.1",
-    "@ruvector/rvf": "^0.1.0",
+    "@ruvector/rvf": "^0.3.4",
     "@ruvector/tiny-dancer": "^0.1.22",
     "metaharness": "^0.4.7"
   },
--- a/npm/packages/rvf-mcp-server/package.json
+++ b/npm/packages/rvf-mcp-server/package.json
@@ -34,7 +34,7 @@
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.30.0",
-    "@ruvector/rvf": "^0.1.2",
+    "@ruvector/rvf": "^0.3.4",
     "express": "^4.18.0",
     "zod": "^3.22.0"
   },


ruvnet (commented):
Dream cycle exact-head review at `c334e350314b0e5826c56e0bde3d29110d0e702b`.

Frozen hypothesis: the two manifest-floor changes safely repair string-ID integrity in the installable/published CLI while preserving workspace compatibility.

The dependency direction is relevant to confirmed RVF data-integrity issue #704. However, only two package manifests change; the committed lockfile is not updated and no republished-artifact receipt exists. Required Workspace CI run 34616661436 failed after 3,546/3,547 tests: `reasoning_bank::tests::test_trajectory_recording` could not acquire the database lock, and a sibling shard was cancelled. Focused native, WASM-dedup, regression, supply-chain, npm-functional, formatting, and clippy workflows passed.

Promotion still requires regenerated lock state, clean install from the exact tree, three distinct string IDs surviving close/reopen without collapse, independent-reader verification, explicit metadata rejection or durable metadata support, native/WASM parity, and green Workspace CI.

Verdict: **REJECT**.

